PR #4, opened 2019-05-05
Setup CircleCI

### LINKED ISSUE #3: Setup CI and publishing
@NeQuissimus do we have a project structure I could use for this? I know my way around CircleCI so you could throw the general instructions at me :-)

ghostdogpr:
Closes #3

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,104 @@
+version: 2.1
+jobs:
+  lint212_jdk8:
+    docker:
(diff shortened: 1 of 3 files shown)


ghostdogpr:
I disabled releases on master because it was creating a “real” release (not a snapshot), therefore polluting the list of versions in sonatype/maven. Another problem was that we could only have a single project in the org releasing at the same time: if we release for every commit to master, the chances of conflict are pretty high.

iravid:
@ghostdogpr gotcha, thanks

PR #12, opened 2019-06-11
Added hello world

tampler:


iravid:
Hi @tampler, the code added probably does not belong on this repository but rather on a tutorial for using the plain Java consumer. Thanks for submitting this.

PR #13, opened 2019-06-11
Added zio hello integration test

tampler:
Hello

I tried to add the simplest Hello world test for ZIO Consumer, which should poll from my localhost:9092 and println. This is a follow up work for my prev PR, which implements the simplest java-style Consumer

However, my test fails with a timeout. Could you pls take a look?

Thanks
Boris

PS: I had to change scalacOptions to make it work. Which parameter in your build.sbt makes sbt to treat unused imports as errors? I tried to find and comment it, but had to change to my parameters to compile the thing.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,6 +4,7 @@ val allScala  = Seq("2.11.12", mainScala)
 organization := "dev.zio"
 homepage := Some(url("https://github.com/zio/zio-kafka"))
 name := "zio-kafka"
+maxErrors := 3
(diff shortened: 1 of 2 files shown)


iravid:
If I am not mistaken, these scenarios are already covered by the existing tests. If you're having problems with using the library, please feel free to reach out on Gitter.

tampler:
Hi @iravid 
No, this case was not covered, since I'm using an external Server and Broker. Could you pls advice on the code? Why do I get a timeout here?

Issue #11, opened 2019-06-04
Add `sendChunk` to the Producer


iravid:
Signature: `def produceChunk(records: Chunk[ProducerRecord[K, V]]): BlockingTask[RecordMetadata]`.

iravid:
Resolved in #14

PR #14, opened 2019-06-23
add Producer.sendChunk (#11)

Wsxqaz:
Took some liberty with the function signature :D

### DIFF
--- a/src/main/scala/zio/kafka/client/Producer.scala
+++ b/src/main/scala/zio/kafka/client/Producer.scala
@@ -1,15 +1,20 @@
 package zio.kafka.client
 
+import java.util.concurrent.Future
+import java.util.concurrent.atomic.AtomicLong
+
 import scala.collection.JavaConverters._
 
 import org.apache.kafka.clients.producer.{ Callback, KafkaProducer, ProducerRecord, RecordMetadata }
 import org.apache.kafka.common.serialization.Serde
 
-import scalaz.zio.{ Promise, UIO, ZIO, ZManaged }
+import scalaz.zio.{ Chunk, Promise, UIO, ZIO, ZManaged }
 import scalaz.zio.blocking._
 
 trait Producer[K, V] {
   def produce(record: ProducerRecord[K, V]): BlockingTask[RecordMetadata]
+
+  def produceChunk(records: Chunk[ProducerRecord[K, V]]): BlockingTask[Array[RecordMetadata]]
 
   def flush: BlockingTask[Unit]
 }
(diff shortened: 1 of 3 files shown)


src/main/scala/zio/kafka/client/Producer.scala
```diff
@@ -39,6 +42,13 @@ object Producer {
 
       def flush: BlockingTask[Unit] =
         effectBlocking(p.flush())
+
+      def produceChunk(records: Chunk[ProducerRecord[K, V]]): BlockingTask[List[RecordMetadata]] =
+        records.foldM(List.empty[RecordMetadata])(
+          (a: List[RecordMetadata], rec: ProducerRecord[K, V]) => {
+            Task.fromFutureJava(() => p.send(rec)).map(r => (r :: a))
```
iravid: Let's try to write this in a more low-level way, as the point of `produceChunk` is to amortize the cost of going through the ZIO runtime.

- We should run the `send` calls for all of the chunk, and only then wait for the resulting futures, otherwise this is going to be VERY slow as you'll be waiting for the producer's inner batching mechanism to run on each record
- Let's avoid the interop-java dependency and write this using plain ZIO; you can use `Promise.unsafeMake` to get a hold of a promise without going through the runtime
- The loop that runs the send calls should use blocking

Let me know if this makes sense!

Wsxqaz: Makes sense, I'll give it another shot!

Wsxqaz:
Let me know if the changes are going in the right direction :D

src/main/scala/zio/kafka/client/Producer.scala
```diff
@@ -39,6 +41,37 @@ object Producer {
...
+      def produceChunk(records: Chunk[ProducerRecord[K, V]]): BlockingTask[List[RecordMetadata]] =
+        if (records.isEmpty) {
+          ZIO.succeed(List.empty[RecordMetadata])
+        } else {
+          for {
+            done    <- Promise.make[Throwable, List[RecordMetadata]]
+            runtime <- ZIO.runtime[Blocking]
+            ref     <- Ref.make(List.empty[RecordMetadata])
+            _ <- records.traverse_ { (rec: ProducerRecord[K, V]) =>
+                  effectBlocking {
```
iravid: Hi @Wsxqaz, sorry, this dropped off my radar. The problem with `traverse_` is that it'll be done sequentially; each send would wait for the record to be acknowledged before proceeding to the next.

A better approach here is to use a `while` loop to iterate over the chunk and run the `send` for each element.

Wsxqaz:
@iravid Is this what you meant but using a while loop?

src/main/scala/zio/kafka/client/Producer.scala
```diff
@@ -39,6 +41,40 @@ object Producer {
...
+      def produceChunk(records: Chunk[ProducerRecord[K, V]]): BlockingTask[List[RecordMetadata]] =
+        if (records.isEmpty) {
+          ZIO.succeed(List.empty[RecordMetadata])
+        } else {
+          for {
+            done    <- Promise.make[Throwable, List[RecordMetadata]]
+            runtime <- ZIO.runtime[Blocking]
+            ref     <- Ref.make(List.empty[RecordMetadata])
+            _ <- effectBlocking {
+                  val it: Iterator[ProducerRecord[K, V]] = records.toArray.iterator
```
iravid: Excellent work! Yes, this is the right direction.

We now have two more items to address:
- The metadata ref might be constructed out-of-order which is unfortunate. To work around this, we can pre-allocate an array of `RecordMetadata` that'll be updated directly (by index) from the callback. That array can be updated directly without a ref. You'll need to do the callback counting manually using an `AtomicLong`.
- This one is nice-to-have but not critical: It'd be nice to support interruption in the loop. `effectBlocking` uses `Thread.interrupt`, so you can check that using `Thread.isInterrupted`. Would also be nice to interrupt the already-in-flight futures with `Future#cancel(true)`.

Wsxqaz: I'm not sure I totally follow what you mean about supporting interruption in the loop. What would we do if the thread has been interrupted?

iravid: The minimum is to break out of the loop early. We could also make the extra effort and run `cancel` on the futures of the already-submitted sends.

Wsxqaz: So if `Thread.isInterrupted == true` then do something like https://github.com/Wsxqaz/zio-kafka/blob/add-sendChunk/src/main/scala/zio/kafka/client/Producer.scala#L68 ?

iravid: Yes, exactly. But I'm not sure if it's wise to do that if an error is encountered. What's your take on that?

Wsxqaz: I figured that since once we encounter an error the counter will never reach the length of the chunk, so we can just give up on everything we've started. 

Though I realise that the array of `futures` should probably be `val futures = Array.fill[Option[Future[RecordMetadata]]](records.length)(None)` just so we don't get any funny business with `null` values. 

Does that make sense?

Wsxqaz:
I'd be happy to handle interruption in a follow-up pr.

iravid:
Thanks @Wsxqaz! Great work!

PR #22, opened 2019-09-25
High-level API for consuming messages with an effect

svroonland:
See docs. This provides a very easy, high-level API to apply an effect to each message, while benefitting from parallelism and automatic committing after processing.


src/main/scala/zio/kafka/client/Consumer.scala
```diff
@@ -47,4 +48,54 @@ object Consumer {
...
+                    case CommittableRecord(record, offset) =>
+                      f(record.key(), record.value()).as(offset)
+                  }.flattenChunks
+              }
+          }
+      }
+      .aggregate(ZSink.foldLeft[Offset, OffsetBatch](OffsetBatch.empty)(_ merge _))
+      .mapM(_.commit)
+      .tap(_ => putStrLn(s"Commit done"))
+      .run(ZSink.drain)
```
```suggestion
      .runDrain
```

src/main/scala/zio/kafka/client/Consumer.scala
```diff
@@ -47,4 +48,54 @@ object Consumer {
...
+                  partitionStream.mapM {
+                    case CommittableRecord(record, offset) =>
+                      f(record.key(), record.value()).as(offset)
+                  }.flattenChunks
+              }
+          }
+      }
+      .aggregate(ZSink.foldLeft[Offset, OffsetBatch](OffsetBatch.empty)(_ merge _))
+      .mapM(_.commit)
+      .tap(_ => putStrLn(s"Commit done"))
```
iravid: No printlns :-)

src/main/scala/zio/kafka/client/Consumer.scala
```diff
@@ -47,4 +48,54 @@ object Consumer {
...
+   * @param subscription Topic subscription parameters
+   * @param f Function that returns the effect to execute for each message. It is passed the key and value
+   * @tparam K Type of keys (an implicit [[Serde]] should be in scope)
+   * @tparam V Type of values (an implicit [[Serde]] should be in scope)
+   * @return Effect that completes with a unit value only when interrupted. May fail when the [[Consumer]] fails.
+   */
+  def consumeM[R, K: Serde, V: Serde](
+    subscription: Subscription,
+    settings: ConsumerSettings
+  )(f: (K, V) => ZIO[R, Nothing, Unit]): ZIO[R with Clock with Blocking with Console, Throwable, Unit] =
```
```suggestion
  )(f: (K, V) => ZIO[R, Nothing, Unit]): ZIO[R with Clock with Blocking, Throwable, Unit] =
```

src/main/scala/zio/kafka/client/Consumer.scala
```diff
@@ -47,4 +48,54 @@ object Consumer {
...
+   *
+   * Messages are processed with 'at least once' consistency: it is not guaranteed that every message
+   * that is processed by the effect has a corresponding offset commit before stream termination.
+   *
+   * Offsets are batched and committed after execution of the effect.
+   *
+   * The effect should must absorb any failures. Failures should be handled by retries or ignoring the
+   * error, which will result in the Kafka message being skipped.
+   *
+   * Usage example:
```
iravid: An example would be nice, yeah!

src/main/scala/zio/kafka/client/Consumer.scala
```diff
@@ -47,4 +48,54 @@ object Consumer {
...
+   *
+   * @param settings Settings for creating a [[Consumer]]
+   * @param subscription Topic subscription parameters
+   * @param f Function that returns the effect to execute for each message. It is passed the key and value
+   * @tparam K Type of keys (an implicit [[Serde]] should be in scope)
+   * @tparam V Type of values (an implicit [[Serde]] should be in scope)
+   * @return Effect that completes with a unit value only when interrupted. May fail when the [[Consumer]] fails.
+   */
+  def consumeM[R, K: Serde, V: Serde](
```
iravid: We don't have a version of this that takes a `(K, V) => Unit`, and we probably won't have. So let's call this `consumeWith`.

src/main/scala/zio/kafka/client/Consumer.scala
```diff
@@ -47,4 +48,54 @@ object Consumer {
...
+  )(f: (K, V) => ZIO[R, Nothing, Unit]): ZIO[R with Clock with Blocking with Console, Throwable, Unit] =
+    ZStream
+      .managed(Consumer.make[K, V](settings))
+      .flatMap { consumer =>
+        ZStream
+          .fromEffect(consumer.subscribe(subscription))
+          .flatMap { _ =>
+            consumer.partitioned
+              .flatMapPar(Int.MaxValue, outputBuffer = settings.perPartitionChunkPrefetch) {
+                case (partition @ _, partitionStream) =>
```
```suggestion
                case (_, partitionStream) =>
```

src/main/scala/zio/kafka/client/Consumer.scala
svroonland: I'm not happy about the `R` type parameter here, but since `K` and `V` need to be specified, I don't see how to get rid of `R`.

iravid: Well, the consuming function could require any `R`, so it is more correct to be polymorphic in the dependencies used (in addition to Blocking and Clock).

iravid:
Ok, let's just add a test and this should be good to go!

PR #23, opened 2019-09-27
Resilience, metadata API and upgrades

iravid:


andreamarcolin:
If I am not mistaken, checks are failing due to scalafmt missing. I noticed that the version specified in the .scalafmt.conf does not match with the scalafmt plugin version. Shouldn't they be aligned? The latest stable release is 2.0.1 so the plugin seems up to date, we should only need to update .scalafmt .conf

iravid:
It was actually a weird issue with Coursier. I disabled it in the build for now.

PR #38, opened 2019-10-01
Add nicer constructors for the Subscription terms

### LINKED ISSUE #27: Add nicer constructors for the `Subscription` terms
```scala
def topics(name: String, names: String*): Subscription
// Overloads for: Scala pattern, Java pattern, String pattern (but that needs to be in Task)
def pattern(pattern: Regex)
```

pgabara:
Fixes #27


pgabara:
Thanks!

PR #44, opened 2019-10-04
Serde followups

iravid:
@svroonland - could you have a look?


src/main/scala/zio/kafka/client/CommittableRecord.scala
```diff
@@ -5,7 +5,33 @@ import org.apache.kafka.common.TopicPartition
 import zio.{ RIO, Task }
 import zio.kafka.client.serde.Deserializer
 
-final case class CommittableRecord[K, V](record: ConsumerRecord[K, V], offset: Offset)
+final case class CommittableRecord[K, V](record: ConsumerRecord[K, V], offset: Offset) {
+  def deserializeWith[R, K1, V1](
+    keyDeserializer: Deserializer[R, K1],
+    valueDeserializer: Deserializer[R, V1]
+  )(implicit ev1: K <:< Array[Byte], ev2: V <:< Array[Byte]): RIO[R, CommittableRecord[K1, V1]] =
```
svroonland: Interesting trick this subtype evidence, never used it before.

src/main/scala/zio/kafka/client/Producer.scala
```diff
@@ -93,8 +93,8 @@ object Producer {
         r: ProducerRecord[K, V]
       ): RIO[R, ByteArrayProducerRecord] =
         for {
-          key   <- keySerializer.serialize(r.key())
-          value <- valueSerializer.serialize(r.value())
+          key   <- keySerializer.serialize(r.topic, r.headers, r.key())
```
svroonland: 👍 
I never saw the need for the serializer to know the topic and headers, but I like the idea of not imposing that.

src/main/scala/zio/kafka/client/serde/Serde.scala
```diff
@@ -1,94 +1,69 @@
...
-trait Serializer[-R, -T] {
-  def serialize(value: T): RIO[R, Array[Byte]]
-}
-
-object Serializer {
-
-  /**
-   * Obtain an instance of a Serializer of some type from the implicit scope
-   */
-  def of[T](implicit ser: Serializer[Any, T]): Serializer[Any, T] = ser
```
svroonland: Any particular reason for removing the `of`s?

src/test/scala/zio/kafka/client/ConsumerTest.scala
```diff
@@ -48,7 +48,7 @@ class ConsumerTest extends WordSpecLike with Matchers with LazyLogging with Defa
           _       <- consumer.subscribe(Subscription.Topics(Set("topic150")))
           kvs     <- ZIO((1 to 5).toList.map(i => (s"key$i", s"msg$i")))
           _       <- produceMany("topic150", kvs)
-          records <- consumer.plainStream(Serde.of[String], Serde.of[String]).flattenChunks.take(5).runCollect
+          records <- consumer.plainStream(Serde.string, Serde.string).flattenChunks.take(5).runCollect
```
svroonland: Agreed that for the primitive serdes there's no advantage of `Serde.of[String]` over `Serde.string`.

src/main/scala/zio/kafka/client/serde/Serde.scala
iravid: I'd like to move away from encouraging users to use implicit scope for the serializer/deserializer instances.

In other words - not treat them like typeclasses at all, but rather as simple functions.

Issue #32, opened 2019-09-28
Add a pre-defined sink on the Consumer object that batches offsets


iravid:
And expand the quickstart example in the README.md to include using it with `ZStream#aggregate`.

andreamarcolin:
Will take this too

iravid:
Resolved by #48

PR #48, opened 2019-10-05
Add a batching sink

andreamarcolin:
Implements #32


src/main/scala/zio/kafka/client/Consumer.scala
```diff
@@ -100,6 +100,9 @@ class Consumer private (
 }
 
 object Consumer {
+  val batchingSink: ZSink[Any, Nothing, Nothing, Offset, OffsetBatch] =
```
```suggestion
  val offsetBatches: ZSink[Any, Nothing, Nothing, Offset, OffsetBatch] =
```

iravid:
Needs a rebase

PR #53, opened 2019-10-17
Add tests for Pattern based Subsctiption

pinikeizman:
Hey, I had some trouble configuring a Subscription with Regex. 
I added this test to make sure it works.
I might be missing something would love to get some help here.

This is what I got so far:
- Metadata refresh is on 1s for the apache Kafka consumer.
- I added some logs in the rebalance listener CB and they didn't invoke (which they did invoke when I used Topic set subscription)

### DIFF
--- a/src/test/scala/zio/kafka/client/ConsumerTest.scala
+++ b/src/test/scala/zio/kafka/client/ConsumerTest.scala
@@ -3,16 +3,15 @@ package zio.kafka.client
 import com.typesafe.scalalogging.LazyLogging
 import net.manub.embeddedkafka.EmbeddedKafka
 import org.apache.kafka.clients.consumer.ConsumerConfig
-import org.apache.kafka.common.TopicPartition
 import org.scalatest.{ EitherValues, Matchers, WordSpecLike }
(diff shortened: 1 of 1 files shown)


iravid:
Thanks for adding the test @pinikeizman! It's currently failing on some compile errors. Let me know if you need help in fixing them.

iravid:
@pinikeizman From what I recall, pattern subscriptions don't work with start/end anchors (`^`, `$`). Can you try something like `topic[0-9]+` for the pattern?

pinikeizman:
@iravid  NP I'll try it.

pinikeizman:
@iravid it still fails.

iravid:
Hmm that's pretty strange. There's no particular difference between pattern subscriptions and topic subscriptions in this library's code.

Thanks for the test, I'll have time to look into this next week. 

Feel free to keep poking around yourself - it may be beneficial to increase the logging level when running the tests and see what's the output from the consumer and the embedded Kafka broker.

iravid:
Cherry-picked into #53

PR #55, opened 2019-10-29
Implement Graceful Shutdown

### LINKED ISSUE #29: Add a way to gracefully shutdown an open stream
Ideally, this would allow to somehow drain the stream until all already-fetched records have been processed and then close.

adamgfraser:
Resolves #29.

### DIFF
--- a/src/main/scala/zio/kafka/client/Runloop.scala
+++ b/src/main/scala/zio/kafka/client/Runloop.scala
@@ -85,6 +85,12 @@ object Runloop {
                        .unbounded[
                          Take[Throwable, (TopicPartition, ZStreamChunk[Any, Throwable, ByteArrayCommittableRecord])]
                        ]
+                       .map { queue =>
+                         queue.mapM {
+                           case Take.End => queue.shutdown.as(Take.End)
+                           case x        => ZIO.succeed(x)
+                         }
+                       }
                        .toManaged(_.shutdown)
         listener <- ZIO
                      .runtime[Blocking]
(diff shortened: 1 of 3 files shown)


adamgfraser:
@iravid Thanks! Tried to address all your comments. Not sure I got everything quite right so please take a look when you have the chance.

src/main/scala/zio/kafka/client/Runloop.scala
```diff
@@ -384,17 +407,34 @@ object Runloop {
                    else doCommit(List(cmd)).as(state)
       } yield newState
 
+    def handleShutdown(state: State, cmd: Command): BlockingTask[State] = cmd match {
+      case Command.Poll() => UIO.succeed(state)
```
iravid: Need to call `handlePoll` here, no?

adamgfraser: Yes! Fixed now.

src/main/scala/zio/kafka/client/Runloop.scala
```diff
@@ -384,17 +407,34 @@ object Runloop {
                    else doCommit(List(cmd)).as(state)
       } yield newState
 
+    def handleShutdown(state: State, cmd: Command): BlockingTask[State] = cmd match {
+      case Command.Poll() => handlePoll(state)
```
iravid: One last thing is bothering me here. When the shutdownRef is switched to `true`, there are probably still some requests pending in the state. This means that:
- We enter `handlePoll` with pending requests; 
- The partitions for those requests will be resumed (line 327);
- The poll will return records for those, and those records are discarded in line 343.

We can probably do better by just making sure that when shutdownRef = true, we always enter handlePoll with no requests. This can be done by eagerly resolving them at this line. Then `handlePoll` can stay unchanged without being aware of the shutdown state.

adamgfraser: Updated eagerly resolve pending requests.

PR #58, opened 2019-11-08
Fix type inference issue in Producer.make

svroonland:
Producer.make(settings, Serde.string, Serde.int) would be inferred to be of type `ZManaged[Blocking, Throwable, Producer[Nothing, String, Int]]`.

This fix changes that to infer to `ZManaged[Blocking, Throwable, Producer[Any, String, Int]]`.

`Producer` could theoretically be contravariant in `K` and `V` as well if it wasn't for the invariant `ProducerRecord[K, V]` from the apache kafka library.

### DIFF
--- a/src/main/scala/zio/kafka/client/Producer.scala
+++ b/src/main/scala/zio/kafka/client/Producer.scala
@@ -11,7 +11,7 @@ import zio.stream.ZSink
 
 import scala.collection.JavaConverters._
 
-class Producer[R, K, V] private (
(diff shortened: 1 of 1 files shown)


iravid:
Thanks @svroonland. Do we have the same issue with `Consumer`?

svroonland:
Don't think so, `Consumer` does not have type parameters. I don't see a problem with  `partitionedStream` and `plainStream`.

iravid:
Alright, thanks!

PR #59, opened 2019-11-09
tests refactored to zio-test

TimPigden:
quite a lot of stuff pulled out into KafkaTestUtils 
Note the creation of the KafkaTestEnvironment which is provided to wrap the initialisation of the embedded kafka and extract the broker list. Would the Kafka module or something like it be of value outside of tests?


TimPigden:
no, you can merge. the client admin is in a 2nd pr
I'll be doing more on docs but not for a few days.

PR #60, opened 2019-11-09
Add Serializer and Deserializer constructors

svroonland:
based on their Kafka equivalents, just like `Serde` has. 

Useful to have now that Producer and Consumer take just a Serializer or Deserializer instead of a whole Serde.


iravid:
Great addition @svroonland!

PR #66, opened 2019-11-13
Manual topic-partition assignment

svroonland:
Makes use of `KafkaConsumer.assign` instead of `KafkaConsumer.subscribe`


svroonland:
Waiting for the zio-test PR to be merged before I'll write some tests for this.

PR #76, opened 2019-11-18
Upgrade to ZIO 1.0.0-RC17

svroonland:


svroonland:
@iravid The tests seem to succeed, but fail on the closing of resources at lines like `_ <- fib.join.ignore`. Probably something to do with the fiber supervision overhaul. I can't quite figure out how to fix it.

iravid:
Uh oh. Looks like we're not releasing today then. If you don't figure it out by tomorrow I will join in on the fun.

svroonland:
Think I fixed everything :)

I did not process the non-mandatory changes from the release notes.

src/test/scala/zio/kafka/client/KafkaTestUtils.scala
```diff
@@ -131,7 +131,7 @@ object KafkaTestUtils {
       ),
       250.millis,
       250.millis,
-      1
+      100
```
svroonland: This improves test speed a bit

TimPigden: Hi - can you explain that please. These numbers were a bit of amystery to me

svroonland: Well, to be honest, I believe it helped to speed up a test I ran but I'd have to look closer at the exact mechanis in `RunLoop` to say more about its exact effect. 

In any case, the setting is used as the parameter to `ZStream.buffer` which:
> Allows a faster producer to progress independently of a slower consumer by bufferingup to `capacity` elements in a queue.

In backpressuring streams it helps to have some buffers to deal with non-constant processing speeds.

PR #80, opened 2019-11-20
Make implementation classes private

svroonland:


src/main/scala/zio/kafka/client/impl/ConsumerAccess.scala
```diff
@@ -1,15 +1,16 @@
-package zio.kafka.client
+package zio.kafka.client.impl
```
iravid: Let's call it `internal`

iravid:
Can you rebase this?

Issue #42, opened 2019-10-02
Handle RetriableCommitFailedException


svroonland:
In my experience commits sometimes fail because of some temporary situation, like a network timeout. Kafka will then throw a `RetriableCommitFailedException`. This should not fail the consumer.

We should probably handle this in the callback passed to `commitAsync` in the `RunLoop`

iravid:
Failed commits will not fail the runloop; the failure will be passed back to the caller of `OffsetBatch#commit`. The alternative is to bake in a `Schedule` that will be used for retrying commits in the runloop, but I feel that is best delegated to the caller.

What do you think?

svroonland:
That makes sense, to leave it to the caller. Might be good to add it as a recommendation in the docs somewhere, and it would be nice to have that behavior in `consumeWith`.

iravid:
👍

svroonland:
This was implemented in #81. Shall we close it @iravid ?

svroonland:
I'll take that as a yes :)

PR #81, opened 2019-11-20
Retry commits in consumeWith when failing due to a RetriableCommitFailedException 

svroonland:
Implements #42 

Adds a `commitOrRetry` method to `Offset` and `OffsetBatch` so it can be used by clients that do not use `consumeWith`


svroonland:
Suggestion how to fix the doc issue thing? It's either this or a unused import warning as far as I can see.

src/main/scala/zio/kafka/client/Offset.scala
```diff
@@ -1,13 +1,35 @@
...
+   * with a [[org.apache.kafka.clients.consumer.RetriableCommitFailedException]]
+   */
+  def commitOrRetry[R, B](policy: Schedule[R, Any, B]): ZIO[R with Clock, Throwable, Unit] =
+    Offset.commitOrRetry(commit, policy)
+}
+
+object Offset {
+  private[client] def commitOrRetry[R, B](
+    commit: Task[Unit],
+    policy: Schedule[R, Any, B]
```
iravid: If you make this `Schedule[R, Throwable, B]`, the caller will be able to provide a schedule that inspects the Throwable. Is that useful?

iravid:
I'm not that familiar with scaladoc so not sure what the right thing to do there is. You can just drop the `[[ ]]` for now.

src/main/scala/zio/kafka/client/Offset.scala
svroonland: Yeah I suppose so, if you want to retry only on some subtypes of `RetriableCommitException`

src/main/scala/zio/kafka/client/Offset.scala
```diff
@@ -1,13 +1,35 @@
...
   def topicPartition: TopicPartition
   def offset: Long
   def commit: Task[Unit]
   def batch: OffsetBatch
+
+  /**
+   * Attempts to commit and retries according to the given policy when the commit fails
+   * with a RetriableCommitFailedException
+   */
+  def commitOrRetry[R, B](policy: Schedule[R, Throwable, B]): ZIO[R with Clock, Throwable, Unit] =
```
iravid:
```suggestion
  def commitOrRetry[R](policy: Schedule[R, Throwable, Any]): ZIO[R with Clock, Throwable, Unit] =
```

src/main/scala/zio/kafka/client/OffsetBatch.scala
```diff
@@ -1,13 +1,21 @@
...
   def offsets: Map[TopicPartition, Long]
   def commit: Task[Unit]
   def merge(offset: Offset): OffsetBatch
   def merge(offsets: OffsetBatch): OffsetBatch
+
+  /**
+   * Attempts to commit and retries according to the given policy when the commit fails
+   * with a RetriableCommitFailedException
+   */
+  def commitOrRetry[R, B](policy: Schedule[R, Throwable, B]): ZIO[R with Clock, Throwable, Unit] =
```
iravid:
```suggestion
  def commitOrRetry[R](policy: Schedule[R, Throwable, Any]): ZIO[R with Clock, Throwable, Unit] =
```

PR #92, opened 2019-11-29
Scala 2.13

lvitaly:
Scala 2.13 support with Embedded Kafka

- Scala 2.13.1
- Kafka 2.4.0-RC1

_Inspired by work done in https://github.com/akka/alpakka-kafka/pull/971_


iravid:
Thank you @lvitaly! Looks good to me. Is this ready for merging?

lvitaly:
@iravid yes, this is ready for merging. I've fixed all cross-compile and CI issues. And it would be great to publish it as RC or something like this.

seglo:
> it would be great to publish it as RC or something like this.

Just an FYI that since Apache Kafka RCs are only published to the Apache Staging repo, users will need to include that repo in their builds if you do a release.

Also, because the RC artifact is called `2.4.0` anyone who downloads it may have a stale artifact in their local cache once the final release drops.

We put some notes about this in our milestone release: https://doc.akka.io/docs/alpakka-kafka/2.0/release-notes/2.0.x.html#2-0-0-m2

Issue #102, opened 2019-12-20
Support external offset retrieval


svroonland:
Currently partitions will be consumed from the latest offset stored in Kafka. We should also support a way to get offsets from an external source.

svroonland:
Working on this one currently

iravid:
Thanks, this really is an important usecase.

svroonland:
Implemented by #104

PR #104, opened 2019-12-23
External offset retrieval

svroonland:
Support retrieving offsets from an external source by providing an `OffsetRetrieval.Manual(getOffsets)` to `ConsumerSettings`.

Implements #102

### DIFF
--- a/src/main/scala/zio/kafka/client/Consumer.scala
+++ b/src/main/scala/zio/kafka/client/Consumer.scala
@@ -6,12 +6,14 @@ import zio._
 import zio.blocking.Blocking
 import zio.clock.Clock
 import zio.duration._
+import zio.kafka.client.Consumer.OffsetRetrieval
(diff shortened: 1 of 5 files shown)


svroonland:
@iravid or @adamgfraser if you have time to review this one?

iravid:
@svroonland Thanks for the reminder. I'll review today.

svroonland:
Good point. Well, there's no reason why clients could not call `seek*` during streaming to change the offset. Although it's not very well determined what the next processed message would be after calling seek. The same can be said for calling `subscribe` and `unsubscribe`. 

I guess the question boils down to: do we want to support that and can we do that in a way that is useful (deterministic) to the users. What do you think?

iravid:
Let's deprecate them in a follow-up. I don't see a way to deterministically use them currently.

PR #107, opened 2020-01-01
Change back to official embedded-kafka

guymers:
The official embedded-kafka is now available for 2.4.0, it should be
used instead of the temporary seglo published version.


iravid:
Thanks @guymers!

PR #109, opened 2020-01-02
WIP: Consumer/Producer Modules

### LINKED ISSUE #106: Modules for the Consumer and Producer
Implement modules (see https://zio.dev/docs/howto/howto_use_module_pattern) for the `Consumer` and `Producer`. This helps to integrate them into larger ZIO applications and for testing/mocking.

See also #93  and #94 for an example

SteveXCIV:
Adds modules for the producer and consumer. Current implementation provides the trait + service value structure common to ZIO modules.

PR opened to collect comments & suggestions.

Below are some outstanding questions/thoughts for improving this implementation:

- `Producer` has type parameters `K` and `V` for the key and value deserializers, this violates the accessible/mockable constraint of having only one type parameter for the environment (commonly `R`)
- `Consumer` implementation has methods that don't return effects; this is fixable, but needs to be done in a way that doesn't break existing downstream code
- `Producer` has varying environment type for the effects it returns (some methods are using `R with Blocking` whereas some return a `BlockingTask`)

Fixes #106

### DIFF
--- a/src/main/scala/zio/kafka/client/Consumer.scala
+++ b/src/main/scala/zio/kafka/client/Consumer.scala
@@ -13,11 +13,60 @@ import zio.stream._
 
 import scala.jdk.CollectionConverters._, scala.collection.compat._
 
+trait KafkaConsumer {
+  val kafakConsumer: KafkaConsumer.Service
+}
+
+object KafkaConsumer {
+  trait Service {
+    def assignment: BlockingTask[Set[TopicPartition]]
+    def beginningOffsets(
+      partitions: Set[TopicPartition],
+      timeout: Duration = Duration.Infinity
+    ): BlockingTask[Map[TopicPartition, Long]]
+    def committed(
+      partitions: Set[TopicPartition],
+      timeout: Duration = Duration.Infinity
+    ): BlockingTask[Map[TopicPartition, Option[OffsetAndMetadata]]]
+    def endOffsets(
+      partitions: Set[TopicPartition],
+      timeout: Duration = Duration.Infinity
+    ): BlockingTask[Map[TopicPartition, Long]]
+    def stopConsumption: UIO[Unit]
+    def listTopics(timeout: Duration = Duration.Infinity): BlockingTask[Map[String, List[PartitionInfo]]]
+    def offsetsForTimes(
+      timestamps: Map[TopicPartition, Long],
+      timeout: Duration = Duration.Infinity
+    ): BlockingTask[Map[TopicPartition, OffsetAndTimestamp]]
+    def partitionedStream[R, K, V](
+      keyDeserializer: Deserializer[R, K],
+      valueDeserializer: Deserializer[R, V]
+    ): ZStream[
+      Clock with Blocking,
+      Throwable,
+      (TopicPartition, ZStreamChunk[R, Throwable, CommittableRecord[K, V]])
+    ]
+    def partitionsFor(topic: String, timeout: Duration = Duration.Infinity): BlockingTask[List[PartitionInfo]]
+    def position(partition: TopicPartition, timeout: Duration = Duration.Infinity): BlockingTask[Long]
+    def plainStream[R, K, V](
+      keyDeserializer: Deserializer[R, K],
+      valueDeserializer: Deserializer[R, V]
+    ): ZStreamChunk[R with Clock with Blocking, Throwable, CommittableRecord[K, V]]
+    def seek(partition: TopicPartition, offset: Long): BlockingTask[Unit]
+    def seekToBeginning(partitions: Set[TopicPartition]): BlockingTask[Unit]
+    def seekToEnd(partitions: Set[TopicPartition]): BlockingTask[Unit]
+    def subscribe(subscription: Subscription): BlockingTask[Unit]
+    def subscribeAnd(subscription: Subscription): SubscribedConsumer
+    def subscription(): BlockingTask[Set[String]]
+    def unsubscribe(): BlockingTask[Unit]
+  }
+}
+
 class Consumer private (
   private val consumer: ConsumerAccess,
   private val settings: ConsumerSettings,
   private val runloop: Runloop
-) {
+) extends KafkaConsumer.Service {
   self =>
 
   /**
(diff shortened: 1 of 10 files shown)


src/main/scala/zio/kafka/client/Consumer.scala
```diff
@@ -13,22 +13,71 @@ import zio.stream._
 
 import scala.jdk.CollectionConverters._, scala.collection.compat._
 
+trait KafkaConsumer {
```
iravid: Let's call this `Consumer`.

```suggestion
trait Consumer {
```

src/main/scala/zio/kafka/client/Consumer.scala
```diff
@@ -13,22 +13,71 @@ import zio.stream._
...
+  val kafakConsumer: KafkaConsumer.Service
+}
+
+object KafkaConsumer {
```
```suggestion
object Consumer {
```

src/main/scala/zio/kafka/client/Consumer.scala
```diff
@@ -13,22 +13,71 @@ import zio.stream._
 
 import scala.jdk.CollectionConverters._, scala.collection.compat._
 
+trait KafkaConsumer {
+  val kafakConsumer: KafkaConsumer.Service
```
```suggestion
  val consumer: Consumer.Service
```

src/main/scala/zio/kafka/client/Consumer.scala
```diff
@@ -13,22 +13,71 @@ import zio.stream._
...
+    def seekToBeginning(partitions: Set[TopicPartition]): BlockingTask[Unit]
+    def seekToEnd(partitions: Set[TopicPartition]): BlockingTask[Unit]
+    def subscribe(subscription: Subscription): BlockingTask[Unit]
+    def subscribeAnd(subscription: Subscription): SubscribedConsumer
+    def subscription(): BlockingTask[Set[String]]
+    def unsubscribe(): BlockingTask[Unit]
+  }
+}
+
 class Consumer private (
```
iravid: Let's move this into the `Consumer` object and call it `Live`.

src/main/scala/zio/kafka/client/Producer.scala
```diff
@@ -11,11 +11,23 @@ import zio.stream.ZSink
 
 import scala.jdk.CollectionConverters._
 
+trait KafkaProducer[K, V] {
```
iravid:
```suggestion
trait Producer[R, K, V] {
```

src/main/scala/zio/kafka/client/Producer.scala
```diff
@@ -11,11 +11,23 @@ import zio.stream.ZSink
...
+  val kafkaProducer: KafkaProducer.Service[K, V, Any]
```
iravid:
```suggestion
  val producer: Producer.Service[R, K, V]
```

src/main/scala/zio/kafka/client/Producer.scala
```diff
@@ -11,11 +11,23 @@ import zio.stream.ZSink
...
+}
+
+object KafkaProducer {
```
iravid:
```suggestion
object Producer {
```

src/main/scala/zio/kafka/client/Producer.scala
```diff
@@ -11,11 +11,23 @@ import zio.stream.ZSink
...
+  trait Service[-R, K, V] {
+    def produce(record: ProducerRecord[K, V]): RIO[R with Blocking, Task[RecordMetadata]]
+    def produceChunk(records: Chunk[ProducerRecord[K, V]]): RIO[R with Blocking, Task[Chunk[RecordMetadata]]]
+    def flush(): BlockingTask[Unit]
+  }
+}
+
 class Producer[-R, K, V] private (
```
iravid: Same here- let's move this into `Producer` and call it `Live`.

SteveXCIV:
Hi @iravid, I implemented the naming and structural changes you asked for. I have a follow-up question regarding your other comments.

I'm not 100% clear on what changes would be required to provide the accessors/constructors you mentioned. When you mentioned accessors, does that mean something similar to the other `Live` traits that ZIO core provides for mixing into the environment? If so, I'm not sure what the signature of that should look like. I looked at how this is done in ZIO core with `Blocking`, `Console`, etc., and those are all implemented with a combination of a `trait Live` that contains the live implementation logic and an `object Live extends Live` to enable singleton access. Aside from the fact that the `Consumer` and `Producer` are contained within a `Managed`, the choice of consumer settings or producer serializer/deserializer is arbitrary, so I'm not sure what a sensible live implementation would look like.

iravid:
Right, so we don't want a singleton implementation of `Consumer`/`Producer` because indeed the choice of settings is up to the user.

What we do want is accessors on the companion objects of `Consumer`/`Producer` of the following form:
```scala
object Consumer {
  def listTopics(timeout: Duration): ZIO[Consumer with Blocking, Throwable, Map[String, List[PartitionInfo]]] = 
  ZIO.accessM[Consumer](_.consumer.listTopics(timeout))
}
```

Unfortunately for the producer, because of `K` and `V`, these accessors will infer terribly. But let's write them out first and then figure out what to do.

SteveXCIV:
Thanks for clarifying - I'm going to start on the consumer object now. I agree that the producer is likely going to be a beast to deal with, so I'll push the consumer changes w/ tests before I start on the producer.

SteveXCIV:
@iravid I took a first-pass at tests for the consumer module. I'm still thinking about the most ergonomic way to expose the three service members that don't return a `ZIO`, namely `partitionedStream`, `plainStream` and `subscribeAnd`.

I thought about just wrapping all of these members in an effect at the companion object level, but that has the downsides of diverging from the service trait as well as hiding information about the actual environment requirements. It might not be so bad though, because I think user code could achieve a similar thing today using a combination of `make` and `use`. What do you think?

Also, because this module isn't a 100% faithful implementation of the module pattern as-described, I found myself having to do a manual mock for the consumer. While it did teach me more about ZIO by way of reading the doc comments and code, I'm concerned that such a thing will have a high maintenance cost. I'll keep on mulling this all over offline, but I'm eager to hear your thoughts/recommendations on this.

iravid:
The master branch is now using the snapshot version of ZIO, with the new environment encodings. If you have the time to rebase this and refactor to use the new encoding, feel free to re-open :-) Thank you for the time already put into this branch!

PR #111, opened 2020-01-05
Extend the readme with changes in 0.5.0

svroonland:


svroonland:
Sure, but still seems flaky

iravid:
Ok, let's just disable it for now and fix it after the release.

PR #115, opened 2020-01-19
Builder-style interface and defaults for Consumer/Producer/Admin settings

iravid:


src/main/scala/zio/kafka/client/AdminClientSettings.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  def withProperties(kvs: (String, AnyRef)*): AdminClientSettings =
+    withProperties(kvs.toMap)
+
+  def withProperties(kvs: Map[String, AnyRef]): AdminClientSettings =
+    copy(properties = properties ++ kvs)
+}
+
+object AdminClientSettings {
+  def apply(bootstrapServers: List[String]): AdminClientSettings =
```
svroonland: Should we require at least 1 here?

iravid: Ah that's a great idea.

iravid: Well, I experimented with it and the API is not very ergonomic with something like an actual NonEmptyList. So I'll skip that for now and we can add it in a follow-up if we feel it's important.

PR #128, opened 2020-02-26
Producer and Consumer services as ZLayers

kciesielski:
This PR exposes producer and consumer interfaces as ZLayers. Top-level types are:
```scala
  type Producer[R, K, V] = Has[Producer.Service[R, K, V]]
  type Consumer[R, K, V] = Has[Consumer.Service[R, K, V]]
```
where:
* `R` is the environment required for serialization/deserialization
* `K` is the type of keys
* `V` is the type of values

Public API for creating service layers:
```scala
val pLayer: NoDeps[Throwable, Producer[R, K, V]] = Producer.make(settings, keySerde, valSerde)
val cLayer: NoDeps[Throwable, Consumer[R, K, V]] = Consumer.make(settings, keySerde, valSerde)
```
and from this point layers can be used as dependencies of programs.
It may be worth adding some additional aliases/methods for common use cases:
- cases without the `R` parameter for cases when serialization and deserialization don't have dependencies
- cases where keys and values are `String` or `Array[Byte]`

### DIFF
--- a/src/main/scala/zio/kafka/client/Producer.scala
+++ b/src/main/scala/zio/kafka/client/Producer.scala
@@ -2,7 +2,6 @@ package zio.kafka.client
 
 import java.util.concurrent.atomic.AtomicLong
 
-import izreflect.fundamentals.reflection.Tags.Tag
 import org.apache.kafka.clients.producer.{ Callback, KafkaProducer, ProducerRecord, RecordMetadata }
 import org.apache.kafka.common.serialization.ByteArraySerializer
 import zio._
@@ -17,7 +16,8 @@ object Producer {
   trait Service[R, K, V] {
     def produce(record: ProducerRecord[K, V]): RIO[Blocking with R, Task[RecordMetadata]]
     def produceChunk(records: Chunk[ProducerRecord[K, V]]): RIO[R with Blocking, Task[Chunk[RecordMetadata]]]
-    def stream: ZSink[R with Blocking, Throwable, Nothing, Chunk[ProducerRecord[K, V]], Unit] =
+
+    final def stream: ZSink[R with Blocking, Throwable, Nothing, Chunk[ProducerRecord[K, V]], Unit] =
       ZSink.drain.contramapM(produceChunk)
   }
 
@@ -114,11 +114,11 @@ object Producer {
     valueSerializer: Serializer[R, V]
   ): Live[R, K, V] = new Live(p, keySerializer, valueSerializer)
 
-  def producer[R: Tag, K: Tag, V: Tag](
+  def producer[R, K, V](
     settings: ProducerSettings,
     keySerializer: Serializer[R, K],
     valueSerializer: Serializer[R, V]
-  ): ZLayer.NoDeps[Throwable, HasProducer[R, K, V]] = {
+  )(implicit tagged: Tagged[Service[R, K, V]]): ZLayer.NoDeps[Throwable, Producer[R, K, V]] = {
     val p: Task[KafkaProducer[Array[Byte], Array[Byte]]] = ZIO {
       val props = settings.driverSettings.asJava
       new KafkaProducer[Array[Byte], Array[Byte]](
--- a/src/main/scala/zio/kafka/client/Producer.scala
+++ b/src/main/scala/zio/kafka/client/Producer.scala
@@ -14,7 +14,33 @@ import scala.jdk.CollectionConverters._
 object Producer {
 
   trait Service[R, K, V] {
+
+    /**
+     * Produce a single record. The effect returned from this method has two layers and
+     * describes the completion of two actions:
+     * 1. The outer layer describes the enqueueing of the record to the Producer's internal
+     *    buffer.
+     * 2. The inner layer describes receiving an acknowledgement from the broker for the
+     *    transmission of the record.
+     *
+     * It is usually recommended to not await the inner layer of every individual record,
+     * but enqueue a batch of records and await all of their acknowledgements at once. That
+     * amortizes the cost of sending requests to Kafka and increases throughput.
+     */
     def produce(record: ProducerRecord[K, V]): RIO[Blocking with R, Task[RecordMetadata]]
+
+    /**
+     * Produces a chunk of records record. The effect returned from this method has two layers
+     * and describes the completion of two actions:
+     * 1. The outer layer describes the enqueueing of all the records to the Producer's
+     *    internal buffer.
+     * 2. The inner layer describes receiving an acknowledgement from the broker for the
+     *    transmission of the records.
+     *
+     * It is possible that for chunks that exceed the producer's internal buffer size, the
+     * outer layer will also signal the transmission of part of the chunk. Regardless,
+     * awaiting the inner layer guarantees the transmission of the entire chunk.
+     */
     def produceChunk(records: Chunk[ProducerRecord[K, V]]): RIO[R with Blocking, Task[Chunk[RecordMetadata]]]
 
     final def stream: ZSink[R with Blocking, Throwable, Nothing, Chunk[ProducerRecord[K, V]], Unit] =
@@ -47,18 +73,6 @@ object Producer {
             }
       } yield done.await
 
-    /**
-     * Produces a chunk of records record. The effect returned from this method has two layers
-     * and describes the completion of two actions:
-     * 1. The outer layer describes the enqueueing of all the records to the Producer's
-     *    internal buffer.
-     * 2. The inner layer describes receiving an acknowledgement from the broker for the
-     *    transmission of the records.
-     *
-     * It is possible that for chunks that exceed the producer's internal buffer size, the
-     * outer layer will also signal the transmission of part of the chunk. Regardless,
-     * awaiting the inner layer guarantees the transmission of the entire chunk.
-     */
     override def produceChunk(records: Chunk[ProducerRecord[K, V]]): RIO[R with Blocking, Task[Chunk[RecordMetadata]]] =
       if (records.isEmpty) ZIO.succeed(Task.succeed(Chunk.empty))
       else {
--- a/src/main/scala/zio/kafka/client/Producer.scala
+++ b/src/main/scala/zio/kafka/client/Producer.scala
@@ -2,6 +2,7 @@ package zio.kafka.client
 
 import java.util.concurrent.atomic.AtomicLong
 
+import izreflect.fundamentals.reflection.Tags.Tag
 import org.apache.kafka.clients.producer.{ Callback, KafkaProducer, ProducerRecord, RecordMetadata }
 import org.apache.kafka.common.serialization.ByteArraySerializer
 import zio._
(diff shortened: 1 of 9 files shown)


src/main/scala/zio/kafka/client/Producer.scala
```diff
@@ -11,126 +12,114 @@ import zio.stream.ZSink
...
-   * @return Producer as a Managed resource that will be closed automatically after use.
-   */
-  def make[R, K, V](
+  private def createLive[R, K, V](
+    p: KafkaProducer[Array[Byte], Array[Byte]],
+    keySerializer: Serializer[R, K],
+    valueSerializer: Serializer[R, V]
+  ): Live[R, K, V] = new Live(p, keySerializer, valueSerializer)
+
+  def producer[R: Tag, K: Tag, V: Tag](
```
adamgfraser: The intention was that you would be able to use `Tagged` here. Did that not work?

src/main/scala/zio/kafka/client/package.scala
```diff
@@ -18,4 +18,6 @@ package object client {
 
   type ByteArrayProducer       = KafkaProducer[Array[Byte], Array[Byte]]
   type ByteArrayProducerRecord = ProducerRecord[Array[Byte], Array[Byte]]
+
+  type HasProducer[R, K, V] = Has[Producer.Service[R, K, V]]
```
adamgfraser: I think normally this would be `type Producer[R, K, V] = Has[Producer.Service[R, K, V]]`.

src/main/scala/zio/kafka/client/package.scala
```diff
@@ -5,8 +5,8 @@ import java.util.{ Map => JMap }
 import org.apache.kafka.clients.consumer.OffsetAndMetadata
 import org.apache.kafka.clients.producer.{ KafkaProducer, ProducerRecord }
 import org.apache.kafka.common.TopicPartition
-
-import zio.ZIO
+import zio.{ Has, ZIO }
+import zio._
```
adamgfraser: Don't need the wildcard import if you are importing the specific types you are using.

src/main/scala/zio/kafka/client/Producer.scala
```diff
@@ -11,126 +12,114 @@ import zio.stream.ZSink
...
-   */
-  def produceChunk(records: Chunk[ProducerRecord[K, V]]): RIO[R with Blocking, Task[Chunk[RecordMetadata]]] =
-    if (records.isEmpty) ZIO.succeed(Task.succeed(Chunk.empty))
-    else {
+object Producer {
+
+  trait Service[R, K, V] {
+    def produce(record: ProducerRecord[K, V]): RIO[Blocking with R, Task[RecordMetadata]]
+    def produceChunk(records: Chunk[ProducerRecord[K, V]]): RIO[R with Blocking, Task[Chunk[RecordMetadata]]]
+    def stream: ZSink[R with Blocking, Throwable, Nothing, Chunk[ProducerRecord[K, V]], Unit] =
```
adamgfraser: `final`?

src/main/scala/zio/kafka/client/Producer.scala
```diff
@@ -2,6 +2,7 @@ package zio.kafka.client
 
 import java.util.concurrent.atomic.AtomicLong
 
+import izreflect.fundamentals.reflection.Tags.Tag
```
adamgfraser: I think this should be able to go away since you already have `Tagged` in scope through the wildcard import.

src/main/scala/zio/kafka/client/package.scala
kciesielski: Of course, I took that name initially and was about to change it later, but forgot. Thanks for pointing this out.

src/main/scala/zio/kafka/client/Producer.scala
kciesielski: I don't know exactly how `Tagged` works yet, but I'm getting
```
[error] Producer.scala:132:23: could not find implicit value for evidence parameter of type zio.Tagged[zio.kafka.client.Producer.Live[R,K,V]]
[error]     ZLayer.fromManaged(
```

adamgfraser: Hmmm, let me take a look.

adamgfraser: You should be able to do:

```scala
def producer[R, K, V](
  settings: ProducerSettings,
)(implicit tagged: Tagged[Service[R, K, V]]): ZLayer.NoDeps[Throwable, Producer[R, K, V]]
```

kciesielski: Thanks, looks like it did the job.

src/test/scala/zio/kafka/client/KafkaTestUtils.scala
```diff
@@ -16,27 +17,23 @@ import zio._
...
-
-  def produceOne(t: String, k: String, m: String) =
-    withProducerStrings { p =>
-      p.produce(new ProducerRecord(t, k, m))
-    }.flatten
+  type StringProducer = Producer[Any, String, String]
+
+  val testProducer: ZLayer[Kafka, Throwable, StringProducer] =
+    (for {
+      settings <- producerSettings.toManaged_
```
kciesielski: I have a feeling that this can be written in a simpler way, without converting the settings to a `Managed` and then back to `Layer.

src/test/scala/zio/kafka/client/KafkaTestUtils.scala
```diff
@@ -16,27 +17,23 @@ import zio._
...
+      producer <- Producer.producer[Any, String, String](settings, Serde.string, Serde.string).build
+    } yield producer).map(_.get).toLayer
+
+  def withProducerStrings[R: Tag, A: Tag](
+    r: Producer.Service[Any, String, String] => RIO[R, A]
+  ): ZIO[R with StringProducer, Throwable, A] =
+    ZLayer.fromService(r).build.use(_.get)
```
kciesielski: Similarly here, I think I should be able to express this without going through a `ZLayer`.

src/main/scala/zio/kafka/client/Producer.scala
```diff
@@ -11,126 +11,129 @@ import zio.stream.ZSink
...
+              }
+        } yield done.await
+      }
 
-object Producer {
+    /**
+     * Flushes the producer's internal buffer. This will guarantee that all records
+     * currently buffered will be transmitted to the broker.
+     */
+    def flush: BlockingTask[Unit] = effectBlocking(p.flush())
```
adamgfraser: Do you want this to be part of the API? If so I think you want to make it a method on `Service` otherwise you are going to be in an awkward position where it is only available on `Live` instead of `Service` but when you access through the environment you are just going to get `Service` so you are not going to be able to call `flush`.

src/main/scala/zio/kafka/client/Producer.scala
```diff
@@ -11,126 +11,129 @@ import zio.stream.ZSink
...
+     * outer layer will also signal the transmission of part of the chunk. Regardless,
+     * awaiting the inner layer guarantees the transmission of the entire chunk.
+     */
+    def produceChunk(records: Chunk[ProducerRecord[K, V]]): RIO[R with Blocking, Task[Chunk[RecordMetadata]]]
+
+    final def stream: ZSink[R with Blocking, Throwable, Nothing, Chunk[ProducerRecord[K, V]], Unit] =
+      ZSink.drain.contramapM(produceChunk)
+  }
+
+  class Live[R, K, V](
```
adamgfraser: `final case class`?

src/main/scala/zio/kafka/client/Producer.scala
```diff
@@ -11,126 +11,129 @@ import zio.stream.ZSink
...
-   * @return Producer as a Managed resource that will be closed automatically after use.
-   */
-  def make[R, K, V](
+  private def createLive[R, K, V](
+    p: KafkaProducer[Array[Byte], Array[Byte]],
+    keySerializer: Serializer[R, K],
+    valueSerializer: Serializer[R, V]
+  ): Live[R, K, V] = new Live(p, keySerializer, valueSerializer)
+
+  def producer[R, K, V](
```
adamgfraser: Maybe call this `make`?

src/main/scala/zio/kafka/client/Producer.scala
```diff
@@ -11,126 +11,129 @@ import zio.stream.ZSink
...
-   * @param settings Producer settings
-   * @param keySerializer Serializer for the record keys
-   * @param valueSerializer Serializer for the record values
-   * @tparam R Environment required by the serializers
-   * @tparam K Type of record keys
-   * @tparam V Type of record values
-   * @return Producer as a Managed resource that will be closed automatically after use.
-   */
-  def make[R, K, V](
+  private def createLive[R, K, V](
```
adamgfraser: Maybe inline this? You are only calling it one place in `make` and it is taking significantly more space just to right the method signature than the actual implementation.

src/main/scala/zio/kafka/client/Producer.scala
```diff
@@ -139,24 +142,9 @@ object Producer {
       )
     }
 
-    p.toManaged(p => UIO(p.close(settings.closeTimeout.asJava)))
-      .map(new Producer[R, K, V](_, keySerializer, valueSerializer))
+    ZLayer.fromManaged(
+      p.toManaged(p => UIO(p.close(settings.closeTimeout.asJava)))
+        .map(producer => createLive(producer, keySerializer, valueSerializer))
+    )
   }
```
adamgfraser: Normally at this point you would write your environmental accessors. For example:

```scala
def produce[R, K, V](record: ProducerRecord[R, K, V]): RIO[R with Producer with Blocking, Task[RecordMetadata]] =
  ZIO.accessM(_.get.produce(record))
```

This will dramatically improve the ergonomics of working with the producer in the environment.

adamgfraser: Normally at this point you would write your environmental accessors. For example:

```scala
def produce[R, K, V](record: ProducerRecord[R, K, V]): RIO[R with Producer with Blocking, Task[RecordMetadata]] =
  ZIO.accessM(_.get.produce(record))
```

This will dramatically improve the ergonomics of working with the producer in the environment.

src/main/scala/zio/kafka/client/Producer.scala
```diff
@@ -139,24 +142,9 @@ object Producer {
       )
     }
```
adamgfraser: It may not be possible given the degree of polymorphism here but are there any specialized "default" constructors that would provide sensible implementations of one or more of the parameters to `make` that would make it a little easier to create the producer layer for common cases?

src/test/scala/zio/kafka/client/KafkaTestUtils.scala
```diff
@@ -2,8 +2,9 @@ package zio.kafka.client
 
 import java.util.UUID
 
+import izreflect.fundamentals.reflection.Tags.Tag
```
adamgfraser: I think the `Tag` concept should be able to go away here as well in favor of `Tagged`.

src/main/scala/zio/kafka/client/Producer.scala
adamgfraser: It may not be possible given the degree of polymorphism here but are there any specialized "default" constructors that would provide sensible implementations of one or more of the parameters to `make` that would make it a little easier to create the producer layer for common cases?

src/test/scala/zio/kafka/client/KafkaTestUtils.scala
adamgfraser: I think this is indicating that we may want to express the dependencies of the producer as services as well:

```scala
object producer {
  def live[R, K, B]: ZLayer[Kafka with Serializer[R, K] with Serializer[R, V], Throwable, Consumer] =
    ???
}

object KafkaTestUtils {
  def testProducer: ZLayer[Kafka, Throwable, StringProducer] =
    (ZLayer.identity[Kafka] ++ ZLayer.succeed(Serde.string)) >>> producer.live
```

src/test/scala/zio/kafka/client/KafkaTestUtils.scala
adamgfraser: `ZIO.accessM(env => r(env.get))`?

src/test/scala/zio/kafka/client/KafkaTestUtils.scala
iravid: My intuition says that we shouldn't force users that want to use `ZLayer` to provide *everything* (incl. settings and serdes) as `R`, but maybe that's misguided.

At any rate, I'm fine with this being written as-is.

src/main/scala/zio/kafka/client/Producer.scala
iravid: Yes, let's keep it as `make`. That's the conventional name for effectful creation of things in ZIO.

src/main/scala/zio/kafka/client/Producer.scala
iravid: Right, flush should definitely be part of the API :-)

src/test/scala/zio/kafka/client/KafkaTestUtils.scala
adamgfraser: I think if we had a definition of `live` like the above we could then offer a method that would take the settings and series as normal parameters and lift them into services so we could provide the same user experience but have the layer be more composable internally.

iravid:
And thank you @adamgfraser for the comprehensive review!

src/test/scala/zio/kafka/client/KafkaTestUtils.scala
iravid: That's a good solution 👍🏻

src/main/scala/zio/kafka/client/Producer.scala
kciesielski: I thought of providing a version that takes String serializers without dependencies, like `makeSimpleStrings(settings: ProducerSettings)`, + maybe another one for `Array[Byte]`.

src/main/scala/zio/kafka/client/Producer.scala
```diff
@@ -11,152 +11,153 @@ import zio.stream.ZSink
...
     settings: ProducerSettings,
     keySerializer: Serializer[R, K],
     valueSerializer: Serializer[R, V]
-  ): ZManaged[Blocking, Throwable, ZSink[R with Blocking, Throwable, Nothing, Chunk[ProducerRecord[K, V]], Unit]] =
-    make[R, K, V](settings, keySerializer, valueSerializer).map { producer =>
-      ZSink.drain.contramapM(producer.produceChunk)
-    }
+  )(
+    implicit ts: Tagged[Serializer[R, K]],
+    td: Tagged[Serializer[R, V]],
```
kciesielski: @adamgfraser I can't seen to get rid of these implicit `Tagged` requirements here.

PR #129, opened 2020-02-28
Remove redundant 'isShutdown' check

svroonland:
This is already checked before the call to handlePoll


iravid:
Hmm, it is indeed checked before the call to handlePoll, but the point is to short-circuit all of the logic if we've been shutdown. The runloop will still be polling between the time that `gracefulShutdown` is called and the stream has actually ended.

The design has become a bit hairy though and is due for a bit of a refactor.

svroonland:
Makes sense, I missed that.

PR #161, opened 2020-04-27
'makeManaged' for creating ZManaged live services

LGLO:
This adds `makeManaged` functions returning `ZManaged[Any, Throwable, Service[R, K, V]`  for `Producer` and `Consumer`.

One place where it's convenient manifests in this PR in `Consumer#consumeWith`.
I've also found out such pattern in my own project:
```scala
  def live: ZLayer[Has[Config], Throwable, KafkaEventPublisher] = {
    ZLayer.fromServiceManaged[Config, Any, Throwable, Service] { cfg =>
      Producer.make(ProducerSettings(cfg.bootstrapHosts), Serializer.string, Serializer.string).build.map {
        hasProducer => new LivePublisher(hasProducer.get, cfg.topic)
      }
    }
  }
```
Going through `build`, `map` and `get` is redundant if `ZManaged` is exposed.
I didn't change `make` signature, I see such function also has it's usablity. I decided to add `makeManaged` instead.

### DIFF
--- a/src/main/scala/zio/kafka/consumer/package.scala
+++ b/src/main/scala/zio/kafka/consumer/package.scala
@@ -265,28 +265,30 @@ package object consumer {
       ZSink.foldLeft[Offset, OffsetBatch](OffsetBatch.empty)(_ merge _)
 
     def live: ZLayer[Clock with Blocking with Has[ConsumerSettings] with Has[Diagnostics], Throwable, Consumer] =
-      ZLayer.fromManaged {
-        ZManaged
-          .accessManaged[Clock with Blocking with Has[ConsumerSettings] with Has[Diagnostics]] { env =>
-            val settings = env.get[ConsumerSettings]
-            for {
-              wrapper <- ConsumerAccess.make(settings)
-              runloop <- Runloop(
-                          wrapper,
-                          settings.pollInterval,
-                          settings.pollTimeout,
-                          env.get[Diagnostics],
-                          settings.offsetRetrieval
-                        )
-            } yield Live(wrapper, settings, runloop)
-          }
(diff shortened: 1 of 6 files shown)


src/main/scala/zio/kafka/producer/package.scala
```diff
@@ -135,29 +135,32 @@ package object producer {
...
 
     def make[R: Tagged, K: Tagged, V: Tagged](
       settings: ProducerSettings,
       keySerializer: Serializer[R, K],
       valueSerializer: Serializer[R, V]
     ): ZLayer[Any, Throwable, Producer[R, K, V]] =
-      (ZLayer.succeed(settings) ++ ZLayer.succeed(keySerializer) ++ ZLayer.succeed(valueSerializer)) >>> live[R, K, V]
+      ZLayer.fromManaged(makeManaged(settings, keySerializer, valueSerializer))
+
+    def makeManaged[R: Tagged, K: Tagged, V: Tagged](
```
iravid, line 150: Let's replace the existing `def make` with your implementation. Also, this method doesn't need the tags:
```suggestion
    def makeManaged[R, K, V](
```

src/main/scala/zio/kafka/consumer/package.scala
```diff
@@ -265,28 +265,30 @@ package object consumer {
...
       }
 
     def make(
       settings: ConsumerSettings,
       diagnostics: Diagnostics = Diagnostics.NoOp
     ): ZLayer[Clock with Blocking, Throwable, Consumer] =
-      ((ZLayer.requires[Clock with Blocking] ++ ZLayer.succeed(settings) ++ ZLayer.succeed(diagnostics)) >>> live)
+      ZLayer.fromManaged(makeManaged(settings, diagnostics))
+
+    def makeManaged(
```
iravid, line 278: Same comment as in the producer - let's replace the existing `def make` with this implementation.

Issue #162, opened 2020-04-27
Injection of key and value serializers can be ambiguous


LGLO:
When I have `Producer[R, A, A]` or `Consumer[R, A, A]` but I want to use different serializers for keys and values `ZLayer` will register only value serializer and will use it to serialize keys.

I propose to introduce a class that encapsulates key and value serializers and use with ZLayer.

svroonland:
Note to future self: this now only applies to  `Producer[R, K, V]`. `Consumer`'s `Deserializer`s are currently passed as regular parameters. 

Also causes some issues on Scala 3: #317

svroonland:
This is fixed with #331

PR #163, opened 2020-04-28
Introduce 'Serializers' to avoid clash of key and value serializers.

LGLO:
This targets #162 I reported.
Scenario of failure is:
```scala
val keySer: Serializer[Any, String] = <some implementation>
val valueSer: Serializer[Any, String] = <other implementation>
Producer.make(someProducerSettings, keySer, valueSer)
```
Created producer will use `valueSer` for both keys and values.

### DIFF
--- a/src/main/scala/zio/kafka/producer/package.scala
+++ b/src/main/scala/zio/kafka/producer/package.scala
@@ -159,12 +159,11 @@ package object producer {
       private[producer] def close: UIO[Unit] = UIO(p.close(producerSettings.closeTimeout.asJava))
     }
 
-    def live[R: Tag, K: Tag, V: Tag]: ZLayer[Has[Serializer[R, K]] with Has[Serializer[R, V]] with Has[
-      ProducerSettings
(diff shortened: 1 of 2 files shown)


src/main/scala/zio/kafka/producer/package.scala
```diff
@@ -159,12 +159,11 @@ package object producer {
       private[producer] def close: UIO[Unit] = UIO(p.close(producerSettings.closeTimeout.asJava))
     }
 
-    def live[R: Tag, K: Tag, V: Tag]: ZLayer[Has[Serializer[R, K]] with Has[Serializer[R, V]] with Has[
-      ProducerSettings
-    ], Throwable, Producer[R, K, V]] =
+    def live[R: Tag, K: Tag, V: Tag]
```
svroonland, line 162: I wonder if the serializers should be a layer / environment at all or just a regular parameter to `live`, like they are in `make`. What's your opinion @iravid ?

PR #172, opened 2020-05-16
Add fuctions to serialize and deserialize to option

egast:
In Kafka topics, records can have null values. Records with null values have a special meaning namely a 'delete' of the record. For log compacted topics, records that have null value will be marked for deletion and will be deleted (eventually).

### DIFF
--- a/src/main/scala/zio/kafka/serde/Deserializer.scala
+++ b/src/main/scala/zio/kafka/serde/Deserializer.scala
@@ -1,10 +1,11 @@
 package zio.kafka.serde
 
-import zio.{ RIO, Task }
-
-import scala.util.{ Failure, Success, Try }
 import org.apache.kafka.common.header.Headers
 import org.apache.kafka.common.serialization.{ Deserializer => KafkaDeserializer }
+import zio.{ RIO, Task, ZIO }
+
+import scala.annotation.nowarn
+import scala.util.{ Failure, Success, Try }
 
 /**
  * Deserializer from byte array to a value of some type T
@@ -42,6 +43,9 @@ trait Deserializer[-R, +T] {
    */
   def asTry: Deserializer[R, Try[T]] =
     Deserializer(deserialize(_, _, _).fold(e => Failure(e), v => Success(v)))
+
+  def asOption(implicit @nowarn ev: T <:< AnyRef): Deserializer[R, Option[T]] =
+    Deserializer((topic, headers, data) => ZIO.foreach(Option(data))(deserialize(topic, headers, _)))
 }
 
 object Deserializer extends Serdes {
--- a/src/main/scala/zio/kafka/serde/Serde.scala
+++ b/src/main/scala/zio/kafka/serde/Serde.scala
@@ -25,6 +25,9 @@ trait Serde[-R, T] extends Deserializer[R, T] with Serializer[R, T] {
    */
   def inmapM[R1 <: R, U](f: T => RIO[R1, U])(g: U => RIO[R1, T]): Serde[R1, U] =
     Serde(mapM(f))(contramapM(g))
+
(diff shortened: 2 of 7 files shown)


src/main/scala/zio/kafka/serde/Deserializer.scala
```diff
@@ -42,6 +42,9 @@ trait Deserializer[-R, +T] {
    */
   def asTry: Deserializer[R, Try[T]] =
     Deserializer(deserialize(_, _, _).fold(e => Failure(e), v => Success(v)))
+
+  def asOptionDeserializer(implicit ev: T <:< AnyRef): Deserializer[R, Option[T]] =
```
svroonland, line 46: How about `asOption` for consistency with `asTry`

src/main/scala/zio/kafka/serde/Serializer.scala
```diff
@@ -24,6 +24,9 @@ trait Serializer[-R, -T] {
    */
   def contramapM[R1 <: R, U](f: U => RIO[R1, T]): Serializer[R1, U] =
     Serializer((topic, headers, u) => f(u).flatMap(serialize(topic, headers, _)))
+
+  def asOptionSerializer[U <: T](implicit ev: Null <:< T): Serializer[R, Option[U]] =
```
svroonland, line 28:
```suggestion
  def asOption[U <: T](implicit ev: Null <:< T): Serializer[R, Option[U]] =
```

src/main/scala/zio/kafka/serde/Deserializer.scala
egast: At first, I named this function `asOption`, but I also wanted an `asOption` function in the serializer but this would give a conflict in the Serde trait. So I decided to add the (De)serializer postfix to the function name. I also added a `asOption` function to the Serde trait to get a `Serde[R, Option[T]]`.

src/main/scala/zio/kafka/serde/Serializer.scala
egast: I have renamed the method to `asOption`. It doesn't seem to conflict with the other `asOption` methods :smile:

src/main/scala/zio/kafka/serde/Serde.scala
```diff
@@ -25,6 +25,9 @@ trait Serde[-R, T] extends Deserializer[R, T] with Serializer[R, T] {
    */
   def inmapM[R1 <: R, U](f: T => RIO[R1, U])(g: U => RIO[R1, T]): Serde[R1, U] =
     Serde(mapM(f))(contramapM(g))
+
+  def asOption(implicit ev: T <:< AnyRef, ev2: Null <:< T): Serde[R, Option[T]] =
+    Serde(super[Deserializer].asOption)(super[Serializer].asOption)
```
svroonland, line 30: Nice, I was not familiar with this construct.

src/main/scala/zio/kafka/serde/Deserializer.scala
```diff
@@ -42,6 +43,9 @@ trait Deserializer[-R, +T] {
    */
   def asTry: Deserializer[R, Try[T]] =
     Deserializer(deserialize(_, _, _).fold(e => Failure(e), v => Success(v)))
+
+  def asOption(implicit @nowarn ev: T <:< AnyRef): Deserializer[R, Option[T]] =
+    Deserializer((topic, headers, data) => deserialize(topic, headers, data).map(Option(_)))
```
iravid, line 48: Small safety suggestion:
```suggestion
    Deserializer((topic, headers, data) => ZIO.foreach(Option(data))(deserialize(topic, headers, _))
```

We can't assume that the underlying deserializer is null-safe.

egast: Good one! I will change this.

PR #173, opened 2020-05-17
Make predefined serdes in Serdes lazy vals to prevent dead-lock in test

egast:


iravid:
Does this solve the issue for you locally @egast?

egast:
Yes, it did. Thanks for your help.

iravid:
Thank you for debugging this!

PR #174, opened 2020-05-17
produce/produceAsync aliases

cipriansofronia:
This adds 2 aliases for `produce` and `produceAsync`, skipping the import for `ProducerRecord`


cipriansofronia:
Thanks @svroonland, could you also merge it?

iravid:
Thank you @cipriansofronia, and sorry for the delay here!

PR #186, opened 2020-06-05
Avoid the use of ZStream#mapM to preserve the chunks

iravid:


svroonland:
I can have a look at this later today. But don't wait if you're in a hurry to release.

iravid:
Thanks @svroonland! A retroactive review would be great :-)

iravid:
@svroonland I'd love to do that but it requires a bit more sophistication in the implementation. See here: https://github.com/zio/zio/issues/3660, https://github.com/zio/zio/pull/3666, https://github.com/zio/zio/issues/3668.

Let me know if you'd like to pick that up!

PR #220, opened 2020-09-27
Upgrade to kafka 2.6.0 and disable testing for Scala 2.11

svroonland:
Embedded kafka is no longer released for Scala 2.11


svroonland:
Well that doesn't make SBT work as I expected..

PR #224, opened 2020-10-05
Expose metrics for producer/consumer

tIGO:
Expose metrics for producer/consumer like [alpakka-kafka](https://doc.akka.io/docs/alpakka-kafka/current/producer.html#accessing-kafkaproducer-metrics) and [fs2-kafka](https://github.com/fd4s/fs2-kafka/blob/aaa82e23025df7f1ac03666f5bf05da1133f2c43/modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala#L325) do.


iravid (approved):
Thank @tIGO!

Issue #230, opened 2020-10-12
Simplify names of option parameters in AdminClient


svroonland:
Follow up from #201:
> On the other hand, however, please note that all other functions in this class use the same naming pattern for options (describeConfigsOptions, createPartitionsOptions, deleteTopicsOptions and 2 or 3 others) which seem to map 1:1 to Java client names, so there's the aspect of the naming consistency involved here as well, that perhaps should be considered as well.

regispl:
I can take a look at it as a followup from my PR.

How much are we concerned with backwards compatibility for APIs like this? What's ZIO's approach to that (generally)? 

Should we introduce a new optional parameter and mark the other one as "deprecated" for now with a plan to completely retire it later or is a simple "rename" just fine (with the understanding that clients will get broken, which means that this shouldn't probably be released right now but as a new `major` version at some point in future?)

svroonland:
That would be much welcome of course!

I suppose deprecation is the nice thing to do here, we can just `orElse` the `Option`s.

In general I would say that for `v < 1.0` there's a bit more room for breaking changes, after 1.0 it's generally appreciated that APIs don't have breaking changes. `zio` core has a binary compatibility check for this.

regispl:
Sure, makes perfect sense 👍

svroonland:
Implemented in #231

PR #201, opened 2020-07-04
Add support for listing and altering consumer offsets to the AdminClient

regispl:
As mentioned in Discord, I was playing a bit with the new functionality in the Kafka Admin Client added in 2.5 (listing and altering consumer offsets) and implemented the support for them to zio-kafka for fun.

One note: I defined enums using `sealed abstract class` - this is my personal preference (if I'm not using enumeratum) and I saw it's used in the code, but I also saw `sealed trait` being used elsewhere, so please let me know if you prefer me to change it.

PR #231, opened 2020-10-12
Rename options params in AdminClient

regispl:
Addresses #230. 

Once this is merged I can update #201 **or** I can wait for #201 to be merged first and fix it all in this PR - whatever suits @svroonland


svroonland:
I just merged #201 @regispl

src/main/scala/zio/kafka/admin/AdminClient.scala
```diff
@@ -42,13 +42,15 @@ case class AdminClient(private val adminClient: JAdminClient) {
    */
   def createTopics(
     newTopics: Iterable[NewTopic],
-    createTopicOptions: Option[CreateTopicsOptions] = None
+    createTopicOptions: Option[CreateTopicsOptions] = None,
```
svroonland, line 45: Can you add @deprecated annotations, since 0.13 (next release)?

regispl: Addressed by 9f81b48

regispl: Ah, deprecations are not silenced for some Scala versions, from what I can see, so the build is failing. 
I'll have a look at it later.

svroonland: Hmm I didn't expect the compiler to complain about the use of a deprecated parameter within the method itself..  Have to admit I never renamed a parameter in this way before.

svroonland: Perhaps this is of use? https://www.scala-lang.org/api/current/scala/deprecatedName.html

regispl: > Hmm I didn't expect the compiler to complain about the use of a deprecated parameter within the method itself.. Have to admit I never renamed a parameter in this way before.

Yes, I was a bit surprised too, but - similarly - have never used `deprecated` this way before.

> Perhaps this is of use? https://www.scala-lang.org/api/current/scala/deprecatedName.html

Thanks, will check it out. 

If this doesn't work I'll play a bit with `silent` - it should work for any statement, so I could limit the scope of the silencing enough to make it quite safe.

regispl: Looks like `deprecatedName` could work, but - ironically - it's older version of it is deprecated in Scala 2.13, while the newer one is not available in 2.12 yet, which is a bit of a pain

The main issue is, however, that using this annotation requires you to remove the old parameter - otherwise you'll get this error:
```
[error] /workspace/zio-kafka/src/main/scala/zio/kafka/admin/AdminClient.scala:173:83: deprecated parameter name alterConsumerGroupOffsetsOptions has to be distinct from any other parameter name (deprecated or not).
[error]     @deprecatedName(Symbol("alterConsumerGroupOffsetsOptions"), "zio-kafka 0.13") options: Option[
```

But when you remove it it's actually gone in Scala 2.12. There's no compile-time magic done by the annotation that would "bring it back" and make it available for the clients - it's just not there, which makes using this annotation a breaking change.

With this deprecation in place:
```scala
  def alterConsumerGroupOffsets(
    groupId: String,
    offsets: Map[TopicPartition, OffsetAndMetadata],
    @deprecatedName(Symbol("alterConsumerGroupOffsetsOptions"), "zio-kafka 0.13") options: Option[
      AlterConsumerGroupOffsetsOptions
    ] = None
  ): RIO[Blocking, Unit] { ... }
```

image

It works in Scala 2.13 though (as described in the doc) - sbt can compile the code from the screenshot (while IntelliJ complains at missing symbol...).

It looked promising, but it looks like making it work for all Scala versions could be quite tricky...

To the `silencer` then... 🤷‍♂️

regispl: Switched back to the `deprecated` and planned to have a look at the `silencer`, but fond this:
https://docs.oracle.com/javase/specs/jls/se7/html/jls-9.html#jls-9.6.3.6

> Use of the @Deprecated annotation on a local variable declaration **or on a parameter declaration** has no effect.

🤦‍♂️ 

So nope, this won't work either. Oh my...

I need to stop shaving this yak for today (looked like a pretty straightforward change, didn't it? 😆  ), but I'm not done 😉

svroonland: Ouch, that sounds a lot harder than it should be! Thanks already for taking the time to investigate how much fur this yak has.

So if I understand correctly:
* `deprecatedName` works fine in 2.13
* .. but in 2.12 you have to use it like `deprecatedName(Symbol("parameter"))` while that is deprecated in 2.13 for `deprecatedName("parameter")`.. jeez, they managed to make deprecation non-backwards compatible 🤷 

But it should be possible to silence that warning in 2.13.

This is a lot of effort for a parameter name change, so I suggest that if the silencer doesn't work we just rename it in a breaking manner.

cc @iravid

iravid:
I would wager that zio-kafka is mainly used in applications rather than in other libraries, which makes breaking changes a bit easier to deal with. So +1 for a breaking change.

src/main/scala/zio/kafka/admin/AdminClient.scala
regispl: > So if I understand correctly: (...)

Yes. On top of that please keep in mind that while both options compile just fine in both Scala versions (bar the deprecation warning for 2.13), they are not handled correctly by IntelliJ. I don't think it's an issue (not everyone uses IntelliJ, it's just an editor, it may push people a bit to upgrade their code etc.), but I'm just highlighting it here for "completeness".

> jeez, they managed to make deprecation non-backwards compatible

Yes, exactly that 😆 

OK, so later today I'll try to use the deprecated version of `deprecatedName` and silence it for 2.13 (don't expect any issues here, should work, but... I didn't expect any of the aforementioned issues either 😆  ). Then a breaking change as a last resort.

Thanks for chiming in @iravid ! 👍

regispl: I need to correct myself - it looks like using deprecated annotation behaves differently than using deprecated "regular" code, so using the old version of `deprecatedName` doesn't seem to trigger a warning that we'd need to suppress. I was quite sure I saw this happening yesterday, but it's perhaps something else that I did while testing one of the options listed above and got confused. 

In other words using the old `deprecatedName` is fine for 2.13 (it's deprecated, but no warning that could break the build unless it's the deprecated param name that's used), and works just fine in 2.12. 

Now I have an issue with 2.11 though, which does not define this annotation in the same way - it doesn't define the second parameter (`since`):
```scala
class deprecatedName(name: Symbol)
```
🤦‍♂️ 

I will remove the second param and leave the oldest version of `deprecatedName` that's compatible with Scala 2.11 and above. 

Hopefully it will be the last change 😉 

🐃  
(not a yak, but close enough...)

regispl:
OK, looks like all is good now. We need to use the oldest, "poorest" version of `deprecatedName`, but it seems to work for all Scala versions and no silencing is needed. I think it's the best we can get without having Scala-specific source directories, which would be a complete overkill in this case.

Let me know what you think @svroonland

svroonland:
Awesome! Looks good to go, thanks a lot @regispl!

svroonland (approved):

PR #246, opened 2020-11-23
Add note about jackson-databind to readme

svroonland:


iravid (approved):

iravid:
Such a weird move on the kafka-clients side.

PR #252, opened 2020-12-02
Don't return nulls in offsetsForTimes

jypma:
The Java Kafka API will put null values in the map returned by
offsetsForTimes, for partitions that don't have any messages in them
yet. This is the case, even the documentation only specifies
null-behavior for "old" message formats.

This commit filters out the null values, since using null in a scala Map
is very un-idiomatic, and would be surprise to most users.

There should be no loss of information, since the caller has the
complete list of partitions in the method argument.


src/main/scala/zio/kafka/consumer/package.scala
```diff
@@ -101,6 +101,13 @@ package object consumer {
 
       def unsubscribe: RIO[Blocking, Unit]
 
+      /**
+       * Look up the offsets for the given partitions by timestamp. The returned offset for each partition is the
+       * earliest offset whose timestamp is greater than or equal to the given timestamp in the corresponding partition.
+       *
+       * The consumer does not have to be assigned the partitions.
```
svroonland, line 108: Alternatively we could return `Map[TopicPartition, Option[OffsetAndTimestamp]]`. Not sure which is better

jypma: As far as I've understood Kafka's documentation (and observed behaviour), the returned map (from kafka) will always have all requested `TopicPartitions` as keys. So, we're not losing anything by simply not having the items present.

However, typically one would want a default value for missing timestamps, which might be slightly easier with `Option[OffsetAndTimestamp]` (allowing `_.map(_.getOrElse(defaultValue))`. However it'd change the return type.

svroonland (approved):
Thanks for the contribution, sorry for the late approval!

iravid (approved):

jypma:
Thanks a lot for the merge! :)

PR #296, opened 2021-04-09
Fix NullPointerException in 'partitionsFor' when Topic is not created

aartigao:
Today we've found that `NullPointerException` is raised when trying to resolve partitions for a particular topic.

Turns out if `allow.auto.create.topics` is set and topic is not created yet, it creates one with default values, but in our organization this flag is disabled.

This PR fixes that by checking nullability and also fixes some SBT warnings about slash syntax.


iravid (approved):
Thank you!

PR #297, opened 2021-04-10
Remove listener for pausing partitions on rebalance revoke

aartigao:
As discussed in Discord channel, seems that revoke callback implies that assigned partitions are no longer available so pausing them may be redundant.

I really don't know the implications of this change, so expert advice is welcomed 😅

### DIFF
--- a/src/main/scala/zio/kafka/consumer/internal/RebalanceListener.scala
+++ b/src/main/scala/zio/kafka/consumer/internal/RebalanceListener.scala
@@ -15,7 +15,7 @@ private[consumer] final case class RebalanceListener(
   /**
    * Combine with another [[RebalanceListener]] and execute their actions sequentially
    */
-  def ++(that: RebalanceListener) =
+  def ++(that: RebalanceListener): RebalanceListener =
     RebalanceListener(
       assigned => onAssigned(assigned) *> that.onAssigned(assigned),
       revoked => onRevoked(revoked) *> that.onRevoked(revoked)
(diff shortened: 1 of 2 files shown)


src/main/scala/zio/kafka/consumer/internal/RebalanceListener.scala
```diff
@@ -35,8 +35,3 @@ private[consumer] final case class RebalanceListener(
     }
 
 }
-
-object RebalanceListener {
-  def onRevoked(action: Set[TopicPartition] => Task[Unit]): RebalanceListener =
```
svroonland, line 40: I would keep this one, since it's also a useful utility method for users creating custom `RebalanceListener`s.

svroonland (approved):
Looks good to me

iravid:
Hold on merging please

src/main/scala/zio/kafka/consumer/internal/RebalanceListener.scala
aartigao: I've removed it because it's in `internal` package and I've supposed that it's not open to end users. 

Shouldn't a user that wants to be notified about revoked partitions use the `DiagnosticEvent Rebalance.Revoked`?

svroonland: You're right of course, I overlooked that.

aartigao:
@iravid do you know if this can be safely merged now? 😄

svroonland:
Let's merge, I assume @iravid just wanted to take a look before merging. 

Is especially redundant with #323

svroonland:
Woops, wrong button. Will merge when build is green again.

svroonland (approved):

PR #305, opened 2021-05-10
Separated plainstream

jkobejs:


svroonland:
Hi @jkobejs, could you give a bit of context on your PR? What is the functionality it implements or issue it solves?

iravid:
This is something I collaborated on with @jkobejs a few weeks ago. The main goal is to simplify the non-partitioned stream functionality but I think we went a little out-of-scope.

I think we will re-focus this on implementing a simplified version of the runloop for streams that don't care to differentiate between partitions.

PR #307, opened 2021-05-12
Convert Admin Client to use Scala case classes

agarella:


iravid (dismissed):
Awesome. Thanks @agarella!

iravid:
@agarella Could you fix the compilation errors? Will merge afterwards.

agarella:
@iravid should be good now

iravid (approved):
🚀

PR #311, opened 2021-05-15
Removes Blocking environment requirement on most service methods

sebver:
Since the use of `Blocking` is related to the underlying implementation, I've removed it from the environment on most methods and added it to the `ZLayer` and `ZManaged` definitions.

There were also a few cases where there was a `Clock` requirement that wasn't necessary (probably refactoring leftovers), I removed those as well.


iravid:
Looks great @sebver. Could you resolve the conflicts? Will merge afterwards.

sebver:
Thanks, done.

[Master build is currently failing though](https://github.com/zio/zio-kafka/runs/2592069175) (so this one as well). That should probably be fixed in a separate PR?

iravid (approved):

PR #312, opened 2021-05-16
Remove call to deprecated Metric.value() method

agarella:


iravid (approved):
Weird that it's AnyRef but it is what it is.

PR #317, opened 2021-05-25
Scala 3

987Nabil:
Try my hands on cross build for Scala 3

Side node: This is my first pr to any zio or open source project ;)


svroonland:
Hi @987Nabil, thanks for the contribution! I enabled github action runs, so you can check the build status on all scala+jdk versions.

svroonland:
There's a few warnings and some errors that look like they can easily be fixed, would you be willing to fix those @987Nabil ?

987Nabil:
Sure. I'll try to do it tomorrow

987Nabil:
@svroonland I guess it should be fine now :)

svroonland:
Great, thanks! It just needs a run of `sbt scalafmtSbt` to fix some formatting and there's an error about the `kind-projector` flag on scala 3. I can't push to your branch myself to fix it, unfortunately.

987Nabil:
Hmm strange. The `kind-projector` flag seems to be already set. But I can't figure out by what. Strange that it did not popup when I tested locally the last time 🤔 
I did run the formatting and testet again. Hopefully the last time needed ;)

987Nabil:
@svroonland Seems like there is a type inference problem. I'll try to find it, but the stack trace does not seem to be very helpful to find the actual position of the problem. But I guess, some type needs to be implicit instead of inferred. Anyway, I assume that is a Scala 3 bug. There are still some problems with inferring.

svroonland:
Yeah, I think we're running into this one: https://github.com/zio/zio/issues/4564

987Nabil:
I am not sure, that this is the Problem.

I tried the suggested fix to the place where I think it comes from (`producer/package.scala`)

current state:

```scala
def live[R: Tag, K: Tag, V: Tag]: RLayer[Has[Serializer[R, K]] with Has[Serializer[R, V]] with Has[
  ProducerSettings
] with Blocking, Producer[R, K, V]] =
  (for {
    keySerializer   <- ZManaged.service[Serializer[R, K]]
    valueSerializer <- ZManaged.service[Serializer[R, V]]
    settings        <- ZManaged.service[ProducerSettings]
    producer        <- make(settings, keySerializer, valueSerializer)
  } yield producer).toLayer
```


maybe fix:
```scala
def live[R, K, V](implicit //I tried it too with :Tag
  rk: Tag[Has[Serializer[R, K]]],
  rv: Tag[Has[Serializer[R, V]]]
): RLayer[Has[Serializer[R, K]] with Has[Serializer[R, V]] with Has[
  ProducerSettings
] with Blocking, Producer[R, K, V]] =
  (for {
    keySerializer   <- ZManaged.service[Serializer[R, K]]
    valueSerializer <- ZManaged.service[Serializer[R, V]]
    settings        <- ZManaged.service[ProducerSettings]
    producer        <- make(settings, keySerializer, valueSerializer)
  } yield producer).toLayer
```

But it does not help.

I also don't fully understand the type he tries to find. `Serializer[-_$R|<Nothing..Any>,-_$K|<Nothing..Any>]`
`Serializer[-R, -T]` has `R` and `T`, but here it is `R` and `K`, which are the type parameter from the calling side. But what does then the `|<Nothing..Any>` mean? The pipe means union type? And what is `<Nothing..Any>`? Any type between Nothing and Any?

iravid:
I think we can reboot this on top of the modified service definitions as they are now monomorphic and will probably cause less trouble.

987Nabil:
@iravid Yes, after merging the master things seem to work fine now.

build.sbt
```diff
@@ -20,6 +21,12 @@ lazy val silencer = {
   )
 }
 
+lazy val kindProjector = Seq(
```
adamgfraser, line 24: Is this necessary?

987Nabil: I assume you mean the plugin at all.  I removed it any everything compiles and tests are running for all Scala versions. So I guess no. I removed it.

adamgfraser: Yes exactly.

iravid (approved):
Thank you @987Nabil and congratulations on your first pull request!

987Nabil:
@iravid I realised, that there is still no maven artifact for for Scala 3 yet. Is it possible to trigger a Scala 3 relese somehow for 0.15(.x)?

svroonland:
@987Nabil I was planning to make a 0.16 release (minor since the API is not compatible), but I wanted to write some migration notes first. I'll try to do that in the coming days

PR #323, opened 2021-06-03
Eager interrupt for revoked partitions' streams

### LINKED ISSUE #280: Rebalance causes some partition streams to be duplicated
When a rebalance happens, there is a chance that some partitions that are revoked and immediately re-assigned to the same consumer get duplicated (all messages received twice in the consumer stream).

Here's how to reproduce the problem (my code is not open source):

First add those 2 println in `newPartitionStream` when a partition stream is created then ended:
```scala
val stream = {
      ZStream.fromEffect(UIO(println(s"Start stream for $tp"))) *>
        ZStream {
          ZManaged.succeed {
            for {
              p      <- Promise.make[Option[Throwable], Chunk[ByteArrayCommittableRecord]]
              _      <- requestQueue.offer(Runloop.Request(tp, p)).unit
              _      <- diagnostics.emitIfEnabled(DiagnosticEvent.Request(tp))
              result <- p.await
            } yield result
          }
        }.ensuring(UIO(println(s"Finish stream for $tp")))
    }
```

1. Create a topic with 300 partitions (more partitions = maximize the chance of running into the bug)
2. Start a producer that produces a message every 20ms
3. Start a node A consuming this topic => we see 300x `Start stream for ...` :white_check_mark: 
4. Start a node B consuming this topic (with the same consumer group) => we see 150x `Start stream for ...` on node B and 150x `Finish stream for ...` on node A :white_check_mark: 
5. Stop B => we see 15x (this number vary, but it's not 150 as it should) `Finish stream for ...` on node A followed by 300x `Start stream for ...` on node A :warning: 

This seems to happen because `endRevoked` only stops partition streams that have a pending `Request`, which is not all of them. In normal cases, this is okay because those other streams are stopped by `handleRequests` when a new request is created and it realizes the partition is no longer assigned. But, in this particular case of rebalance, the partitions are reassigned right after being revoked so `handleRequests` doesn't help. Instead, a new partition stream is created and the old one keeps running too.

One way to solve this might be to defer the revoke/assign logic while we're in the middle of rebalancing? Not sure if it's safe enough though.

aartigao:
Fixes #280

Keep track of assigned streams and eagerly stop them by interruption when a rebalance revokes the associated partition assignation.


aartigao:
I've created this draft PR to discuss the implementation.

I've tried to follow @iravid instructions on #280 but any feedback is welcomed 🙏🏽 

For the moment, only the implementation is done (test is missing).

iravid:
Hey @aartigao, thank you so much for picking this up.

I'll check this out locally and review now. If I am not mistaken, the test-case was ported from fs2-kafka. If that is so, could you please add attribution?

Will be back with feedback soon.

src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -295,9 +318,10 @@ private[consumer] final class Runloop(
                            doSeekForNewPartitions(c, newlyAssigned) *> endRevoked(
                              state.pendingRequests,
                              state.addBufferedRecords(unrequestedRecords).bufferedRecords,
+                             state.assignedStreams,
                              revoked(_)
```
iravid, line 322: I think it would slightly simplify the code to stop tracking `revoked` and just revoke everything that isn't in the current assignment. You can do that now that you have the map of running streams.

iravid (commented):
Great work. Left some comments.

I have another concern around rebalances that take two polls to complete; when this happens, we will see the revocation callback called without the assignment callback called. I'll read the code again to see if there's any trouble that will happen in that case, would be great if you could think about that as well!

src/test/scala/zio/kafka/ConsumerSpec.scala
```diff
@@ -479,7 +480,102 @@ object ConsumerSpec extends DefaultRunnableSpec {
...
+          _ <- check(Set(0, 1, 2))
+          _ <- fiber2.interrupt
+          _ <- allAssignments.update(_ - 2)
+          _ <- check(Set(0, 1))
+          _ <- fiber1.interrupt
+          _ <- allAssignments.update(_ - 1)
+          _ <- check(Set(0))
+          _ <- fiber0.interrupt
+        } yield assertCompletes
+      } @@ flaky(5)
```
iravid, line 578: This being flaky is a bit concerning. Should we harden the test? Or does that mean that the fix is not perfect?

aartigao:
> If I am not mistaken, the test-case was ported from fs2-kafka. If that is so, could you please add attribution?

Sure. Where do I find the original source? Do you have the link or maybe @narma has it?

src/test/scala/zio/kafka/ConsumerSpec.scala
aartigao: It's something that I keep investigating. On the surface both implementation and test look right to me but definitively something is wrong.

src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -126,34 +126,43 @@ private[consumer] final class Runloop(
...
     val buf = mutable.Map[TopicPartition, Chunk[ByteArrayConsumerRecord]]()
     buf ++= bufferedRecords
 
-    var revokeAction: UIO[_] = UIO.unit
+    val (revokedStreams, assignedStreams) = currentAssignedStreams.partition(es => revoked(es._1))
+
+    val revokeAction: UIO[Unit] =
+      revokedStreams.foldLeft(UIO.unit) {
+        case (revokeAction, (_, p)) => revokeAction *> p.succeed(()).unit
+      }
```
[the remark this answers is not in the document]
aartigao: Yup! 🙏🏽

src/main/scala/zio/kafka/consumer/internal/Runloop.scala
aartigao: Do you mean get rid of `revoked: TopicPartition => Boolean` and instead use directly `val revoked: Set[TopicPartition]`?

src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -126,34 +126,43 @@ private[consumer] final class Runloop(
...
    * @return New pending requests and new buffered records
    */
   private def endRevoked(
     reqs: List[Runloop.Request],
     bufferedRecords: Map[TopicPartition, Chunk[ByteArrayConsumerRecord]],
+    currentAssignedStreams: Map[TopicPartition, Promise[Throwable, Unit]],
     revoked: TopicPartition => Boolean
   ): UIO[
-    (List[Runloop.Request], Map[TopicPartition, Chunk[ByteArrayConsumerRecord]])
+    (
```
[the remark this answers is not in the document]
aartigao: Seems right 👍🏽. Maybe we can reuse `State` for that?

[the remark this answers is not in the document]
aartigao: In fact I'm thinking about refactoring method signatures for all involved methods in call chain like `bufferRecordsForUnrequestedPartitions`, `fulfillRequests`, `endRevoked` to accept `State` as argument and return new `State` as result. I think it'll be more readable. WDYT?

aartigao:
> I have another concern around rebalances that take two polls to complete; when this happens, we will see the revocation callback called without the assignment callback called. I'll read the code again to see if there's any trouble that will happen in that case, would be great if you could think about that as well!

Thinking out loud, but if rebalance occurs in the middle of a poll, what will happen if we keep the same state until the next _safe_ poll?

src/test/scala/zio/kafka/ConsumerSpec.scala
narma: Guys, I've found the issue in this test case - it have a race condition between adding and removing partitions in `run` at allAssignments updates. 

 Because removing partition when stream stopped can be happens after when new partition stream was attached and allAssignments state was updated. Probably fix is re-add partition on partitionStream message handler. 

Original fs2-kafka test doesn't have this issue because it relies on different consumer method - partitionMapStream which produces whole batch of  assignments and streams.

src/test/scala/zio/kafka/ConsumerSpec.scala
```diff
@@ -479,7 +480,102 @@ object ConsumerSpec extends DefaultRunnableSpec {
...
+                  }
+                }))
+            }
+            .flattenParUnbounded()
+            .runDrain
+        }
+
+        def checkAssignments(allAssignments: Ref[Map[Int, ListBuffer[Int]]])(instances: Set[Int]) =
+          ZStream
+            .repeatEffectWith(allAssignments.get, Schedule.spaced(30.millis))
```
narma, line 520: `SubscriptionRef` did not exist in ZIO when I ported the test, with `SubscriptionRef` it will be easier, I think.

src/test/scala/zio/kafka/ConsumerSpec.scala
aartigao: I not sure if it's the same issue but I slightly modified the gist ported test case to track partitions as `ListBuffer` instead of `Set` just for this reason. Is the same thing or I'm talking about something different?

narma: Yeah, it's the same issue that I talking about. It a little bit complicated to track assigned (and running) partitions in that way.
For example with the current implementation in next case
1. Kafka assign partition 1 and then revoke partition 1
2. In test first happens is removing partition 1
3. Then adding partition 1
4. In result we have inconsistent state in allAssigments which says as we have partition 1.

aartigao: Shouldn't that be covered by the fact that partition tracking is increased before actually start processing the stream, and decreased after stream is interrupted?

```
case (tp, partStream) =>
                ZStream.fromEffect(allAssignments.update({ current =>
                  current.get(instance) match {
                    case Some(currentList) => current.updated(instance, currentList :+ tp.partition())
                    case None              => current.updated(instance, ListBuffer(tp.partition()))
                  }
                })) ++ partStream.fixed(10.millis) ++ ZStream.fromEffect(allAssignments.update({ current =>
                  current.get(instance) match {
                    case Some(currentList) =>
                      val idx = currentList.indexOf(tp.partition())
                      if (idx != -1) currentList.remove(idx)
                      current
                    case None => current
                  }
                }))
```

narma: Nope, generally issue is
newPartitionStream will emit a new stream, and if we have previously attached stream (previous one) we don't have a guarantee order here. Revoke effect for updating state here for the previous stream can be fired after effect on attaching the new one.

aartigao: Exactly, that's why original snippet was a `Set` but current one is `ListBuffer`. You might end up at some point in time the list buffer with duplicated partitions but eventually it'll be balanced when old stream finishes.

aartigao: I have to learn the same lesson with 🔥 Don't play with mutability 🤦🏽 

Flaky test is fixed removing `ListBuffer` and using a plain `List`. I guess something was broken with atomic `Ref`s but... anyway, I ran with `@nonFlaky(10)` locally and worked well.

narma: 🎉  Got it, looks like it covers the issue :+1:

aartigao:
I've added a last commit that fixes what @ghostdogpr said about `null` records. This way we can ensure that partition revocation is executed always, no matter if underlying `poll` fails or not.

iravid:
@aartigao Rebalances can take a while to complete. So what would happen is that on the first poll, onRevoke is called. The assignment is cleared. On a second (or perhaps on a 3rd or 4th or more) call to poll, only then is onAssign called.

aartigao:
> @aartigao Rebalances can take a while to complete. So what would happen is that on the first poll, onRevoke is called. The assignment is cleared. On a second (or perhaps on a 3rd or 4th or more) call to poll, only then is onAssign called.

@iravid By reading the Kafka API it's not clear to me: [ConsumerRebalanceListener](https://kafka.apache.org/26/javadoc/org/apache/kafka/clients/consumer/ConsumerRebalanceListener.html)

From what I understand both `onRevoke` and `onAssign` are called within a `poll` call, but shouldn't they be called within the same `poll` call? 🤔 

From what I see in our logs if an instance has, for example, partitions 0,1,2,3 assigned and a new instance joins Consumer Group, we see:
- `Instance A` revokes all 0,1,2,3 partitions
- `Instance A` get assigned 0,1 partitions
- `Instance B` get assigned 2,3 partitions

What you say is that `Instance A` revoke occur within 1st `poll` and `Instance A` assign occur in another subsequent `poll` (be it 2nd, 3rd, etc...). This is where I doubt, because seems weird to me, for me `poll` should return when rebalance has finished. The docs for `poll` state:

 ```This method returns immediately if there are records available. Otherwise, it will await the passed timeout. If the timeout expires, an empty record set will be returned. Note that this method may block beyond the timeout in order to execute custom ConsumerRebalanceListener callbacks.```

Maybe it's the way it is in Kafka, but sounds weird that `poll` returns control to user in the middle of a rebalance 😕

iravid:
Yeah I don't think there's a guarantee that they are both called on the same poll call. This is because in the rebalance protocol, there's an arbitrarily long duration (which is unrelated to the timeout given to poll) between the start of the rebalance (which causes revocation) and its end (the assignment).

src/main/scala/zio/kafka/consumer/internal/Runloop.scala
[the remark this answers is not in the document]
svroonland: My 2c: sounds good, but I think it would be good to focus on just the fix in this PR. It can be hard to see the actual behavioral changes in PRs that include refactorings.

aartigao:
Maybe then we should skip the whole `handlePoll` if it's still rebalancing?

src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -126,34 +126,40 @@ private[consumer] final class Runloop(
 
   /**
    * Does all needed to end revoked partitions:
-   * 1. Fail the Requests's continuation promises
+   * 1. Complete the emitted streams continuation promises
    * 2. Remove from the list of pending requets
    * 3. Remove from buffered records
    * @return New pending requests and new buffered records
```
svroonland, line 132: Comment about the return value should be updated or removed

svroonland (approved):
Nice work, implementation looks correct to me! Two nitpicks

svroonland:
> Maybe then we should skip the whole handlePoll if it's still rebalancing?
@aartigao I'm not 100% sure, but I believe polling is how the consumer communicaties about the rebalancing process.

aartigao:
> > Maybe then we should skip the whole handlePoll if it's still rebalancing?
> > @aartigao I'm not 100% sure, but I believe polling is how the consumer communicaties about the rebalancing process.

True 😳

src/main/scala/zio/kafka/consumer/internal/Runloop.scala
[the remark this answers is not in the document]
aartigao: You're right, let's leave it for another refactor PR 👍🏽

src/main/scala/zio/kafka/consumer/internal/Runloop.scala
aartigao: Fixed 🔧

aartigao:
Fixed all conflicts with `master` branch 👍🏽  🎉

src/main/scala/zio/kafka/consumer/internal/Runloop.scala
svroonland: I think he means that
```scala
                           val revoked         = prevAssigned -- currentAssigned
```
is no longer necessary if we just pass `currentAssigned` to `endRevoked` and calculate the diff with the keys of  `state.assignedStreams`.

svroonland:
Good to go for me! Let's merge it and we can always tackle minor improvements in a next PR.

svroonland (approved):

PR #325, opened 2021-06-05
Replace RequestBuffer with ZQueue

aartigao:
Now that #298 is merged, I can open the discussion for this change that I had in my backpack for some time.

Are there any benefits from using `RequestBuffer` instead of `ZQueue`? Performance reasons? Or maybe it's something that has been in the codebase since initial implementation and now that `ZQueue` is more mature it's safe to use it?


svroonland:
Interesting. I think it indeed stems from the early zio days. It says in the comments:
> Will be replaced once zio.Queue is enhanced with a `takeUpTo` method.

and I see no reason not to replace it. ZQueue will likely have better performance with its ring buffer structure.

PR welcome of course!

svroonland (approved):

PR #329, opened 2021-06-05
Make `consumer` available to OffsetRetrieval.Manual's getOffsets.

ithinkicancode:
Taking a crack at this.

@jypma @aartigao @svroonland, I added `ConsumerAccess` to the signature of `getOffsets` in `OffsetRetrieval.Manual`, as in:
`getOffsets: (Set[TopicPartition], ConsumerAccess) => Task[Map[TopicPartition, Long]]`

A couple of concerns:
1. I'm not sure if it's kosher to expose `consumer` to userland.
2. Users would probably have to duplicate the implementations such as `endOffsets` and `offsetsForTimes` of `Consumer.Live`. It's not hard but it's inconvenient.

Please have a look at this draft and let me know your thoughts.

### DIFF
--- a/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -222,7 +222,7 @@ private[consumer] final class Runloop(
   private def doSeekForNewPartitions(c: ByteArrayKafkaConsumer, tps: Set[TopicPartition]): Task[Unit] =
     offsetRetrieval match {
       case OffsetRetrieval.Manual(getOffsets) =>
-        getOffsets(tps)
+        getOffsets(tps, consumer)
           .tap(offsets => ZIO.foreach_(offsets) { case (tp, offset) => ZIO(c.seek(tp, offset)) })
           .when(tps.nonEmpty)
 
--- a/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -8,7 +8,7 @@ import zio._
 import zio.blocking.Blocking
 import zio.clock.Clock
 import zio.duration._
-import zio.kafka.consumer.Consumer.OffsetRetrieval
+import zio.kafka.consumer.Consumer.{ ConsumerForAdminAccess, OffsetRetrieval }
 import zio.kafka.consumer.diagnostics.{ DiagnosticEvent, Diagnostics }
 import zio.kafka.consumer.CommittableRecord
 import zio.kafka.consumer.internal.ConsumerAccess.ByteArrayKafkaConsumer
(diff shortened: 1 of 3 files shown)


aartigao:
I see your concerns and I agree that `ConsumerAccess` is an internal implementation that shouldn't be exposed to final users.

An ideal solution for me will be to change the signature `Set[TopicPartition] => Task[Map[TopicPartition, Long]]` to `Set[TopicPartition] => RIO[Consumer, Map[TopicPartition, Long]]` but I'm afraid this is difficult (if not impossible) because that would mean you need a `Consumer` to create a `Consumer` 😄 

I'll try to think a solution and if I come up with something I'll tell you 👍🏽

ithinkicancode:
@aartigao I added a wrapper class, ConsumerForAdminAccess, to limit what users can do with `ConsumerAccess`, although it still requires an instance of `ConsumerAccess`.

aartigao:
Honestly, I don't have a clear opinion about that 😳 It solves the problem, yes, but still there's something that doesn't sound to me 😕 

I want to hear what @svroonland or @iravid think about it...

svroonland:
It solves the problem indeed, but it doesn't quite feel like it's the right design, with a part of the consumer code now copied. I don't have a good alternative yet though. 

Ideally I'd see a `getOffsets: RIO[Consumer.Service, Map[TopicPartition, Long]]`, that should be possible I think.

iravid:
I have some thoughts which I will write down shortly.

iravid:
First, I would like to note that the AdminClient offers support for fetching offsets by timestamp through the `listOffsets` method. The `OffsetSpec` allows asking for the offsets by time. With this said, do we actually need this change? :-) I am fairly sure that everything you'd be able to access through the consumer can be accessed through the admin client.

Beyond that, creating a facade of the consumer's interface that contains the subset of read-only methods for the purpose of supplying it to OffsetRetrieval.Manual seems reasonable to me. The only tricky part is that there is an implicit assumption here that by the time `getOffsets` is called, the consumer is already subscribed (in case someone wants to look at the assignment or anything relating to the subscription state).

I would also consider creating `OffsetRetrieval.ByTimestamp` for the purposes of seeking the offsets with `offsetsForTimes` but like I said, this can be done through the AdminClient.

aartigao:
Creating `OffsetRetrieval.ByTimestamp` indeed will be really useful and other libraries have this feature. 

It'll give @jypma the feature needed. But still, don't cover more advanced use cases like the one we are using. Anyway, from my part I'll be happy to see `OffsetRetrieval.ByTimestamp` implemented.

In fact, given that we can extend `OffetRetrieval` trait, I'll create a proposal for seeking at last message offset (not last offset, but last offset - 1, to start consuming last already created record). It's something that we use in our code for some patterns but didn't have a direct translation in Kafka Consumer features. But let's discuss it in another place 🙂

iravid:
@aartigao Do you actually need access to the consumer given what the admin client gives you?

aartigao:
> @aartigao Do you actually need access to the consumer given what the admin client gives you?

TBH, now I've revisited my code and _I think_ that the only missing part was getting the offsets for a given Group ID 🤔 But it's weird because I've this logic implemented in a local branch and decided to not create a PR for that because it wasn't enough... 🤷🏽

Anyway, I'll submit a PR with the code to get offsets by Group ID in `AdminClient`. Now, I think it's the only missing part... 😅

ithinkicancode:
I'm closing this draft as https://github.com/zio/zio-kafka/pull/330 superseded this.

PR #330, opened 2021-06-07
Add `listConsumerGroupOffsets` to AdminClient

aartigao:
As commented in #329, added `listConsumerGroupOffsets` to AdminClient.

This will allow more complex use cases for getting manual offsets if logic involves getting the offsets for a given Consumer Group.


iravid (approved):
Thank you!

PR #334, opened 2021-06-18
Move configuration of serdes to the creation effect

iravid:
I noticed that we were running `Deserializer#configure` in the consumer but nowhere in the producer. This was quite inconsistent so I figured it would make more sense for `configure` to not be present on our traits, and it would be run when converting a kafka-clients De/serializer in the construction effect.

Appreciate comments from anyone using de/serializers from Kafka - particularly the Confluent ones for Avro.

### DIFF
--- a/src/main/scala/zio/kafka/serde/Deserializer.scala
+++ b/src/main/scala/zio/kafka/serde/Deserializer.scala
@@ -3,6 +3,7 @@ package zio.kafka.serde
 import org.apache.kafka.common.header.Headers
 import org.apache.kafka.common.serialization.{ Deserializer => KafkaDeserializer }
 import zio.{ RIO, Task, ZIO }
+import zio.blocking.{ blocking => zioBlocking, Blocking }
 
 import scala.util.{ Failure, Success, Try }
 import scala.jdk.CollectionConverters._
@@ -18,7 +19,12 @@ import scala.annotation.nowarn
  */
 trait Deserializer[-R, +T] {
   def deserialize(topic: String, headers: Headers, data: Array[Byte]): RIO[R, T]
-  def configure(props: Map[String, AnyRef], isKey: Boolean): Task[Unit]
+
+  /**
+   * Returns a new deserializer that executes its deserialization function on the blocking threadpool.
+   */
+  def blocking: Deserializer[R with Blocking, T] =
+    Deserializer((topic, headers, data) => zioBlocking(deserialize(topic, headers, data)))
 
   /**
    * Create a deserializer for a type U based on the deserializer for type T and a mapping function
@@ -48,6 +54,9 @@ trait Deserializer[-R, +T] {
   def asTry: Deserializer[R, Try[T]] =
     Deserializer(deserialize(_, _, _).fold(e => Failure(e), v => Success(v)))
 
+  /**
+   * Returns a new deserializer that deserializes values as Option values, mapping null data to None values.
+   */
   def asOption(implicit @nowarn ev: T <:< AnyRef): Deserializer[R, Option[T]] =
     Deserializer((topic, headers, data) => ZIO.foreach(Option(data))(deserialize(topic, headers, _)))
 }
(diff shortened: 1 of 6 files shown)


src/main/scala/zio/kafka/serde/Serdes.scala
```diff
@@ -4,15 +4,116 @@ import java.nio.ByteBuffer
...
-  lazy val long: Serde[Any, Long]             = Serde(KafkaSerdes.Long()).inmap(Long2long)(long2Long)
-  lazy val int: Serde[Any, Int]               = Serde(KafkaSerdes.Integer()).inmap(Integer2int)(int2Integer)
-  lazy val short: Serde[Any, Short]           = Serde(KafkaSerdes.Short()).inmap(Short2short)(short2Short)
-  lazy val float: Serde[Any, Float]           = Serde(KafkaSerdes.Float()).inmap(Float2float)(float2Float)
-  lazy val double: Serde[Any, Double]         = Serde(KafkaSerdes.Double()).inmap(Double2double)(double2Double)
-  lazy val string: Serde[Any, String]         = Serde(KafkaSerdes.String())
-  lazy val byteArray: Serde[Any, Array[Byte]] = Serde(KafkaSerdes.ByteArray())
-  lazy val byteBuffer: Serde[Any, ByteBuffer] = Serde(KafkaSerdes.ByteBuffer())
-  lazy val uuid: Serde[Any, UUID]             = Serde(KafkaSerdes.UUID())
+  lazy val long: Serde[Any, Long] =
```
svroonland, line 11: It's quite a bit more verbose now, any way to still make some convenience function for this?

src/main/scala/zio/kafka/serde/Deserializer.scala
```diff
@@ -16,7 +17,12 @@ import scala.jdk.CollectionConverters._
  */
 trait Deserializer[-R, +T] {
   def deserialize(topic: String, headers: Headers, data: Array[Byte]): RIO[R, T]
-  def configure(props: Map[String, AnyRef], isKey: Boolean): Task[Unit]
```
svroonland, line 19: Nice, I like that serdes are less side-effecty now.

src/main/scala/zio/kafka/serde/Deserializer.scala
```diff
@@ -58,18 +68,21 @@ object Deserializer extends Serdes {
...
-  def apply[T](deserializer: KafkaDeserializer[T]): Deserializer[Any, T] = new Deserializer[Any, T] {
-    override def deserialize(topic: String, headers: Headers, data: Array[Byte]): Task[T] =
-      Task(deserializer.deserialize(topic, headers, data))
-
-    override def configure(props: Map[String, AnyRef], isKey: Boolean): zio.Task[Unit] =
-      Task(deserializer.configure(props.asJava, isKey))
-  }
+  def fromKafkaDeserializer[T](
+    deserializer: KafkaDeserializer[T],
+    props: Map[String, AnyRef],
```
svroonland, line 78: It's not very clear from the type what should be passed here. Can we document that?

src/main/scala/zio/kafka/serde/Serdes.scala
iravid: Ah yeah. Will do that momentarily.

aartigao:
We use Confluent Avro Serdes at work but I won't have a chance to try this until Monday. But at first glance looks good to me and I love the fact that `Blocking` can be used for the potential I/O calls to Avro Schema Registry  👍🏼

iravid:
Hey @aartigao - let us know if you have any input before we merge this 🙏

aartigao:
Whoops! I completely forgot it! 😅 

I've just tried and works fine 👍🏽

iravid:
Perfect, thank you!

kmatasflp:
We use Confluent Serdes exclusively, everything in our Kafka topics is avro, also we started experimenting with protobuf, introduced in Confluent Platform version 5.5. This change is definitely welcome one and it would simplify how we create our Avro Serdes (https://scastie.scala-lang.org/iXgywvtaS1OfXFOhDH9UNQ)

But even bigger help for us would be if `Consumer` and `Producer` creation were similar: Currently you can create `Consumer` without providing Serdes, but you need to pass them in when you create the `Producer`, so we need to have something like this and sadly it does not work for all our use cases

```
for {
  sourceKeySerde <- AvroSerdes.genericKeyM[Foo]
  sourceValueSerde <- AvroSerdes.genericValueM[Bar]
  sinkValueSerde <- AvroSerdes.genericValueM[FooBar]
  consumerAndProducer = { // be much cleaner if we did not need to create this layer here and could provide it from outside
    val kafkaBrokers = kafkaClientSettings.kafkaSslBrokers.split(',').toList
            val consumerSettings = ConsumerSettings(kafkaBrokers)
              .withGroupId(appConfig.groupId)
              .withProperties(kafkaClientSettings.props)
            val producerSettings = ProducerSettings(kafkaBrokers)
              .withProperties(kafkaClientSettings.props)

            ZLayer.fromManaged(Consumer.make(consumerSettings)) ++
              ZLayer.fromManaged(
                Producer.make(
                  producerSettings,
                  Serde.string, // inflexible in case of your app acts as a router - writes to multiple topics with different schemas
                  sinkValueSerde // inflexible in case of your app acts as a router - writes to multiple topics with different schemas
                )
              )
   }
   _ <- Consumer.subscribeAnd(Subscription.topics("some-topic"))
            .plainStream(sourceKeySerde.asTry, sourceValueSerde.asTry)
            .map(someTransformation)
            .mapChunksM(produceAndCommit)
            .runDrain
            .provideCustomLayer(consumerAndProducer) // need to use `provideSomeLayer` in case your transformations also depend on some R, and that can become quite ugly, ie something like `provideSomeLayer[R1 with R2 with R3 with ... with Logging with ZEnv](consumerAndProducer)`
} yield ()
```

PR #348, opened 2021-08-18
Make AdminClient a module

gurghet:
I just mechanically copied the work done on consumer and producer.

### DIFF
--- a/src/main/scala/zio/kafka/admin/AdminClient.scala
+++ b/src/main/scala/zio/kafka/admin/AdminClient.scala
@@ -40,11 +40,7 @@ import zio.duration.Duration
 import scala.collection.compat._
 import scala.jdk.CollectionConverters._
 
-/**
- * Thin wrapper around apache java AdminClient. See java api for descriptions
- * @param adminClient
- */
-case class AdminClient(private val adminClient: JAdminClient, private val blocking: Blocking.Service) {
+trait AdminClient {
   import AdminClient._
 
   /**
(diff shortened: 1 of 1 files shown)


src/main/scala/zio/kafka/admin/AdminClient.scala
```diff
@@ -247,47 +145,271 @@ case class AdminClient(private val adminClient: JAdminClient, private val blocki
...
+}
+
+object AdminClient {
+
+  /**
+   * Thin wrapper around apache java AdminClient. See java api for descriptions
+   * @param adminClient
+   */
+  private final case class LiveAdminClient(
+    private[admin] val adminClient: JAdminClient,
```
gurghet, line 167: Any nicer way to do this?

src/main/scala/zio/kafka/admin/AdminClient.scala
```diff
@@ -622,10 +744,10 @@ object AdminClient {
       KafkaConfig(jConfig.entries().asScala.map(e => e.name() -> e).toMap)
   }
 
-  def make(settings: AdminClientSettings) =
+  def make(settings: AdminClientSettings): ZManaged[Has[Blocking.Service], Throwable, AdminClient] =
```
gurghet, line 747: Without explicit type it says I'm exporting `LiveAdminClient`, I don't know if is supposed to.

iravid: Not having the type annotation there is an oversight - thank you for adding it!

src/main/scala/zio/kafka/admin/AdminClient.scala
iravid: This looks pretty good to me. What are you referring to specifically?

iravid (commented):
Awesome work. Let's also add a `def live: ZLayer[Has[Blocking.Service], Throwable, Has[AdminClient]]` on the `AdminClient` object?

gurghet:
I added `def live: ZLayer[Has[Blocking.Service] with Has[AdminClientSettings], Throwable, Has[AdminClient]]` instead of passing the `AdminClientSettings` as a parameter. I don't completely understand why this was done and did it for sake of uniformity.

src/main/scala/zio/kafka/admin/AdminClient.scala
gurghet: thinking about it, it actually looks ok, I guess I was bothered by the asymmetry of the two params scope.

src/main/scala/zio/kafka/admin/AdminClient.scala
```diff
@@ -247,47 +145,277 @@ case class AdminClient(private val adminClient: JAdminClient, private val blocki
...
   }
-}
 
-object AdminClient {
+  val live: ZLayer[Has[Blocking.Service] with Has[AdminClientSettings], Throwable, Has[AdminClient]] =
+    (for {
+      settings <- ZManaged.service[AdminClientSettings]
+      admin    <- make(settings)
+    } yield admin).toLayer
+
```
gurghet, line 418: `live` helper function, I guess one could always use the make function if he wants to pass the `settings` by parameter.

iravid:
Anything else you'd like to add here @gurghet? Or can we merge and release this?

gurghet:
No, it's fine :shipit:

iravid (approved):

Issue #33, opened 2019-09-28
Add support for exactly-once messaging


iravid:
I've never used that feature of Kafka, so if anyone's knowledgeable about this, feel free to chime in and describe what's needed.

ragb:
This is something I may want to use so I'd like at least to take a look.

Kafka Exactly once delivery in kafka is based around transactions in the producer and isolation levels in the consumer, plus some coordination between them so offset committing is part of a transaction.

I'm not sure if one should just provide the base low level operations (beginTransaction, commitTransaction, etc.) and leave coordination and management to the client or a more high level API would be required, such as:

```scala

Producer.withTransaction { transactionalProducer =>
    processRecords(records) *> transactionalProducer.send(...)
}
```

I'd go for the 2nd option but I'm afraid of losing some flexibility.

Relevant documentation is found here: 

https://www.confluent.io/blog/transactions-apache-kafka/
https://www.confluent.io/blog/exactly-once-semantics-are-possible-heres-how-apache-kafka-does-it/

Please tell me me if this is something to proceed on or not.

iravid:
Your help would be greatly appreciated @ragb! I don’t have any particular opinion on how this should look like, currently. I’ll try to form one over the next few days.

ragb:
I've been researching this a bit more and here goes a very high level proposal which I believe might cover all needs and possible use cases:

1. In `zio.producer.Producer.Service` provide wrappers  around the producer transaction API: `initTransactions()`, `beginTransaction()`,  `abortTransaction()`, `commitTransaction()` and `sendOffsetsToTransaction(offsets, consumerGrouopId)`.
2. In a new zio.transactional` (or similar) package create a more high level API with transaction management as `ZManaged` and other goodies based on existing Producer and Consumer APIs. Still not sure how this would look like, but at least provide some easy way of doing a transaction per consumer record pattern.

ragb:
reflecting a bit more (and while coding it:

) not sure if exposing begin / abort / commit transactions to the end user would be correct since one can't guarantee correctness on usage... A different approach could be providing a Managed object containing only the producer methods that make sense in a transaction context (e.g. send, sendOffsetsToTransaction, flush)... then commit the transaction on success and abort on exception / interruption.

svroonland:
Hi @ragb, Using `Managed` for transactions seems like the right tool for the job to me!

gurghet:
[...]

PR #351, opened 2021-08-25
First draft of transactional API, `ZManaged` kind 

gurghet:
Addresses https://github.com/zio/zio-kafka/issues/33
This is a first (crude) attempt to outline an API for exactly-once semantic. A few comments:

- This would prevent exposing the lower level APIs directly (except for `initTransactions` which is done automatically) in a later implementation because mixing and matching becomes unsafe.
- This API is more powerful that a simple `produceRecords` because it allows for user operated abort and interleaving of non-transactional messages. The manual abort might actually be more harmful than useful because if the user forgets to schedule it on errors he might actually commit a half baked transaction.
- ~There is a lot of duplication due to the creation of a new `LiveTransactionalProducer` which calls its underlying `Live`~ I removed much of the duplication by relying more on inheritance. I think it's possible to do a cleaner job without any inheritance but since it would require a more extensive refactoring I didn't attempt it.
- Also, inheritance is used instead of composition to avoid even more duplication
- It is currently possible to open more than one transaction concurrently which will result in a runtime error, we could decide to either make it a typelevel error (will probably require zio-prelude), leave it a runtime error, or create a transactional producer message to queue together with the other messages.


gurghet:
Closed by https://github.com/zio/zio-kafka/pull/356

PR #356, opened 2021-09-01
Transactional second draft

gurghet:
This version of https://github.com/zio/zio-kafka/pull/351, while simplified, probably strikes a good balance.
The main difference is that the transactional producer has only the *transactional* modality. It is possible to interleave with non-transactional messages (as demonstrated in the integration tests) but it requires a second, non-transactional producer. This is because of the greatly limited kafka API.
It would be possible in the future to build a general purpose producer that internally contains both the producers and even queues concurrent transactions, but the current implementation probably covers >80% of use cases already.

### DIFF
--- a/src/main/scala/zio/kafka/producer/Transaction.scala
+++ b/src/main/scala/zio/kafka/producer/Transaction.scala
@@ -1,28 +1,42 @@
 package zio.kafka.producer
 
 import org.apache.kafka.clients.producer.{ ProducerRecord, RecordMetadata }
+import zio.kafka.consumer.{ Offset, OffsetBatch }
 import zio.kafka.producer.TransactionalProducer.UserInitiatedAbort
-import zio.{ Cause, IO, RIO, RefM, ZIO }
+import zio.{ Cause, Chunk, IO, RIO, Ref, ZIO }
 import zio.kafka.serde.Serializer
 
-final private[producer] class Transaction(
-  private val producer: Producer
+final class Transaction private[producer] (
+  private val producer: Producer,
+  private[producer] val offsetBatchRef: Ref[OffsetBatch]
 ) {
   def produce[R, K, V](
     topic: String,
     key: K,
     value: V,
     keySerializer: Serializer[R, K],
-    valueSerializer: Serializer[R, V]
+    valueSerializer: Serializer[R, V],
+    offset: Option[Offset]
   ): RIO[R, RecordMetadata] =
-    produce(new ProducerRecord[K, V](topic, key, value), keySerializer, valueSerializer)
+    produce(new ProducerRecord[K, V](topic, key, value), keySerializer, valueSerializer, offset)
 
   def produce[R, K, V](
     producerRecord: ProducerRecord[K, V],
     keySerializer: Serializer[R, K],
-    valueSerializer: Serializer[R, V]
+    valueSerializer: Serializer[R, V],
+    offset: Option[Offset]
   ): RIO[R, RecordMetadata] =
-    producer.produce[R, K, V](producerRecord, keySerializer, valueSerializer)
+    ZIO.whenCase(offset) { case Some(offset) => offsetBatchRef.update(_ merge offset) } *>
+      producer.produce[R, K, V](producerRecord, keySerializer, valueSerializer)
+
+  def produceChunk[R, K, V](
+    records: Chunk[ProducerRecord[K, V]],
+    keySerializer: Serializer[R, K],
+    valueSerializer: Serializer[R, V],
+    offset: Option[Offset]
+  ): RIO[R, Chunk[RecordMetadata]] =
+    ZIO.whenCase(offset) { case Some(offset) => offsetBatchRef.update(_ merge offset) } *>
+      producer.produceChunk[R, K, V](records, keySerializer, valueSerializer)
 
   def abort: IO[TransactionalProducer.UserInitiatedAbort.type, Nothing] =
     ZIO.haltWith(t => Cause.traced(Cause.fail(UserInitiatedAbort), t()))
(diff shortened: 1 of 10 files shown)


sirocchj:
@iravid how do you feel about this? It would be good to have some support for transactions - this seems like a good start?

iravid:
Hi there, sorry for the delay here. It would be amazing to have transactional support - thank you for submitting this @gurghet. I'll try and have a look today.

src/main/scala/zio/kafka/producer/Transaction.scala
```diff
@@ -0,0 +1,80 @@
...
+    offset: Option[Offset]
+  ): RIO[R, Chunk[RecordMetadata]] =
+    haltIfClosed *>
+      ZIO.whenCase(offset) { case Some(offset) => offsetBatchRef.update(_ merge offset) } *>
+      producer.produceChunk[R, K, V](records, keySerializer, valueSerializer)
+
+  def abort: IO[TransactionalProducer.UserInitiatedAbort.type, Nothing] =
+    ZIO.fail(UserInitiatedAbort)
+
+  private[producer] def markAsClosed: UIO[Unit] = closed.set(true)
```
gurghet, line 74: This doesn't cover some edge cases which are covered by regular Kafka exception. The only edge case which is not covered by either (that I can think of) is whenever we check if a transaction is still open, and by the time this returns, not only the transaction is closed, but another transaction is open. This would allow the producer to send a message in the wrong transaction.

src/test/scala/zio/kafka/ProducerSpec.scala
```diff
@@ -338,6 +335,30 @@ object ProducerSpec extends DefaultRunnableSpec {
...
+          transactionThief <- Ref.make(Option.empty[Transaction])
+          _                <- TransactionalProducer.createTransaction.use { t =>
+            transactionThief.set(Some(t))
+          }
+          t                <- transactionThief.get
+          _ <- t.get.produce("any-topic", 0, 0, Serde.int, Serde.int, None)
+        } yield ()
+        assertM(test.run)(failsCause(containsCause(Cause.fail(TransactionLeaked(OffsetBatch.empty)))))
+      },
+      testM("fails if transaction leaks in an open transaction") {
```
gurghet, line 350: This should pass every time but if it was done by 2 fibers it could fail with a bad timing (see comment above). I think this is ok for most intent and purposes and, by listing it as a known issue, users for which this is critical can find a workaround by syncing the fibers.

Other than that, the target state is probably an inversion of control, similar to what you did with the consumer, with a global queue, this way *everything* is serialized thus making it easier to prevent such bad ordering.

sirocchj:
Sorry for paging you on this @iravid but I'm personally very interested in getting something like this in the library asap, of course only if you feel the tradeoffs and choices being made here are compatible with the library's philosophy and with where you see it going in the future.

FYI we are using this today via a fork but think others could benefit from it too.

gurghet:
Related to https://github.com/zio/zio-kafka/issues/33

gurghet:
@iravid I'll start adding docs and javadocs after your blessings

iravid (commented):
Thank you @gurghet, I'm 100% onboard with adding this feature to the producer. However, I'd like us to explore modifying the existing producer to add transactional support. 

In particular, I'm bothered by the fact that nothing prevents multiple concurrent TransactionalProducers from running.

Adding transactionality to the existing producer would be pretty similar. We'd use the same semaphore to ensure we don't have two concurrent transactions, and a `beginTransaction: ZManaged[Transaction]` for acquiring a new transaction.

I'm wondering if it'd be useful to also expose a `currentTransaction: ZManaged[Transaction]`. That would give you the current transaction and possibly we could setup a reference counting mechanism that would only finalize the transaction if all fibers that acquired `currentTransaction` are done. But that can be done in a follow-up.

Additionally I don't mind getting the current version out and doing the changes in a follow-up.

Let me know what you think and thank you for this awesome contribution!

gurghet:
This makes sense. I'll explore modifying the existing producer now. Only one thing comes to mind: if we only have one type `Producer` it means that calling `beginTransaction` will not be typesafe anymore because creating a normal `Producer` doesn't require a `transactionalProducerSettings`. Do you think it's a tradeoff we can have or did you have something in mind to keep it typesafe?

iravid:
We could enhance ProducerSettings to keep track of the type of producer it can create and make `TransactionalProducer` extend `Producer`. But I'd skip that for now and follow-up later; for now you could just fail at runtime if the producer wasn't created with transactional settings.

If we're happy with the API after it sees some usage, we can enhance the type safety.

gurghet:
Definitely agree. So I would add something like:
```
  /**
   * This opens a new transaction, blocking the fiber until all previous transactions 
   * are closed and initializing the transactions for the first transaction.
   * It will fail with `NonTransactionalProducer` if the configuration does not contain
   * a transactional id.
   */
  def beginTransaction: TaskManaged[Transaction]
```
to the `Producer`.

iravid:
Sounds good! Please ping me again when you'd like me to review :-)

gurghet:
Hi @iravid after looking at the alternative in the other PR and playing with it for a bit, I would suggest to merge this and build on it for improvements.
I would also like to know what do you think about the idea of eventually creating a global queue to serialize messages and avoid rare edge cases like the ones [theorized above](https://github.com/zio/zio-kafka/pull/356#discussion_r713705790). This would help also for your concern of multiple producers if I understood it correctly.

iravid (approved):
Hi @gurghet, I'd like to retract my previous comment because I've misunderstood the code. I was under the impression that a TransactionalProducer is created by a user from an existing Producer, but that's not the case as the underlying Producer is not accessible at all.

The only issue I currently see is that if a transaction is leaked out of the managed scope, it could result in a use-after-free. This could happen if, e.g., a call to `produce` is forked. But that is usually a bad practice and probably not that common.

So let's get this merged and released :-)

gurghet:
Thanks, and yes, that's a possible use-after-free but, as discussed in my comments, only in the edge case that a produce is scheduled before the transaction is marked as exhausted, fiber suspended, transaction marked as exhausted, closed, and a new one reopened.

PR #365, opened 2021-09-23
Transactional api (only Producer)

gurghet:
Hi @iravid, I created a new PR because I don't think this is an improvement after all.
The main issue here is that the underlying kafka client sees the transaction id in the config and disables all the non-transactional methods. I thought that, before calling `initTransactions` everything goes, [apparently not](https://github.com/apache/kafka/blob/01ab888dbd08ccd4b0de9333d21581ce24fe2c3b/clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java#L964). This means that doing this:
```
for {
_ <- Producer.produce("topic", "key", "value", Serde.string, Serder.string)
} yield ()
```
will throw a runtime error. I think this is very counter-intuitive and makes having a separate `TransactionalProducer` with `beginTransaction` as the only method, simpler to understand.
Additionally, mixing transactional and non-transactional messages becomes harder to program, but this is probably a rare use case.

I would suggest to go with the other approach for the moment.

If you elaborate more on the concurrent `TransactionalProducers` problem I'll think of a mitigation, but I want to make sure to understand your concern.


src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -247,6 +277,36 @@ object Producer {
...
+        blocking.effectBlocking(p.commitTransaction())
+
+    private def makeTransactionalOrFail: Task[Unit] = ZIO.whenM(canTransact.get.map(_ == false)) {
+      if (producerSettings.properties.contains(ProducerConfig.TRANSACTIONAL_ID_CONFIG)) {
+        blocking.effectBlocking(p.initTransactions()) *>
+          canTransact.set(true)
+      } else {
+        IO.fail(NonTransactionalProducer)
+      }
+    }
```
gurghet, line 307: Calling `initTransactions` just in time doesn't make any difference because the Kafka producer will refuse to work until then.

iravid:
@gurghet Thank you for exploring this approach! I agree that it is useless if the producer is not usable outside of a transaction. Closing this PR; let's continue in #356.

Issue #375, opened 2021-10-18
It's no longer possible to consume a topic manually


mbaechler:
When adding transaction support, zio-kafka lost the ability to consume a topic manually.

It's a use-case we use a lot at conduktor (the company I work for).

To be specific, the problem comes from https://github.com/zio/zio-kafka/commit/c9f2952345d9fb0d9fff82083bbc5fdbfffc64e9#diff-4bb68de9081869e01fdb038a48581bb121471b3de1ddad2cf0e834e4e3e356fdR184 where we query the `groupMetadata` of the consumer, which unfortunately throws in the Java driver.

PR #376, opened 2021-10-18
GH-365 Allow manual subscription to consume a topic

mbaechler:
Fixes https://github.com/zio/zio-kafka/issues/375

I required to make groupId optional and thus swap arguments for consumer construction.

The fix is not perfect as I found no solution to know if I can call `groupMetadata` without getting an exception ( see https://github.com/zio/zio-kafka/compare/master...mbaechler:allow-manual-consumer?expand=1#diff-4bb68de9081869e01fdb038a48581bb121471b3de1ddad2cf0e834e4e3e356fdR186 )

The same test is passing on 0.16.0.

I would appreciate a release 0.17.1 to fix this regression is possible.

### DIFF
--- a/src/main/scala/zio/kafka/consumer/CommittableRecord.scala
+++ b/src/main/scala/zio/kafka/consumer/CommittableRecord.scala
@@ -39,7 +39,7 @@ object CommittableRecord {
   def apply[K, V](
     record: ConsumerRecord[K, V],
     commitHandle: Map[TopicPartition, Long] => Task[Unit],
-    consumerGroupMetadata: ConsumerGroupMetadata
+    consumerGroupMetadata: Option[ConsumerGroupMetadata]
   ): CommittableRecord[K, V] =
     CommittableRecord(
       record,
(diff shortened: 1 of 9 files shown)


iravid:
Thanks @mbaechler and sorry we broke manual consumption which is an important use case to support.

This solution works for now, though I'd prefer to maybe model the types differently to prevent this from happening. Though I think we'd need a module with path dependent types to completely prevent this from happening.

Either way let's merge this and release 0.17.1 and deal with a better solution later. Luckily we don't guarantee binary compatibility yet. :-)

iravid:
Thank you @mbaechler!

mbaechler:
You're welcome

guizmaii:
@iravid 

> Though I think we'd need a module with path dependent types to completely prevent this from happening.

Can you maybe explain a bit more your idea? 🤔
I'm working on the performances of the consumer (see https://github.com/zio/zio-kafka/pull/697) and this `consumerGroupMetadata` computation, only needed when using a `TransactionalProducer` alongside a `Consumer`, is expensive while probably not used by 90% of our users

PR #390, opened 2021-12-10
Update README to fix #389

### LINKED ISSUE #389: Changes on Producer didn't make it to the README
#331 changed the signature of make from:

```scala
def make[R, K, V](
      settings: ProducerSettings,
    ): RManaged[Blocking, Service[R, K, V]]
```

to:

```scala
def make(settings: ProducerSettings): RManaged[Blocking, Producer]
```

but the documentation on README hasn't been updated to reflect the changes:

```scala
val consumerAndProducer = 
  ZLayer.fromManaged(Consumer.make(consumerSettings)) ++
    ZLayer.fromManaged(Producer.make(producerSettings, Serde.int, Serde.string))
```

DiegoTavares:
With the changes on #331, keySerializer and valueSerializer need
to be passed on every produce* call. The reasoning behind this
change can be found at the comments on #331.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ val producerSettings: ProducerSettings = ProducerSettings(List("localhost:9092")
 
 val consumerAndProducer = 
   ZLayer.fromManaged(Consumer.make(consumerSettings)) ++
-    ZLayer.fromManaged(Producer.make(producerSettings, Serde.int, Serde.string))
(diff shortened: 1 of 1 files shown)


adamgfraser (changes requested):
Can we use `mdoc` for these examples so this doesn't happen again?

PR #408, opened 2022-01-10
Expose the internal Java `AdminClient`

guizmaii:
So that this lib is easier to use when some of the Java `AdminClient` methods are not yet implemented in the zio-kafka lib


strokyl:
:+1:

PR #410, opened 2022-01-11
Allow lib user to construct they own java client

strokyl:
There is some case where we need both the ZIO#AdminClient and have an access to the underlying Kafka#AdminClient.
In this case, I think it's convenient to allow building a ZIO#AdminClient providing an already existing Kafka#AdminClient


svroonland (dismissed):
Looks useful 👍

svroonland:
@strokyl If you could update with the master branch, I can merge it

svroonland (approved):

svroonland:
Thanks!

PR #411, opened 2022-01-13
Implements delete records

mdulac:
This is a passthrough function to the Admin Client for deleting records

### DIFF
--- a/src/main/scala/zio/kafka/admin/AdminClient.scala
+++ b/src/main/scala/zio/kafka/admin/AdminClient.scala
@@ -1,6 +1,6 @@
 package zio.kafka.admin
 
-import java.util.Optional
+import org.apache.kafka.clients.admin.ListOffsetsResult.{ ListOffsetsResultInfo => JListOffsetsResultInfo }
 import org.apache.kafka.clients.admin.{
   AdminClient => JAdminClient,
   AlterConsumerGroupOffsetsOptions => JAlterConsumerGroupOffsetsOptions,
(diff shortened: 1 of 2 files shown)


strokyl:
:+1:

src/main/scala/zio/kafka/admin/AdminClient.scala
```diff
@@ -74,6 +74,14 @@ trait AdminClient {
    */
   def deleteTopic(topic: String): Task[Unit]
 
+  /**
+   * Delete records.
+   */
+  def deleteRecords(
+    recordsToDelete: Map[TopicPartition, RecordsToDelete],
+    deleteRecordsOptions: Option[DeleteRecordsOptions] = None
```
svroonland, line 82: Most of the admin methods have a scala variant of such an options object, like `listConsumerGroups` has `ListConsumerGroupsOptions`. Can we implement that here as well?

svroonland (changes requested):
Thanks for the contribution! One request regarding the options parameter

src/main/scala/zio/kafka/admin/AdminClient.scala
```diff
@@ -810,6 +836,13 @@ object AdminClient extends Accessible[AdminClient] {
     }
   }
 
+  case class DeleteRecordsOptions(timeout: Option[Duration]) {
```
svroonland, line 839: It's a bit weird to have an `Option[DeleteRecordOptions]` which only has an `Option[Duration]` field, but I guess this matches the java API more closely 👍

svroonland (approved):
Looks good, thanks!

svroonland:
I see that there's other methods where we're using the java classes directly, raised an issue for that: https://github.com/zio/zio-kafka/issues/418

PR #414, opened 2022-01-18
Handle `Command.Poll` while subscribed only

aartigao:
In our systems, once in a blue moon, one of our ZIO instances has network issues between Kafka Client and Kafka Broker Coordinator, leaving processing of messages stuck. We use different technologies for different services so we have some systems using Akka Kafka and others ZIO Kafka. Both of them use the same Kafka Clients (3.0.0) and interact with the same Kafka Cluster (0.10.2.0). We only see this pattern on the ZIO instances.

Recently the issue happened again and we were able to profile the system with Async Profiler:

image

We found out that `KafkaConsumer.poll` throws `IllegalStateException` for multiple reasons, but ZIO Kafka is treating all of them as if they were only when no call to `subscribe` has been made yet. This PR fixes this by adding a guard to ensure we're on the mentioned scenario (and let it crash otherwise).

We're not completely sure that this issue will be resolved with this fix, but we believe it's a good idea to not shadow all the `IllegalStateException`s under the umbrella of not being subscribed. In this same scenario, Alpakka Kafka closes the related Actor, returning the control to user which can decide what to do to handle the fatal situation.

### DIFF
--- a/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -30,7 +30,8 @@ private[consumer] final class Runloop(
   diagnostics: Diagnostics,
   shutdownRef: Ref[Boolean],
   offsetRetrieval: OffsetRetrieval,
-  userRebalanceListener: RebalanceListener
+  userRebalanceListener: RebalanceListener,
+  subscribedRef: Ref[Boolean]
 ) {
   private val isRebalancing = rebalancingRef.get
   private val isShutdown    = shutdownRef.get
@@ -69,6 +70,10 @@ private[consumer] final class Runloop(
 
     trackRebalancing ++ emitDiagnostics ++ userRebalanceListener
   }
+
+  def markSubscribed: UIO[Unit] = subscribedRef.set(true)
+
+  def markUnsubscribed: UIO[Unit] = subscribedRef.set(false)
 
   private def commit(offsets: Map[TopicPartition, Long]): Task[Unit] =
     for {
@@ -239,21 +244,15 @@ private[consumer] final class Runloop(
     if (toPause.nonEmpty) c.pause(toPause.asJava)
   }
 
-  private def doPoll(c: ByteArrayKafkaConsumer, requestedPartitions: Set[TopicPartition]) =
-    try {
-      val pollTimeout =
-        if (requestedPartitions.nonEmpty) this.pollTimeout.asJava
-        else 0.millis.asJava
+  private def doPoll(c: ByteArrayKafkaConsumer, requestedPartitions: Set[TopicPartition]) = {
+    val pollTimeout =
+      if (requestedPartitions.nonEmpty) this.pollTimeout.asJava
+      else 0.millis.asJava
 
-      val records = c.poll(pollTimeout)
+    val records = c.poll(pollTimeout)
 
-      if (records eq null) ConsumerRecords.empty[Array[Byte], Array[Byte]]() else records
-    } catch {
-      // The consumer will throw an IllegalStateException if no call to subscribe
-      // has been made yet, so we just ignore that. We have to poll even if c.subscription()
-      // is empty because pattern subscriptions start out as empty.
-      case _: IllegalStateException if c.subscription.isEmpty => ConsumerRecords.empty[Array[Byte], Array[Byte]]()
-    }
+    if (records eq null) ConsumerRecords.empty[Array[Byte], Array[Byte]]() else records
+  }
 
   private def pauseAllPartitions(c: ByteArrayKafkaConsumer) = ZIO.effectTotal {
     val currentAssigned = c.assignment()
@@ -387,7 +386,7 @@ private[consumer] final class Runloop(
   private def handleOperational(state: State, cmd: Command): RIO[Blocking, State] =
     cmd match {
       case Command.Poll() =>
-        handlePoll(state)
+        ZIO.ifM(subscribedRef.get)(handlePoll(state), UIO(state))
       case Command.Requests(reqs) =>
         handleRequests(state, reqs).flatMap { state =>
           // Optimization: eagerly poll if we have pending requests instead of waiting
@@ -470,7 +469,8 @@ private[consumer] object Runloop {
                           )
                       }
                       .toManaged(_.shutdown)
-      shutdownRef <- Ref.make(false).toManaged_
+      shutdownRef   <- Ref.make(false).toManaged_
+      subscribedRef <- Ref.make(false).toManaged_
       runloop = new Runloop(
                   consumer,
                   pollFrequency,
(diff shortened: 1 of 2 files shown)


iravid:
Hi @aartigao, thanks so much for investigating and fixing this. I agree that generally it's better to crash the consumer if there's an issue.

What bothers me here is whether `c.subscription.isEmpty()` is the right indicator for retrying. Maybe we should only retry if the subscription is a Pattern subscription? And crash in all other cases?

Unfortunately we don't have snapshot releases ATM so we can't merge this and have you test the snapshot. Do you have a way to publish locally and test if this fixes your issue?

aartigao:
> What bothers me here is whether `c.subscription.isEmpty()` is the right indicator for retrying. Maybe we should only retry if the subscription is a Pattern subscription? And crash in all other cases?

Right, I'll take a look and see if I can change it 👍🏽 

> Unfortunately we don't have snapshot releases ATM so we can't merge this and have you test the snapshot. Do you have a way to publish locally and test if this fixes your issue?

We can publish this fixed version into our internal Nexus. But as I said, this happens very rarely (in fact that it just happened 2 times the last 6 months 😅). Due to this, it's unlikely that I can come back here soon and corroborate that the issue is fixed 😞

iravid:
Ok. So I think our best bet here is to tighten down the predicate to only apply to the pattern subscription case. That would be a good improvement.

aartigao:
Let me know if those are the changes you had in mind.

To me, the code looks a little messy with the interactions between `Consumer` and `Runloop` but I didn't find another way to know which subscription was being used without a `Ref`.

aartigao:
That makes me wonder if we can't rollback to the initial proposal 🤔 

Maybe we can even handle it at a higher level: does make sense to request a `KafkaConsumer.poll` is you know beforehand that `KafkaConsumer.subscription` is empty? Shouldn't we do it that way could be removed the `try - catch IllegalStateException` for all the use cases (no matter what kind of `Subscription` is placed?)

Something like:

```scala
private def doPoll(c: ByteArrayKafkaConsumer, requestedPartitions: Set[TopicPartition]) =
  if (c.subscription().isEmpty) ConsumerRecords.empty[Array[Byte], Array[Byte]]()
  else {
    val pollTimeout =
      if (requestedPartitions.nonEmpty) this.pollTimeout.asJava
      else 0.millis.asJava
    val records = c.poll(pollTimeout)
    if (records eq null) ConsumerRecords.empty[Array[Byte], Array[Byte]]() else records
  }
```

iravid:
No, that won’t work for pattern subscriptions. As far as I know, you need to poll the consumer for it to refresh the topic metadata and see which topics match a pattern.

aartigao:
Do you have a source for this information about the pattern subscription? All I find is answers like [this](https://stackoverflow.com/a/66444976/2719284) or [this other](https://stackoverflow.com/a/38833347/2719284) where it's suggested that `metadata.max.age.ms` is in charge of this.

I guess for the context that requesting a `poll` also refreshes the metadata but is this a good practice? Perhaps a mention in the docs is enough for this so people using Pattern can tweak the `metadata.max.age.ms` to control the refreshing of newly discovered matches.

aartigao:
For the record, I'm doing some tests using the raw `KafkaConsumer` and subscribing via Pattern and then polling doesn't throw any exception. I've both subscribed to matching and non-matching patterns:

```scala
val consumer = new KafkaConsumer(
  Map[String, Object](
    ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG -> "XXX",
    ConsumerConfig.GROUP_ID_CONFIG          -> "XXX@stuart.com",
    ConsumerConfig.AUTO_OFFSET_RESET_CONFIG -> "latest",
    ConsumerConfig.MAX_POLL_RECORDS_CONFIG  -> "1"
  ).asJava,
  Serdes.String().deserializer(),
  Serdes.ByteArray().deserializer()
)
//consumer.subscribe(".*".r.pattern)
consumer.subscribe(UUID.randomUUID().toString.r.pattern)
consumer.poll(Duration.ofSeconds(5)).iterator().asScala.foreach(println(_))
```

aartigao:
I came with a simpler solution. 

Basically if some form of `Subscription` has been placed, then `Command.Poll` can be executed, otherwise is ignored. I think this fixes the original issue in Spec where `IllegalStateException` thrown when calling `poll` without `subscribe`.

Moreover, if `poll` also fetches metadata, this also still applies for Pattern Subscription, because the condition is only based on the `Ref` holding it. And given that using the raw `KafkaConsumer` I tested that no exception is thrown no matter if pattern matches something or not, there's not need to continue catch anything, thus `try-catch` statement has been removed.

svroonland:
@aartigao Nice work! I was wondering what is the expected behavior when the IllegalStateException is thrown? Should the runloop ignore it and try later, or should the stream fail? Is it implemented in the way we expect?

aartigao:
> @aartigao Nice work! I was wondering what is the expected behavior when the IllegalStateException is thrown? Should the runloop ignore it and try later, or should the stream fail? Is it implemented in the way we expect?

IMO the stream should fail, because you really don't know the internal state of the `KafkaConsumer` instance (maybe it's in an unrecoverable state). It sad because the `KafkaConsumer` API for `poll` states that:

```
@throws java.lang.IllegalStateException if the consumer is not subscribed to any topics or manually assigned any
            partitions to consume from
```

which is completely false because it's thrown, if you inspect the source code, in various places (even when the consumer is closed!).

With this changes, we enforce that `poll` won't be called before `subscribe`, assuming that any other errors thrown by `poll` are, in fact, unrecoverable errors.

svroonland:
> which is completely false because it's thrown..
Strictly speaking it doesn't say that it is exclusively thrown for that case only ;)

src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -239,21 +244,15 @@ private[consumer] final class Runloop(
     if (toPause.nonEmpty) c.pause(toPause.asJava)
   }
 
-  private def doPoll(c: ByteArrayKafkaConsumer, requestedPartitions: Set[TopicPartition]) =
-    try {
-      val pollTimeout =
-        if (requestedPartitions.nonEmpty) this.pollTimeout.asJava
-        else 0.millis.asJava
+  private def doPoll(c: ByteArrayKafkaConsumer, requestedPartitions: Set[TopicPartition]) = {
```
svroonland, line 247: If this throws the IllegalStateException, is it handled correctly? I'm not sure if `Task.effectSuspend` handles an exception in the creation of the effect correctly, or if it would result in a defect. `Runloop.run` handles ZIO failures with `onError`, but not defects.

I wonder what would happen if you manually insert a `throw new IllegalStateException` here.

aartigao: It's a failure 🙂 

```scala
testM("Effect suspend") {
  val effect = Task.effectSuspend {
    val cond = true
    if (cond) throw new IllegalArgumentException("foo")
    UIO("done")
  }
  assertM(effect.run)(fails(hasMessage(equalTo("foo"))))
}
```

svroonland (approved):

PR #424, opened 2022-02-01
Update CI pipelines

guizmaii:

PR #425, opened 2022-02-01
The `KafkaFuture` implementation is, since Kafka v3.0.0, backed by a `CompletableFuture`

guizmaii:
See:
  - https://issues.apache.org/jira/browse/KAFKA-6987
  - https://cwiki.apache.org/confluence/display/KAFKA/KIP-707%3A+The+future+of+KafkaFuture
  - https://github.com/apache/kafka/pull/9878
 
#### Why do we need to change this code?
In my project, we have tests testing that when a Topic doesn't exist but we are trying to do something with this topic, the code correctly fails.
In these tests, we are validating that the error we are getting back from the Kafka client is `org.apache.kafka.common.errors.UnknownTopicOrPartitionException`.

Since the update of the kafka-clients to v3+, our tests are now failing because we're not getting a `org.apache.kafka.common.errors.UnknownTopicOrPartitionException` error anymore but a `java.util.concurrent.CompletionException` error wrapping the `org.apache.kafka.common.errors.UnknownTopicOrPartitionException`:

```scala
[info]     - should fail if topic does not exist
[info]       java.util.concurrent.CompletionException: org.apache.kafka.common.errors.UnknownTopicOrPartitionException: <error message> did not satisfy isSubtype(UnknownTopicOrPartitionException)
```

This PR fixes this problem

### DIFF
--- a/src/main/scala/zio/kafka/admin/AdminClient.scala
+++ b/src/main/scala/zio/kafka/admin/AdminClient.scala
@@ -44,6 +44,7 @@ import zio.blocking.Blocking
 import zio.duration.Duration
 
 import java.util.Optional
+import java.util.concurrent.CompletionException
 import scala.jdk.CollectionConverters._
 
 trait AdminClient {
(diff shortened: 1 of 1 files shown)


src/main/scala/zio/kafka/admin/AdminClient.scala
```diff
@@ -507,20 +508,35 @@ object AdminClient extends Accessible[AdminClient] {
...
+     */
+    def unwrapCompletionException(isFatal: Throwable => Boolean)(t: Throwable): Task[Nothing] =
+      t match {
+        case e: CompletionException => Task.fail(e.getCause)
+        case e if !isFatal(e)       => Task.fail(e)
+        case e                      => Task.die(e)
+      }
+
+    kfv.flatMap(f =>
+      Task.effectSuspendTotalWith { (p, _) =>
```
svroonland, line 527: Did you consider using `ZIO.fromCompletionStage` or `ZIO.fromCompletableFuture`?

guizmaii: Yes, that was my first implementation for this fix but `ZIO.fromCompletionStage` doesn't handle cancellation and so the tests for the `KafkaFuture` in this repo were failing (thanks for these tests! Really helpful! 🙏)

svroonland: Fair enough. I wonder if cancellation handling should be added to `fromCompletionStage` then. @adamgfraser Do you happen to know if there's a reason for it not being implemented?

svroonland (approved):

svroonland:
Hmm, again stuck on the old version checks..

src/main/scala/zio/kafka/admin/AdminClient.scala
adamgfraser: @svroonland No. Seems like it would be a good addition.

svroonland:
Cool, thanks @guizmaii!  we can migrate to the new ZIO.fromCompletionStage when https://github.com/zio/zio/pull/6316 is merged and released.

guizmaii:
@svroonland, about:
> Hmm, again stuck on the old version checks..
Did you see my other PR: https://github.com/zio/zio-kafka/pull/424?

svroonland:
Yeah I did, but it turns out that those check names are not just in mergify.yml but also in the github branch protection rules.

I'll try to get around to your other PR soon!

PR #427, opened 2022-02-07
Transaction + offset batches related fixes

vigoo:
This is a set of fixes I had to make in order to implement a consumer with the following behavior:

- Consuming a kafka topic, buffering data for a while
- Periodically flush this buffered data (upload to S3 actually), and if that succeeds, _then_ commit all the offsets _and_ publish an event to another kafka topic, in a transaction.

The following three changes are implemented:

### 1. Missing method on `Transaction`

Adds missing `produceChunkBatch` method on `Transaction` so we can add all the offsets to the committed transaction together

### 2. Exposing `onLost` on the rebalance listener

The `onPartitionsLost` on the Java API has a default implementation that just calls `onPartitionsRevoked`. I exposed it on the `RebalanceListener`. In the above mentioned workflow it can be used to avoid extra uploads and failed transactions:

- in case of a revoke, we want to upload the buffered data and produce the transaction before finishing the rebalancing event
- in case of lost partitions we just free up resources but don't try to upload/commit

### 3. A new runloop mode

When committing a transaction, if it contains offsets with an old generation id (which is part of the consumer metadata and gets increased on each rebalance) we get a `Specified group generation id is not valid` error. With the current runloop implementation there was no way to guarantee that this does not happen during a rebalancing, because even if you try to commit everything during the revoke event, that is completely independent from feeding the open partition streams which are kept open if they don't get revoked.

This change introduces a new flag in the consumer settings called `restartStreamsOnRebalancing` that disables this optimization and guarantees that during revoke, all existing partition streams get finished. 

This new runloop mode enables implementing the mentioned example by:

- setting `restartStreamsOnRebalancing`
- using the `partitionedAssignmentStream` to get a new chunk of partition streams for each rebalance
- drain all the partition streams for each such chunk (`mapM` + `runDrain`)
- after the partition streams got drained, perform the flush and transactional  commit
- in the `onRevoked` callback wait for the flush and transactional commit to be performed


svroonland:
Question: when a rebalancing starts and you immediately drain the open partition streams and flush the transaction, how can you ensure that the consumer is still on the previous generation ID? Is there no race condition there between the rebalancing process and the flush/commit?

src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -261,6 +293,7 @@ private[consumer] final class Runloop(
 
   private def handlePoll(state: State): RIO[Blocking, State] =
     for {
+      _ <- currentState.set(state)
```
svroonland, line 296: I understand why you're doing this, in order to access the state from the rebalance listener, but can we be sure that it's the latest version of the state? As the Runloop's stream is also folding over the State and it gets updated when commits are done.

svroonland:
Is there any documentation/implementation on with transactions during rebalancing I can compare with?

svroonland:
> That way if you block the rebalance listener until these commits are done, you are guaranteed to complete during the rebalancing before the next generation.

But from `ConsumerRebalanceListener.onPartitionsRevoked` docs:
>  In cooperative rebalancing, it will be called at the end of a rebalance on the set of partitions being revoked iff the set is non-empty.

Doesn't that imply that there already may be a new consumer group generation when `onPartitionsRevoked` is called?

Sorry for all the questions, I'm not too familiar with the transaction features, so just trying to help validate the implementation 😅

src/main/scala/zio/kafka/producer/Transaction.scala
```diff
@@ -30,6 +30,13 @@ trait Transaction {
     offset: Option[Offset]
   ): RIO[R, Chunk[RecordMetadata]]
 
+  def produceChunkBatch[R, K, V](
```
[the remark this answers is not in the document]
vigoo: I would not do that in this PR if it's OK. Feel like it would lead to some more refactoring if we do it.

src/main/scala/zio/kafka/consumer/internal/Runloop.scala
vigoo: I wanted to write:

> It is safe because we set it unconditionally in each `handlePoll` to the latest, and it is only used by the rebalance listener which is always called as part of the `poll`.

but realized this is actually wrong, as `handlePoll` is called _after_ the `poll` so the state ref should be updated in `doPoll` before calling `poll` to make the above statement true. I'm fixing it.

svroonland: Alright, but there's the risk that the next change to the Runloop that wants to use the State does it in such a way that it can be overwritten later by the stream's fold. Then maybe it's better to remove the fold over the state completely and just use the Ref everywhere in a safe manner (i.e. atomic updates). What do you think?

vigoo: Would it be better if we would just call it `lastStateBeforePoll` instead of `currentState`?

svroonland:
I suppose it would not hurt to see if the Runloop can be simplified a bit more, with the complexity it has accrued with recent changes, but that's for a future PR.

svroonland (approved):

svroonland:
@vigoo I think you have to sign the CLA 😅

vigoo:
> @vigoo I think you have to sign the CLA 😅

is it a different one than for the other ZIO repos? checking

erikvanoosten:
Are there any documentation or examples on how to use the transactional producer?

erikvanoosten:
> Are there any documentation or examples on how to use the transactional producer?

Proposal docs added in https://github.com/zio/zio-kafka/pull/440

Issue #418, opened 2022-01-24
Scala case classes for all parameters to AdminClient methods


svroonland:
Many of the methods of `AdminClient` have a case class that mimics the Java variant of parameters, but there are some that use the apache kafka java classes directly. It would be nice to have case classes for all of them.

PR #435, opened 2022-03-04
Make case classes for options in Admin Client methods

mdulac:
This PR is related to issue https://github.com/zio/zio-kafka/issues/418

- It created the missing case classe for some AdminClient methods
- It adds the optional timeout where it missed
- It adds / fix the default values for boolean, according to Java option classes


svroonland (changes requested):
Thanks @mdulac !

Could you change the parameter order for better backwards compatibility? For example `CreatePartitionsOptions(validateOnly, retryOnQuotaViolation, timeout)` will keep `validateOnly` as the first argument and the others are optional. 

For consistency I suppose that means having the `timeout` parameter always last.

svroonland (approved):
Looks good, thanks!

PR #437, opened 2022-03-06
Handle RebalanceInProgressExceptions transparently

### LINKED ISSUE #413: Proper commits during cooperative rebalances
## Current commit logic
It is [assumed](https://github.com/zio/zio-kafka/blob/68977a1b93f3b249e28b9ecac8cae3dee457c787/src/main/scala/zio/kafka/consumer/internal/Runloop.scala#L61-L62) that a rebalance is started whenever the `onPartitionsRevoked` method is called and is finished whenever the `onPartitionsAssigned` method is called. [Afterwards](https://github.com/zio/zio-kafka/blob/68977a1b93f3b249e28b9ecac8cae3dee457c787/src/main/scala/zio/kafka/consumer/internal/Runloop.scala#L338) the library tries to commit pending messages if there is no ongoing rebalance.

## The problem
Described behaviour is correct for the eager rebalance. However, the assumptions about a rebalance start/end are incorrect for the cooperative rebalance strategy (which is [going](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=177048248) to become a default):
1. `onPartitionsRevoked` is called at the end of the rebalance. From the [docs](https://kafka.apache.org/28/javadoc/org/apache/kafka/clients/consumer/ConsumerRebalanceListener.html#onPartitionsRevoked(java.util.Collection)):
> In eager rebalancing, it will always be called at the start of a rebalance and after the consumer stops fetching data. In **cooperative rebalancing, it will be called at the end** of a rebalance on the set of partitions being revoked iff the set is non-empty
2. `onPartitionsAssigned` is called more than once, since a cooperative rebalance consist of multiple phases

This results in a consumer attempting commits while a rebalance is still in progress, which triggers an unhandled `RebalanceInProgressException`.

## Possible solution
To address the issue we could implement retries whenever this exception bubbles up.

Also, it might be a good idea to override the `onPartitionsLost` method: its invocation usually means that something's going wrong and the cleanup should be started (e.g. offsets commit). The default implementation of this method is equivalent to the `onPartitionsRevoked` call and is considered as a rebalance start

svroonland:
Fixes #413. Also added partitions lost diagnostic event


erikvanoosten:
This PR should be closed as it is superseded by #427 .

svroonland:
@erikvanoosten Isn't that only the case for when `restartStreamsOnRebalancing` is set to true?

erikvanoosten:
Ah, sorry. In that case I misunderstood the comments on #427.

vigoo:
@svroonland are you planning to finish this?

svroonland:
@vigoo I'll try to make some time

svroonland:
Managed to write a test that reproduces the `RebalanceInProgressException`, but the `catchAll` logic needs to be moved to the offset commit callback somehow, the current logic does not catch it. Because committing is async, new pending commits with higher offsets may have been added to the `State` and we need to avoid that we commit those higher offsets first and then the lower offsets. We need some way to guarantee the correct order of pending commits.

svroonland:
I wonder what is the expected behavior for retrying commits on partitions that were revoked after the rebalance..? If I don't commit fast enough, I may have a queue of offsets to commit for a partition that is now already being processed by another consumer in the group, starting from the last committed offset. 

cc @susliko

svroonland:
@vigoo Ready for review :)

vigoo (approved):

vigoo:
Thanks!

PR #440, opened 2022-03-11
Documents usage of transactional producer

erikvanoosten:
The code was not tested, a thorough review of the authors is appreciated.

### DIFF
--- a/docs/gettingStarted.md
+++ b/docs/gettingStarted.md
@@ -147,6 +147,51 @@ val consumeProduceStream = Consumer
   .provideSomeLayer(consumerAndProducer)
 ```
 
+## Example: consuming, producing and committing offset transactionally
+
+This example does the same as the previous, except that messages are produced in a transaction. A transactional message is committed by committing the corresponding consumer offset. Consumers that use the isolation strategy "read-committed", will only read committed messages.
+
+Again messages are consumed from topic `topic_a` and produced to `topic_b`. Also here processing is done in chunks for more efficiency.
+
+```scala
+import zio.Chunk
+import zio.ZLayer
+import zio.kafka.consumer._
+import zio.kafka.producer._
+import zio.kafka.serde._
+import org.apache.kafka.clients.producer.ProducerRecord
+
+val consumerSettings: ConsumerSettings = ConsumerSettings(List("localhost:9092")).withGroupId("group")
+val producerSettings: TransactionalProducerSettings =
+  TransactionalProducerSettings(List("localhost:9092"), "transaction-1")
+
+val consumerAndProducer =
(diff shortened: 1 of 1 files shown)


erikvanoosten:
@iravid I retained the structure of the other example. There is however a way to commit offsets that is slightly more efficient:

```
  .mapChunksZIO { chunk =>
    val records     = chunk.map(_._1)
    val offsetBatch = chunk.foldLeft(OffsetBatch.empty)(_ merge _.2)

    TransactionalProducer.createTransaction.use { t =>
      t.produceChunkBatch[Any, Int, String](records, Serde.int, Serde.string, offsetBatch)
       .as(Chunk(offsetBatch))
  }
  .aggregateAsync(ZTransducer.foldLeft[OffsetBatch, OffsetBatch](OffsetBatch.empty)(_ merge _))
  .run(ZSink.foreach(_.commit))
```

This runs committing on another fiber then producing the messages. So delays in committing does not affect producing messages. (BTW, it would be great if the needed transducer would become part of the library as well.)
Also, in this fragment I tweaked the creation of`offsetBatch` slightly so that it does not need an intermediate chunk.

WDYT?

erikvanoosten:
Hi @gurghet, can you validate the documentation this PR adds please?

erikvanoosten:
So we needed a few weeks to find out that this PR is quite wrong. This documentation is very much needed!
I'll update this PR if we manage to figure it out. Any input is still very welcome.

vigoo:
Sorry I somehow missed this PR before. I will try to add an example of how we use transactions with zio-kafka.

erikvanoosten:
@vigoo We have something that works but it is quite ugly ('work' pending #469 😉). There are almost 2 pages of control code while we only need to consume, map and produce/commit.

I keep thinking we should be able to totally hide the generations from the user of the library. Since it is possible the user's code combines messages from multiple generations (and commit them together), the only safe way to do this is by not offering messages from a new generation from the consumer, until all messages from the previous generation have been committed.

Roughly like this:
• Upon a re-balance, we store the offsets for the current generation and we stop offering messages to the consumer.
• As soon as those offsets are all committed, we unblock the consumer.

WDYT?

jdegoes:
@vigoo @erikvanoosten 

What's the status of this?

erikvanoosten:
Unless @vigoo wants to spend some more time on this I propose we close this issue. The text I had written is not correct. In addition, I am following @svroonland 's work (helping out where I can, which unfortunately is not much, this code is very complex) that later can be expanded to make this problem go away completely.

erikvanoosten:
Lacking proper documentation for now, the transaction test added in #644 can be used for learning purposes.

PR #454, opened 2022-04-09
ZIO 2.0.0-RC5

vigoo:


src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -193,21 +193,40 @@ object Consumer {
...
-          _.map { case (tp, partition) =>
-            val partitionStream =
-              if (settings.perPartitionChunkPrefetch <= 0) partition
-              else partition.bufferChunks(settings.perPartitionChunkPrefetch)
-
-            tp -> partitionStream.mapChunksZIO(_.mapZIO(_.deserializeWith(keyDeserializer, valueDeserializer)))
-          }
+    ): Stream[Throwable, Chunk[(TopicPartition, ZStream[R, Throwable, CommittableRecord[K, V]])]] = {
+      val partitions = runloop.partitions
+      val stream =
```
svroonland, line 198: Any particular reason for this change?

svroonland (approved):
LGTM, assuming the build becomes green of course. One question about one of the changes

src/main/scala/zio/kafka/consumer/Consumer.scala
vigoo: Yes, there is no `mapZIO` on `Queue` anymore so I basically mixed the old `Queue#mapZIO` and `ZStream#fromQueue` to achieve the same. Would be nice if there is a better solution :)

adamgfraser: If you want to `Queue` is a trait now so you can implement a new queue that delegates to the old one if you really need that functionality. I think I had to do that one place in ZIO Stream.

src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -409,8 +409,7 @@ private[consumer] final class Runloop(
       }
       .onError(cause => partitions.offer(Take.failCause(cause)))
       .unit
-      .toManaged
-      .fork
+      .forkScoped
```
adamgfraser, line 412: This needs to be `forkScoped` because it was previously `ZManaged#fork`.

adamgfraser (approved):

PR #471, opened 2022-05-27
Add `case` to `AclOperation` members

zalbia:
Would be nice to have this change shipped in `0.x` as well. 🙏


svroonland (approved):
Thanks @zalbia !

PR #472, opened 2022-05-27
Add consumeRecordWith function

duxet:
I've added a new variant of `consumeWith` function, that provides access to more low-level ConsumerRecord. It should be useful when you need to access not only key and value, but also some metadata like record headers or partition number.


svroonland:
Maybe this should just replace `consumeWith`. Ease of use is hardly reduced when not taking `(K,V)` but `ConsumerRecord[K,V]` as input for the handler function. What do you think?

duxet:
I think it would be better, as the logic for both functions is exactly the same and this parameter type is the only difference. The only downside is that such change would be a breaking one for all existing zio-kafka consumers, but it shouldn't be hard to migrate though 🤔

PR #477, opened 2022-06-28
Add withCredentials method to admin, consumer and producer settings

### LINKED ISSUE #476: Add helper method for SSL authentication
FS2 has [withCredentials](https://github.com/fd4s/fs2-kafka/blob/series/2.x/modules/core/src/main/scala/fs2/kafka/ConsumerSettings.scala#L404) helper method for SSL authentication. It would be nice to have something similar in zio-kafka, since SSL authentication is quite popular.

mhodovaniuk:
Add withCredentials method to admin, consumer and producer settings.

Fixes #476.


mhodovaniuk:
Looks like CI failed due to a flaky test.

vigoo (approved):

PR #478, opened 2022-06-29
Stop forking tests

swoogles:
I haven't been able to dig up much detail about this setting and everything it impacts, but I discussed with Adam and he thought it was worth proposing here.

When the tests are forked, we don't get the summary at the end, and the developer needs to scroll up through the (potentially huge) output in order to find it-

Screen Shot 2022-06-28 at 9 08 27 PM

When they aren't forked, we get the summary as expected.

Screen Shot 2022-06-28 at 9 09 37 PM

I would love to hear feedback about why this setting was originally enabled, as it might give some clues about how to handle this situation in ZIO itself.

Ideally, we will get ZIO to support this setting, but I do not know how difficult/long that will be. So turning it off in this project would be a quick band-aid, as long as it doesn't cause other issues.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -19,7 +19,6 @@ inThisBuild(
     scalaVersion             := mainScala,
     crossScalaVersions       := allScala,
     Test / parallelExecution := false,
-    Test / fork              := true,
     run / fork               := true,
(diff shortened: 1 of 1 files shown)


swoogles:
Closing because we have resolved this situation in ZIO code.

PR #486, opened 2022-07-21
Add describeConfigsAsync to AdminClient

azhur:
This allows per resource error handling, ie the user can filter out resources that kafka-user doesn't have access to.

### DIFF
--- a/src/main/scala/zio/kafka/admin/AdminClient.scala
+++ b/src/main/scala/zio/kafka/admin/AdminClient.scala
@@ -116,6 +116,14 @@ trait AdminClient {
   ): Task[Map[ConfigResource, KafkaConfig]]
 
   /**
+   * Get the configuration for the specified resources async.
+   */
+  def describeConfigsAsync(
+    configResources: Iterable[ConfigResource],
+    options: Option[DescribeConfigsOptions] = None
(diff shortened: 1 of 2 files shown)


src/main/scala/zio/kafka/admin/AdminClient.scala
```diff
@@ -115,6 +115,14 @@ trait AdminClient {
     options: Option[DescribeConfigsOptions] = None
   ): Task[Map[ConfigResource, KafkaConfig]]
 
+  /**
+   * Get the configuration for the specified resources async.
```
erikvanoosten, line 119:
```suggestion
   * Gives the configuration for the specified resources.
```
There is no need to state that the response is asynchronously.

erikvanoosten (commented):
I am not a committer of this library. Feel free to ignore me 🙂

src/main/scala/zio/kafka/admin/AdminClient.scala
azhur: then the description will be the same as for the existing `describeConfigs` method, that I wanted to avoid

strokyl:
@iravid  is it possible to know if we have any chance to merge that or you have no interest in it?

vigoo (approved):

Issue #155, opened 2020-04-18
zio-kafka-testkit


mvillafuertem:
Hi @iravid , maybe a little module with test utils can be interesting, what do you think? especially embedded kafka or another version with kafka test container.

iravid:
@mvillafuertem Sure. What did you have in mind?

mvillafuertem:
Hi again @iravid, I am not a ZIO expert like you, but first I thought about starting to create another implementation with TestContainer and testing your current test with this implementation, then splitting all this code into another module like ZLayers and add asserts, utils to create topic, groupId, Kafka cluster.

saraiva132:
I have my own test-kit based on the original utils in test. I extended with a couple more things useful for my own usages. Perhaps the answer would be to extend that and eventually when we are happy enough with the result and it is generic enough extract that to its own module?

guizmaii:
@saraiva132 Is it open-source? I'd love to see your work.

tewecske:
Now there is a test utils module. :)
If you need more stuff in it open a ticket, or add to this one maybe.

Issue #61, opened 2019-11-11
a home for test utils?


TimPigden:
adapting the tests for zio-test showed a need for non-trivial support methods to provide a proper providedManagedShared environment for running the tests. It's hard to see how anyone using the library and zio-test will not have the same issues. For example, practically everything needs the live clock.
Is there a place for some sort of kafka-test-util?

adamgfraser:
@TimPigden One thing that might be worth thinking about here is if you want to do something like:

```scala
trait KafkaSpec extends RunnableSpec
```

Right now you are using `DefaultRunnableSpec` which provides a `TestEnvironment`, so you need to use `provideManagedShared` to build an environment that also has the Kafka service (you can probably make this a little more efficient now that RC17 is out using `provideSomeManagedShared`). This definitely works, but the other approach you could consider is creating your own `KafkaSpec` trait that automatically provides the `TestEnvironment` with the Kafka service.

To do that you would extend `RunnableSpec` like shown above and then implement your own `TestRunner`. You can use make a `TestRunner` by providing a `TestExecutor`, which is just a function that knows how to execute specs requiring a certain environment. The default implementation in `DefaultTestExecutor.scala` is only about 20 lines of code.

I'm not sure if this is better or not but it is something that might be worth thinking about.

tewecske:
Now there is a place for test utils :)

erikvanoosten:
Test support is available now via the zio-kafka-test-utils library.

PR #491, opened 2022-08-03
Introduce zio kafka test

tewecske:
What do you think about this? I think these test utils are useful for anyone using zio-kafka so how about publishing them? This is a preview and I'm open to any suggestions especially as I'm not an sbt or zio expert at all :)


tewecske:
@iravid
I actually forgot to check the issues before I started this but this is basically the base for https://github.com/zio/zio-kafka/issues/61 and https://github.com/zio/zio-kafka/issues/155
When I made a sample app for zio-kafka I was wishing for something like this :)
This is a draft so I'll check what else can be done here and also open for any suggestions.

LeonardMeyer:
This is something we'd need as well. Long story short we used Testcontainers at work for integration testing but ran into some weird issues with it. I thought it would've been nice if zio-kafka provided stubs for `Producer` and `Consumer` using the same in memory queue (ZIO has one). We started doing it for `Producer` and it looked easy enough but `Consumer` seems a bit trickier with the multiple method and polling mechanism. 

But this PR looks even better and more general, what is the hold up to merge this besides the conflicts ?

tewecske:
@LeonardMeyer I just didn't get any response if this is good or not. I can clean this up and if necessary add more stuff to it.

LeonardMeyer:
@iravid Could we move this forward ? This is a nice addition to the lib IMO.

@tewecske I saw a few things in the PR that should be changed but maybe because it hasn't been rebased ? The most important one are the produce methods in `KafkaTestUtils` that should somewhat mirror those in [Producer](https://github.com/zio/zio-kafka/blob/master/src/main/scala/zio/kafka/producer/Producer.scala) by being generic and taking serializers directly.

tewecske:
@LeonardMeyer I don't want to change any code in this PR if not necessary to keep it simple. This is only an extraction of test utils into a separate module.
I will clean this up and resolve conflicts and ask for a merge.
If you have ideas for additional work then create a new ticket please. I can work on that after this one is done.

vigoo:
@tewecske can you make this up to date? I merged a couple of open PRs so there are conflicts

tewecske:
@vigoo Updated.

vigoo (approved):

vigoo (approved):

tewecske:
@vigoo Thanks for the merge! :)

@LeonardMeyer If you want more stuff in test utils open a new ticket please.

Issue #212, opened 2020-08-31
consuming without groupid


axaluss:
i'd love to consume messages without group id. 

Zio kafka does not seem to allow for that

svroonland:
Hi @axaluss, could you explain why you want to do that?

axaluss:
There are use cases where you want to control the offset and partition assignment in the client. Then you don't need a group id at all.

svroonland:
From the apache kafka docs:

```
group.id
A unique string that identifies the consumer group this consumer belongs to. This property is required if the consumer uses either the group management functionality by using subscribe(topic) or the Kafka-based offset management strategy.
```

So you should be able to consume without group ID by using manual partition assignment + manual/external offset retrieval:
 `Consumer.subscribe(Subscription.manual("my_topic" -> 1, "my_topic" -> 2))` + `OffsetRetrieval.Manual(getOffsets: Set[TopicPartition] => Task[Map[TopicPartition, Long]])`. See the zio-kafka [readme on this topic](https://github.com/zio/zio-kafka#partition-assignment-and-offset-retrieval).

I haven't tried it, but let me know if you run into errors.

PR #514, opened 2022-09-21
avoid throwing catching invalid group id exception

strokyl:
When starting a consumer without a group id, the code is going to throw and catch a lot of time the same exception. This is not really good for performance

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
@@ -351,6 +351,7 @@ object Consumer {
     for {
       wrapper <- ConsumerAccess.make(settings)
       runloop <- Runloop(
+                   settings.hasGroupId,
                    wrapper,
                    settings.pollInterval,
                    settings.pollTimeout,
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
@@ -39,6 +39,9 @@ case class ConsumerSettings(
   def withGroupId(groupId: String): ConsumerSettings =
     withProperty(ConsumerConfig.GROUP_ID_CONFIG, groupId)
 
+  private[consumer] def hasGroupId: Boolean =
+    properties.contains(ConsumerConfig.GROUP_ID_CONFIG)
+
   def withGroupInstanceId(groupInstanceId: String): ConsumerSettings =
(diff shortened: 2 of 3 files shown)


src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -218,9 +219,17 @@ private[consumer] final class Runloop(
...
-          concatenatedChunk.map(CommittableRecord(_, commit(_), Try(consumer.consumer.groupMetadata()).toOption))
-        )
+        fulfillAction = fulfillAction *> req.cont.succeed(concatenatedChunk.map { record =>
+          CommittableRecord(
+            record = record,
+            commitHandle = commit,
+            consumerGroupMetadata =
+              if (hasGroupId)
+                try Some(consumer.consumer.groupMetadata())
+                catch { case _: Throwable => None }
```
erikvanoosten, line 229: Catching everything is not recommended
```suggestion
                catch { case NonFatal(_) => None }
```

Or simply: ```Try(consumer.consumer.groupMetadata()).toOption```

strokyl: Well catch thanks!

strokyl: `Try(..)` will allocate a try object to discard it right away, minimal optim but better not doing that somewhere we might pass a lot of time.

strokyl:
Test failing:

>   - Consumer Streaming - partitions for topic doesn't fail if doesn't exist

Maybe it's a flaky?

vigoo:
Isn't specifying group id mandatory?

azhur:
@vigoo it's not. You can create a consumer without `group.id` when you don't need to spread the consumption to many consumers and don't want to store the offsets in Kafka.

strokyl:
> Isn't specifying group id mandatory?

See:
https://github.com/zio/zio-kafka/issues/212
And:
https://github.com/zio/zio-kafka/pull/376
For more information about consuming without group id

strokyl:
@erikvanoosten I reworked a bit the PR (see last commit) while rebasing

strokyl:
Failing test:
```
 [info]       - it should not fail all the stream when error while fetching topic record stat
[info] Timeout of 30 s exceeded.
```

erikvanoosten:
> @erikvanoosten Are you still interested in this contribution?

I have no say in that. I am not a maintainer of this project. 😀

vigoo (approved):

vigoo:
Thanks for the explanation

Issue #217, opened 2020-09-24
Support multiple streams with a shared consumer


svroonland:
Sharing a single `KafkaConsumer` instance from the apache kafka library would be more efficient when consuming from multiple Kafka topics. Each instance allocates a thread pool and maintains connections with the Kafka brokers.

svroonland:
Looked into alpakka-kafka a bit and apparently it only supports this feature for manual topic-partition assignments.

We could support a larger class of subscriptions, but I don't see how to for the ones for which the union of two `Subscription`s is not defined as below:

```scala
  def union(s1: Subscription, s2: Subscription): Subscription = (s1, s2) match {
    // Same types
    case (Pattern(p1), Pattern(p2)) => Pattern(s"$p1|$p2".r)
    case (Topics(t1), Topics(t2))   => Topics(t1 ++ t2)
    case (Manual(tp1), Manual(tp2)) => Manual(tp1 ++ tp2)
    // Combinations
    case (Pattern(p1), Topics(t1))  => union(s1, Pattern(t1.mkString("|").r))
    case (Pattern(t1), Manual(tp2)) => ???
    case (Topics(t1), Manual(tp2))  => ???
    // Mirror combinations
    case (Topics(t1), Pattern(p1)) => union(s2, s1)
    case _                         => ???
  }

```

svroonland:
More feasible would be to have users subscribe to the union of topics they are interested in manually. Then have an optional `subscription` parameter on `partitionedStream` that would act as a filter on the topic-partitions of interest. A remaining challenge is to distribute the messages to two or more streams and how to prevent double-committing of offsets if there is overlap in the subscriptions. Or find some way to prevent that.

svroonland:
> A remaining challenge is to distribute the messages to two or more streams ...
`ZHub` will be of help there.

PR #517, opened 2022-10-12
Support for multiple subscriptions on the same Consumer

svroonland:
With this PR, a single Consumer instance and its backing apache kafka `Consumer` instance, can support multiple topic/pattern/manual subscriptions (as long as they are of the same type). 

Inspired by https://doc.akka.io/docs/alpakka-kafka/current/consumer.html#sharing-the-kafkaconsumer-instance

* Subscriptions are now passed as parameter to `plainStream`/`partitionedStream`/`partitionedAssignmentStream`.
* Removed `subscribeAnd`, `SubscribedConsumer`, `SubscribedConsumerFromEnvironment`.
* Subscriptions are scoped to the lifetime of the `plainStream`/`partitionedStream`/`partitionedAssignmentStream`.  
* `subscribe` and `unsubscribe` have been removed from the `Consumer` trait in favour of scoped subscriptions.
* When an error occurs downstream of `plainStream/partitionedStream`, the stream may be restarted thanks to the scoped subscriptions. This was not possible previously, requiring to make a new Consumer instance (or perhaps unsubscribe & resubscribe).

Implements #217.


guizmaii:
@svroonland Can you rebase your branch, please?

svroonland:
The only merge conflict is in some docs. Let's gather some feedback on these changes first before merging.

nmcb (commented):
by lack of understanding - given the description of this pr - and difference between: restartable `plainStream`, `partitionedStream` and non restartable `partitionedAssignmentStream` i would expect the type signatures and/or their documentation to differ more prominently - am i missing something ?

zio-kafka/src/main/scala/zio/kafka/consumer/InvalidSubscriptionUnion.scala
```diff
@@ -0,0 +1,4 @@
+package zio.kafka.consumer
+
+case class InvalidSubscriptionUnion(subscriptions: Seq[Subscription])
```
guizmaii, line 3:
```suggestion
final case class InvalidSubscriptionUnion(subscriptions: Seq[Subscription])
```

zio-kafka/src/main/scala/zio/kafka/consumer/InvalidSubscriptionUnion.scala
```diff
@@ -0,0 +1,4 @@
...
+    extends Exception(s"Unable to calculate union of subscriptions: ${subscriptions.mkString(",")}")
```
guizmaii, line 4:
```suggestion
    extends RuntimeException(s"Unable to calculate union of subscriptions: ${subscriptions.mkString(",")}")
```

zio-kafka-test/src/test/scala/zio/kafka/ConsumerSpec.scala
```diff
@@ -56,33 +56,14 @@ object ConsumerSpec extends ZIOSpecWithKafka {
...
-                     .subscribeAnd(Subscription.Topics(Set(topic)))
-                     .plainStream(Serde.string, Serde.string)
+                     .plainStream(Subscription.Topics(Set(topic)), Serde.string, Serde.string)
                      .take(100)
                      .mapChunks(c => Chunk(c.size))
                      .runCollect
                      .provideSomeLayer[Kafka](consumer(client, Some(group)))
         } yield assert(sizes)(forall(isGreaterThan(1)))
       },
-      test("Consumer.subscribeAnd works properly") {
```
guizmaii, line 67: 🟢 Why removing this test? 🤔

zio-kafka/src/main/scala/zio/kafka/consumer/Subscription.scala
```diff
@@ -75,4 +74,21 @@ object Subscription {
...
   def manual(topic: String, partition: Int): Manual =
     manual((topic, partition))
+
+  def union(s1: Subscription, s2: Subscription): Option[Subscription] = (s1, s2) match {
+    case (Pattern(p1), Pattern(p2)) => Some(Pattern(s"$p1|$p2".r))
+    case (Topics(t1), Topics(t2))   => Some(Topics(t1 ++ t2))
+    case (Manual(tp1), Manual(tp2)) => Some(Manual(tp1 ++ tp2))
+    case _ =>
+      None // Although we could combine Pattern and Topics, the Kafka consumer will throw exceptions when changing the subscription type
+  }
```
guizmaii, line 84: Seeing that it's not possible to combine two `Subscription` which are of different type, I believe that we would be able to avoid the `Option` here by adding a phantom type on the `Subscription` trait:

```
sealed trait Subscription[Kind <: SubscriptionKind]
object Subscription {
  final case class Topics(topics: Set[String])                  extends Subscription[TopicsKind]
  final case class Pattern(pattern: Regex)                      extends Subscription[PatternKind]
  final case class Manual(topicPartitions: Set[TopicPartition]) extends Subscription[ManualKind]

  sealed trait SubscriptionKind
  case object TopicsKind extends SubscriptionKind
  case object PatternKind extends SubscriptionKind
  case object ManualKind extends SubscriptionKind

  type TopicsKind = TopicsKind.type
  type PatternKind = PatternKind.type
  type ManualKind = ManualKind.type
  
  ...
}
```

zio-kafka-test/src/test/scala/zio/kafka/ConsumerSpec.scala
svroonland: It's because the whole SubscribedConsumer has been removed

guizmaii:
I tried to find a way to be more typesafe: no more runtime error possible with the "union" operation, but I didn't find anyway because in the end you're just composing streams 🤔

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -191,50 +216,60 @@ object Consumer {
...
-        }
+      def extendSubscriptions = subscriptions.updateZIO { existingSubscriptions =>
+        val newSubscriptions = NonEmptyChunk.fromIterable(subscription, existingSubscriptions)
+        ZIO
+          .fromOption(Subscription.unionAll(newSubscriptions))
+          .orElseFail(InvalidSubscriptionUnion(newSubscriptions.toSeq))
+          .flatMap { union =>
+            ZIO.logInfo(s"Changing kafka subscription to $union") *>
+              subscribe(union).as(newSubscriptions.toSet)
+          }
```
guizmaii, line 231:
```suggestion
        Subscription.unionAll(newSubscriptions) match {
          case None => ZIO.fail(InvalidSubscriptionUnion(newSubscriptions.toSeq))
          case Some(union) =>
            ZIO.logInfo(s"Changing kafka subscription to $union") *>
              subscribe(union).as(newSubscriptions.toSet)
        }
```

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -360,7 +389,14 @@ object Consumer {
                    settings.rebalanceListener,
                    settings.restartStreamOnRebalancing
                  )
-    } yield Live(wrapper, settings, runloop)
+      subscriptions <- Ref.Synchronized.make(Set.empty[Subscription])
+
+      partitionAssignments <- ZStream
+                                .fromQueueWithShutdown(runloop.partitions)
```
guizmaii, line 395: Are we sure we want to use `fromQueueWithShutdown` here? The shutdown of the queue is already probably ensured by the `RunLoop`, no? 🤔

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -360,7 +389,14 @@ object Consumer {
...
+                                .map(_.exit)
+                                .flattenExitOption
+                                .toHub(32)
```
guizmaii, line 398: Why `32` here? 🤔

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -191,50 +216,60 @@ object Consumer {
...
+        for {
+          stream <- ZStream.fromHubScoped(partitionAssignments)
+          _      <- extendSubscriptions.withFinalizer(_ => reduceSubscriptions.orDie)
+        } yield stream
+          .map(_.exit)
+          .flattenExitOption
+          .flattenChunks
+          .map {
+            _.collect {
+              case (tp, partition) if Subscription.subscriptionMatches(subscription, tp) =>
```
guizmaii, line 256: What happens if `Subscription.subscriptionMatches(subscription, tp)` is false? 🤔

svroonland:
> by lack of understanding - given the description of this pr - and difference between: restartable plainStream, partitionedStream and non restartable partitionedAssignmentStream i would expect the type signatures and/or their documentation to differ more prominently - am i missing something ?
@nmcb The changes in this PR for the documentation for these three methods is indeed the same. The existing documentation and type signatures are only subtly different, since the methods are three ways to get the same data from Kafka (the `CommittableRecord`s) but with some additional structure related to partitions and rebalancing.

jdegoes:
@svroonland @guizmaii What's the status of this one?

svroonland:
@jdegoes This one is ready to be merged. I was hoping to get some more feedback on the interface changes because it's a breaking change, but the feedback I've had so far is good.

svroonland:
If we want to merge this, I suggest to do it along with #620 since that is also a non-compatible change.

adamgfraser:
@svroonland Can you resolve conflicts?

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
svroonland: You are right

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
svroonland: Well since it's `collect` with a partial function on a `Chunk`, the `(tp, partition)` tuple will not be included in the resulting chunk.

svroonland:
Not sure how to resolve the readme error, the suggested patch is not desirable:

```
-libraryDependencies += "dev.zio" %% "zio-kafka" % "2.0.7"
+libraryDependencies += "dev.zio" %% "zio-kafka" % "0.17.8"
```

adamgfraser:
@khajavi Can you take a look?

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
svroonland: Maybe 1 is enough here, I'd have to think about it a bit more and run tests.

guizmaii: Can we maybe document this magic number to explain a bit better what its influence is on the behaviour of the code?

svroonland: It's hard to describe in a couple of words.

Under most circumstances, a value of 1 would be fine here. Values are coming from `runloop.partitions` which is already an unbounded Queue, no need for an additional buffer.

But consider this: say there are two streams for two separate subscriptions. If one of them is not consuming partition assignments fast enough for some reason, it could backpressure the Hub and lead to the situation where the other stream gets delayed partition assignments. A larger hub capacity here could accomodate for these speed differences. Since partition assignments and rebalances are rather infrequent, this capacity does not need to be high, 32 seems high enough. I'm a bit hesitant to make it configurable since it's another confusing setting for users that should not need to be tuned..

What the effects on the above scenario would be if one of the subscription streams is processing records transactionally (see #644), I'm not sure. But it would be safe anyway to recommend to use a single-subscription consumer in such a case.

docs/sharing-consumer.md
```diff
@@ -0,0 +1,42 @@
+---
+id: sharing-consumer
+title: "Sharing a Consumer between multiple streams"
+---
+
+Often in a single application, you want to consume from multiple Kafka topics and process each topic in a distinct way. With `zio-kafka` you can use a single `Consumer` instance for multiple streams from different topics. It is not only easier to create a single `Consumer` layer instead of one for each topic, but it may be more resource efficient as well. The underlying Apache Kafka consumer, its thread pool and communication with the Kafka brokers will be shared, resulting in less resource consumption compared to when you create a Consumer instance for every topic.
+
+For each of the topics / patterns you subscribe to, you can define your own stream to process records as well as a `Deserializer` for the keys and values. Other settings like poll interval and offset strategy are common to each of the subscriptions. For example, the value of the `max.poll.records` setting is the maximum number of records returned in each poll for all subscriptions combined. If you need different settings for each topic, you should create a `Consumer` instance per topic.
```
guizmaii, line 8: I'm not a native speaking English speaker but I'd like to suggest these changes

```suggestion
For each of the topics/patterns you subscribe to, you can define a dedicated stream to process records, a dedicated `Deserializer` for the keys and a dedicated `Deserializer` for the values. Other settings like poll interval and offset strategy are common to all subscriptions. For example, the value of the `max.poll.records` setting is the maximum number of records returned in each poll for all subscriptions combined. If you need different settings for each topic/pattern, you need to create a `Consumer` instance per topic/pattern.
```

docs/sharing-consumer.md
```diff
@@ -0,0 +1,42 @@
...
+  .mapZIO(_.commit)
+  .runDrain
+
+// Run both streams with the same consumer instance
+(stream1 zipPar stream2).provideSomeLayer(consumer)
+```
+
+Each of the streams can have different lifetimes. Subscriptions are tied to the Stream's scope, so when the stream completes (succesfully, via failure or interruption), that stream's subscription is removed. 
+
+Consumer sharing is only possible when using the same type of `Subscription`: `Topics`, `Pattern` or `Manual`. Mixing different types will result in a failed stream. Note that every `Topic` subscription can be written as a `Pattern` subscription as well, so if you want to mix `Topic` and `Pattern` subscriptions, use just `Pattern`.
```
guizmaii, line 40: Some suggestions

```suggestion
Consumer sharing is only possible when using the same type of `Subscription`: `Topics`, `Pattern` or `Manual`. Mixing different types will result in a failed stream. Note that every `Topic` subscription can also be written as a `Pattern` subscription, so if you want to mix `Topic` and `Pattern` subscriptions, you can just use `Pattern`. 
```

zio-kafka-test/src/test/scala/zio/kafka/ConsumerSpec.scala
```diff
@@ -849,6 +813,167 @@ object ConsumerSpec extends ZIOKafkaSpec {
...
+              Consumer
+                .plainStream(Subscription.topics(topic1), Serde.string, Serde.string, 32)
+                .take(5)
+                .tap(_ => ZIO.unlessZIO(errored.getAndSet(true))(ZIO.fail(new RuntimeException("Stream failure 1"))))
+                .runCollect
+                .retryN(1)
+                .provideSomeLayer[Kafka with Scope](consumer(client, Some(group)))
+          } yield assertCompletes
+        }
+      ),
```
guizmaii, line 976: 🟢 What do you think about moving this suite to another file? This one already quite big

guizmaii, line 976: 🟢 What do you think about moving this suite to another file? This one already quite big

zio-kafka-test/src/test/scala/zio/kafka/ConsumerSpec.scala
```diff
@@ -849,6 +813,167 @@ object ConsumerSpec extends ZIOKafkaSpec {
...
+                    Serde.string,
+                    32
+                  )
+                  .runCollect)
+                .provideSomeLayer[Kafka & Scope](consumer(client, Some(group)))
+                .unit
+                .exit
+          } yield assert(result)(fails(isSubtype[InvalidSubscriptionUnion](anything)))
+        },
+        test("distributes records (randomly) from overlapping subscriptions over all subscribers") {
```
guizmaii, line 905: Hum. This behaviour can be very surprising and hard to debug 🤔

Why do we need to have this behaviour? Can't we send the records from the overlapped topic(s) to both streams? 🤔

zio-kafka-test/src/test/scala/zio/kafka/ConsumerSpec.scala
```diff
@@ -849,6 +813,167 @@ object ConsumerSpec extends ZIOKafkaSpec {
...
+
+            _ <-
+              (Consumer
+                .plainStream(Subscription.topics(topic1), Serde.string, Serde.string, 32)
+                .take(100)
+                .merge(
+                  Consumer
+                    .plainStream(Subscription.topics(topic2), Serde.string, Serde.string, 32)
+                    .take(50) *> ZStream.fromZIO(produceMany(topic1, kvs))
+                )
```
guizmaii, line 951: 🥇 Smart way of testing this 🙂

It took me 2min to understand 😅

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -327,7 +327,14 @@ private[consumer] final class Runloop(
...
       else 0.millis.asJava
 
-    val records = c.poll(pollTimeout)
+    val records =
+      try
+        c.poll(pollTimeout)
+      catch {
+        case _: IllegalStateException =>
+          // When not subscribed, poll() will throw an exception. This can happen in a race condition between calling markUnsubscribed and calling handlePoll()
+          null
```
guizmaii, line 336: Isn't this a bit risky? 🤔
What if the `.poll` call throw an `IllegalStateException` for another important reason? 🤔

zio-kafka-test/src/test/scala/zio/kafka/ConsumerSpec.scala
erikvanoosten: Kafka works with consumer-group-ids to control this. By necessity a single Kafka consumer will have a single consumer-group-id. So here the user has no way of controlling this behavior and either behavior can be unexpected.

This let me to consider the question: why would we allow overlapping subscriptions in the first place?

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
guizmaii: 🔴 Can we put some doc in the code about this value, please?

So we'll remember why we use it and decided to use it

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -191,50 +216,59 @@ object Consumer {
...
+        val newUnion         = newSubscriptions.flatMap(Subscription.unionAll)
 
-          tp -> partitionStream.mapChunksZIO(_.mapZIO(_.deserializeWith(keyDeserializer, valueDeserializer)))
-        }
+        (newUnion match {
+          case Some(union) =>
+            ZIO.logInfo(s"Reducing kafka subscription to $union") *> subscribe(union)
+          case None =>
+            ZIO.logInfo(s"Unsubscribing kafka consumer") *> unsubscribe
+        }).as(newSubscriptions.map(_.toSet).getOrElse(Set.empty))
```
guizmaii, line 242: 🟢
```suggestion
        }).as(newSubscriptions.fold(Set.empty[Subscription])(_.toSet))
```

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -191,50 +216,59 @@ object Consumer {
...
-                  chunk => ZIO.succeed(Take.chunk(chunk))
-                )
-              }
-            }
-            .catchAllCause((c: Cause[Nothing]) =>
-              partitions.isShutdown.flatMap { down =>
-                if (down && c.isInterrupted) Pull.end
-                else Pull.failCause(c)
-              }
-            )
```
guizmaii, line 216: The challenging part for me to understand in your PR is this removed part.

Why were we having to repeat `ZStream.repeatZIOChunkOption` and also to manually manage the `partitions.shutdown` and now we're not anymore repeating nor managing the partition shutdown? 🤔

zio-kafka-test/src/test/scala/zio/kafka/ConsumerSpec.scala
guizmaii: > why would we allow overlapping subscriptions in the first place?

Good question. Could we be able to detect overlapping subscriptions? With `topics`, it could be easy but what about `pattern`? 🤔

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
svroonland: Of course. This is a bit of ZIO legacy. The whole `repeatZIOChunkOption { .. }` bit is equivalent to `ZStream.fromQueueWithShutdown(partitions).flattenTake`. I'd have to double check since it's been a while since I made this change, but at least all of the unit tests still work.

Also we're now using a Hub, hence the `ZStream.fromHubScoped` on line 247.

zio-kafka-test/src/test/scala/zio/kafka/ConsumerSpec.scala
svroonland: This whole test is describing 'undefined behavior' in a way. I wanted to make sure that at least nothing crashes. 

I would agree that it would be better to prevent overlapping subscriptions in the first place, but that's not possible with patterns as you say. 

> Can't we send the records from the overlapped topic(s) to both streams?
That would be nice, but these streams are pull-based, so both streams would be making Requests. Each stream would get every other chunk of polled records probably.

Since it's such an edge case to have overlapping subscriptions - the intended use case is just consuming a few topics with the same consumer - a word of caution to the user would be enough imo.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
svroonland: Implemented a more 'thread safe' version in one of the last commits.

zio-kafka-test/src/test/scala/zio/kafka/ConsumerSpec.scala
```diff
@@ -849,6 +813,167 @@ object ConsumerSpec extends ZIOKafkaSpec {
...
+            topic2 <- randomTopic
+            client <- randomClient
+            group  <- randomGroup
+
+            _ <- produceMany(topic1, kvs)
+            _ <- produceMany(topic2, kvs)
+
+            records <-
+              (Consumer
+                .plainStream(Subscription.topics(topic1), Serde.string, Serde.string, 32)
```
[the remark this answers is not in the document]
svroonland: I'm not sure why that is in there, but I don't see a reason. Removed it

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -319,42 +325,17 @@ object Consumer {
...
-        }
-      } *> runloop.markSubscribed
+      changeSubscription(Some(subscription))
 
     private def unsubscribe: Task[Unit] =
-      runloop.markUnsubscribed *> consumer.withConsumer(_.unsubscribe())
+      changeSubscription(None)
+
+    private def changeSubscription(subscription: Option[Subscription]): Task[Unit] =
+      ZIO.runtime[Any].flatMap { runtime =>
```
guizmaii, line 334: 🟢 What do you think about passing this `runtime` in the `Consumer` constructor like we did for the `RunLoop`?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -671,18 +729,18 @@ private[consumer] object Runloop {
...
                   rebalancingRef,
                   diagnostics,
                   shutdownRef,
                   offsetRetrieval,
                   userRebalanceListener,
-                  subscribedRef,
                   restartStreamsOnRebalancing,
                   currentStateRef
                 )
+      _ <- ZIO.addFinalizer(ZIO.logInfo("Shut down Runloop"))
```
guizmaii, line 743: 🌕 I don't think we should log this in out users logs at the INFO level. IMO, they don't care and they shouldn't have to care.

```suggestion
      _ <- ZIO.addFinalizer(ZIO.logDebug("Shut down Runloop"))
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -534,34 +537,84 @@ private[consumer] final class Runloop(
...
-          if (state.pendingRequests.nonEmpty) handlePoll(state)
+          if (state.pendingRequests.nonEmpty && state.isSubscribed) handlePoll(state)
           else ZIO.succeed(state)
         }
       case cmd @ Command.Commit(_, _) =>
         handleCommit(state, cmd)
+      case cmd @ Command.ChangeSubscription(_, _, _, _) =>
+        handleChangeSubscription(state, cmd)
+
     }
```
guizmaii, line 561: 🟢
```suggestion
        handleChangeSubscription(state, cmd)
    }
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -527,34 +537,84 @@ private[consumer] final class Runloop(
...
       case Command.Poll =>
         // The consumer will throw an IllegalStateException if no call to subscribe
-        ZIO.ifZIO(subscribedRef.get)(onTrue = handlePoll(state), onFalse = ZIO.succeed(state))
+        if (state.isSubscribed) handlePoll(state) else ZIO.succeed(state)
       case Command.Requests(reqs) =>
         handleRequests(state, reqs).flatMap { state =>
           // Optimization: eagerly poll if we have pending requests instead of waiting
           // for the next scheduled poll.
-          if (state.pendingRequests.nonEmpty) handlePoll(state)
+          if (state.pendingRequests.nonEmpty && state.isSubscribed) handlePoll(state)
```
guizmaii, line 553: What happens to the `pendingRequests` when we're not subscribed anymore? 🤔

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -72,6 +71,20 @@ private[consumer] final class Runloop(
...
+  def changeSubscription(
+    subscription: Option[Subscription],
+    offsetRetrieval: OffsetRetrieval,
+    rebalanceListener: ConsumerRebalanceListener
+  ): Task[Unit] =
+    Promise
+      .make[Throwable, Unit]
+      .flatMap { cont =>
+        commandQueue.offer(Command.ChangeSubscription(subscription, offsetRetrieval, rebalanceListener, cont)) *>
+          cont.await
```
guizmaii, line 83: It's really not clear in my mind what `Promise#await` does, so sorry if my questions are dumb.

What is this `.await` does? Why do we need it here? Aren't we blocking the `Task` returned by this function? 🤔

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -527,34 +537,84 @@ private[consumer] final class Runloop(
         ZIO.foreachDiscard(reqs)(_.end).as(state)
       case cmd @ Command.Commit(_, _) =>
         handleCommit(state, cmd)
+      case _ @Command.ChangeSubscription(_, _, _, cont) =>
+        cont.succeed(()).as(state)
```
guizmaii, line 541: 🟢 What do you think about "nullifying" the state subscription here?

```suggestion
        cont.succeed(()).as(state.copy(subscription = None))
```

Useless? 🤔

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -273,60 +312,30 @@ object Consumer {
...
+      changeSubscription(Some(subscription))
 
-    override def unsubscribe: Task[Unit] =
-      runloop.markUnsubscribed *> consumer.withConsumer(_.unsubscribe())
+    private def unsubscribe: Task[Unit] =
+      changeSubscription(None)
+
+    private def changeSubscription(subscription: Option[Subscription]): Task[Unit] =
+      ZIO.runtime[Any].flatMap { runtime =>
+        val rc = RebalanceConsumer.Live(consumer.consumer)
```
guizmaii, line 335: 🌕 Why do we need to instantiate this `RebalanceConsumer` and the corresponding `ConsumerRebalanceListener` (ie. `runloop.rebalanceListener.toKafka(runtime, rc)`) here?

Why not move these to the `RunLoop`?

I made a POC to demonstrate this change: https://github.com/zio/zio-kafka/pull/662

guizmaii:
Here's a proposal to improve a bit the readability: https://github.com/zio/zio-kafka/pull/663

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
svroonland: That's correct. The Promise is a mechanism to be able to put it on a queue and get notified when it's actually completed.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
guizmaii: What do you think about adding this line of code in order to be able to detect if it happens (might be a bit extreme):
```scala
        handleRequests(state, reqs).flatMap { state =>
          // Optimization: eagerly poll if we have pending requests instead of waiting
          // for the next scheduled poll.
          if (state.pendingRequests.nonEmpty) {
            if (state.isSubscribed) handlePoll(state)
            else
              ZIO.fail(
                new IllegalStateException(
                  "Should Never Happen: Some requests are pending but the consumer is not subscribed"
                )
              )
          } else ZIO.succeed(state)
        }
```

this kind of defensive code helped me find edge cases/bugs in the past

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -347,7 +352,17 @@ object Consumer {
...
-  ): ZIO[Scope, Throwable, Consumer] =
+  ): ZIO[Scope, Throwable, Consumer] = {
+    /*
+    We must supply a queue size for the partitionAssignments hub below. Under most circumstances,
+    a value of 1 should be sufficient, as runloop.partitions is already an unbounded queue. But if
+    there is a large skew in speed of consuming partition assignments (not the speed of consuming kafka messages)
+    between the subscriptions, there may arise a situation where the faster stream is 'blocked' from
+    getting new partition assignments by the faster stream. A value of 32 should be more than sufficient to cover
+    this situation.
+     */
```
guizmaii, line 363: 🙏

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -525,34 +534,107 @@ private[consumer] final class Runloop(
...
         }
       case cmd @ Command.Commit(_, _) =>
         handleCommit(state, cmd)
+      case cmd @ Command.ChangeSubscription(_, _, _) =>
+        handleChangeSubscription(state, cmd).flatMap { state =>
+          if (state.isSubscribed) {
+            handlePoll(state)
+          } // This updates the assignment and will end partition streams which we are no longer subscribed to
+          else {
+            // End pending requests
```
guizmaii, line 568: 🟢
```suggestion
          if (state.isSubscribed) {
            // This updates the assignment and will end partition streams which we are no longer subscribed to
            handlePoll(state)
          } else {
            // End pending requests
```

guizmaii (dismissed):

guizmaii (approved):

PR #539, opened 2022-11-16
Avoid using a `Task` when code is pure. Prefer to use `Try`

guizmaii:


zio-kafka/src/main/scala/zio/kafka/admin/AdminClient.scala
```diff
@@ -1470,15 +1494,32 @@ object AdminClient {
...
     def toScala: Option[T] = if (v.isPresent) Some(v.get()) else None
   }
 
-  implicit final class OptionOps[T](val v: Option[T]) extends AnyVal {
+  implicit final class OptionOps[T](private val v: Option[T]) extends AnyVal {
     def toJava: Optional[T] = v.fold(Optional.empty[T])(Optional.of)
   }
+
+  implicit final class ListOps[A](private val list: List[A]) extends AnyVal {
+    def forEach[B](f: A => Try[B]): Try[List[B]] = {
```
svroonland, line 1510: It's more scala/cats idomatic to have a `List[Try[A]]` and a `sequence: Try[List[A]]` method. Have a look at `Future.sequence` for example.

zio-kafka/src/main/scala/zio/kafka/admin/AdminClient.scala
```diff
@@ -1470,15 +1494,32 @@ object AdminClient {
       ZIO.succeed(client.close(settings.closeTimeout))
     )
 
-  implicit final class MapOps[K1, V1](val v: Map[K1, V1]) extends AnyVal {
+  implicit final class MapOps[K1, V1](private val v: Map[K1, V1]) extends AnyVal {
```
svroonland, line 1497: IMO the `private` has no added value for these implicit classes, whose only purpose is to define extension methods.

svroonland (changes requested):
Looks fine overall, but a few points

zio-kafka/src/main/scala/zio/kafka/admin/AdminClient.scala
guizmaii: `forEach` comes from zio-prelude ;)

zio-kafka/src/main/scala/zio/kafka/admin/AdminClient.scala
guizmaii: This is just copying how cats and co. are defining implicit class. I know they did a lot of research to find the best way to define them to be sure there is not overhead at runtime, I'm just copying their way of coding here

svroonland: Ah I was not aware of that. Alright

zio-kafka/src/main/scala/zio/kafka/admin/AdminClient.scala
svroonland: Right, makes sense.

svroonland (approved):

jdegoes (approved):

PR #554, opened 2022-11-22
Fix startup issue with high poll intervals

svroonland:
When using high poll interval and timeouts (> 100 ms), the Consumer would not get any assigned partitions and stay idle. 

Exactly what is the mechanism causing this, I haven't been able to figure out. Everything works fine with a poll interval and timeout of 100 ms.


svroonland:
Will reopen when I have a fix that works

guizmaii:
Isn't this the fix you're looking for: https://github.com/zio/zio-kafka/pull/661? 🤔

svroonland:
Unfortunately not, several polls are necessary to do the setup phase with the broker.

Issue #531, opened 2022-11-07
Producer.produceAsync performance


lukestephenson:
I was making use of `produceAsync` in combination with `ZIO.foreach` to achieve batching / sending multiple records in parallel.

While I was expecting this to be slightly slower than the `produceChunkAsync` API, the order of magnitude of which it is slower was not.

For my experiment, I was trying to publish 1 million messages to Kafka (just a single broker / local docker). The timings I saw were:
| Experiment  | Duration |
| ------------- | ------------- |
| ZIO publishing using a chunk aware publisher (ie avoiding more ZIO invocations)  | 1.3 seconds  |
| ZIO publishing one message at a time, and relying on "sequence" to publish batches | 23 seconds  |
| Monix publishing of one message at a time, and relying on "sequence" to publish batches | 3.5 seconds |
| Fs2-kafka publishing one message at a time and relying on sequence. | 6 seconds |

For the monix, fs2-kafka implementations, I tried to keep it as close to ZIO-kafka as I could.

Note the fs2-kafka batch API is implemented in terms of the single publish API (and a `sequence`), so will be basically the same.

The monix implementation (our internal publisher lib, so I can't share, but I'll try to copy the relevant bits into the demo project soon) makes use of `scala.concurrent.blocking` to flag that calling the underlying `KafkaProducer.send` method may block.

One of my colleagues played with the zio-kafka implementation, and found that the majority of the performance hit is from [`ZIO.attemptBlocking`](https://github.com/zio/zio-kafka/blob/cc0c620da288aa87a5451d4a6a446532a1e6e2ff/zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala#L140).

For a properly configured KafkaProducer, the majority of calls to `send` shouldn't block, but the current implementation appears optimised for all the calls blocking, and always shifting execution onto another thread.

The code I used for the experiment is in https://github.com/lukestephenson/kafka-perf.

Now, you might be thinking just use the chunking API and the problem goes away.  Well yes to a certain extent.  But regardless of this I do have serious concerns about just how much overhead using ZIO imposes. Also, the current chunking API doesn't support what we need.  Our project has the ability to drop messages if publishing to kafka fails (eg message too large).  The zio-kafka bulk publish API doesn't tell you which message failed (just that something in the batch failed), or have the ability to continue with the other messages past that point.  So we would first need to improve that. Regardless of that, I don't think the `produceAsync` method should have overheads as high as it does currently.

[...]

PR #555, opened 2022-11-22
Execute KafkaProducer#send calls directly and not on the blocking thread pool

svroonland:
See #531


iravid:
I think this is generally fine but just want to say that in catastrophic cases where the broker is down (for example) and the producer buffer fills up, calls to `send` most definitely will block. I think this is documented in the Java API.

iravid (approved):

svroonland:
Right, found it:
```
When the buffer space is exhausted additional send calls will block. The threshold for time to block is determined by <code>max.block.ms</code> after which it throws a TimeoutException.
```

Thanks!

guizmaii:
Can't we bypass this mechanism from the Java client by implementing our own buffer and using a non-blocking "blocking mechanism"? Might be a crazy idea 🤔😅

svroonland:
I suppose we could yes, but it's quite some effort for an exceptional case. Feel free to do so of course! :)

PR #580, opened 2023-01-10
Replace usage of `with` by `&`

guizmaii:


build.sbt
```diff
@@ -71,7 +72,7 @@ def stdSettings(prjName: String) = Seq(
   name              := s"$prjName",
   scalafmtOnCompile := true,
   Compile / compile / scalacOptions ++= {
-    if (scalaBinaryVersion.value == "2.13") Seq("-Wconf:cat=unused-nowarn:s")
+    if (scalaBinaryVersion.value == "2.13") Seq("-Wconf:cat=unused-nowarn:s", "-Xsource:3")
```
adamgfraser, line 75: ZIO already provides `&` as a type alias for `with` on Scala 2 so I don't think you should need this.

PR #581, opened 2023-01-10
Ensure calls to Producer.produce* do not block

svroonland:
Records to be produced are placed in Chunks in a bounded ZIO Queue, which is processed by a stream running on the Blocking thread pool. The result is communicated back via a Promise. This way we can be sure that calls to `Producer.produce*` never block the ZIO thread pool without incurring the overhead of shifting to the blocking thread pool for every record or chunk of records.

Also
* Close the Producer on the blocking thread pool
* Eliminate serializedRecords from the callback closure


zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -256,20 +251,21 @@ object Producer {
...
+                         props.asJava,
+                         new ByteArraySerializer(),
+                         new ByteArraySerializer()
                        )
-      } yield Live(rawProducer, settings)
-    } { producer =>
-      producer.close
-    }
+                     )
+      runtime <- ZIO.runtime[Any]
```
guizmaii, line 263: +1

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -256,20 +251,21 @@ object Producer {
...
+      sendQueue <-
+        Queue.bounded[(Chunk[ByteRecord], Promise[Throwable, Chunk[RecordMetadata]])](settings.sendBufferSize)
+      producer <- ZIO.acquireRelease(ZIO.succeed(Live(rawProducer, settings, runtime, sendQueue)))(_.close)
```
guizmaii, line 266: Something that has nothing to do with your change but that I spotted while reviewing: The `.close` function is IMO a bit to optimistic, using `ZIO.succeed`. IMO, it should use `ZIO.attempt`. We're calling Java code here so it can always throw and the code of `.close` does indeed throw in some cases

zio-kafka/src/main/scala/zio/kafka/producer/ProducerSettings.scala
```diff
@@ -33,9 +34,11 @@ final case class ProducerSettings(
...
   def withCredentials(credentialsStore: KafkaCredentialStore): ProducerSettings =
     withProperties(credentialsStore.properties)
+
+  def withSendBufferSize(sendBufferSize: Int) = copy(sendBufferSize = sendBufferSize)
 }
 
 object ProducerSettings {
   def apply(bootstrapServers: List[String]): ProducerSettings =
-    new ProducerSettings(bootstrapServers, 30.seconds, Map())
+    new ProducerSettings(bootstrapServers, 30.seconds, 4096, Map())
```
guizmaii, line 43:
```suggestion
    new ProducerSettings(bootstrapServers, 30.seconds, 4096, Map.empty)
```

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -164,40 +151,48 @@ object Producer {
...
+    /**
+     * Calls to send may block when updating metadata or when communication with the broker is (temporarily) lost,
+     * therefore this stream is run on a the blocking thread pool
+     */
+    val sendFromQueue: ZIO[Any, Nothing, Any] =
+      ZStream
+        .fromQueueWithShutdown(sendQueue)
+        .mapZIO { case (serializedRecords, done) =>
+          ZIO.attempt {
+            val it: Iterator[(ByteRecord, Int)] = serializedRecords.iterator.zipWithIndex
```
guizmaii, line 168: Why preferring an iterator here instead of just a call to `.foreach`?

that'd more Scala idiomatic and would be the same code in the end 🤔

```scala
    val sendFromQueue: ZIO[Any, Nothing, Any] =
      ZStream
        .fromQueueWithShutdown(sendQueue)
        .mapZIO { case (serializedRecords, done) =>
          ZIO.attempt {
            val res: Array[RecordMetadata] = new Array[RecordMetadata](serializedRecords.length)
            val count: AtomicLong          = new AtomicLong

            serializedRecords.zipWithIndex.foreach { case (rec, idx) =>
              p.send(
                rec,
                (metadata: RecordMetadata, err: Exception) =>
                  Unsafe.unsafe { implicit u =>
                    (if (err != null) runtime.unsafe.run(done.fail(err)).getOrThrowFiberFailure(): Unit
                     else {
                       res(idx) = metadata
                       if (count.incrementAndGet == serializedRecords.length) {
                         runtime.unsafe.run(done.succeed(Chunk.fromArray(res))).getOrThrowFiberFailure(): Unit
                       }
                     }): @nowarn("msg=discarded non-Unit value")
                    ()
                  }
              )
            }
          }
            .foldCauseZIO(done.failCause, _ => ZIO.unit)
        }
        .runDrain
```

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
svroonland: You're right. However, the finalizer must be a `ZIO[R, Nothing, Any]`, so we need to have an `orDie`.

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
svroonland: Existing code that was just moved. They're functionally similar. The only thing I can think of is that `chunk.iterator.zipWithIndex` is a bit more lazy, instead of allocating a new Chunk.

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
guizmaii: What about: `ZIO.attemptBlocking(...).orDie`?

Moving it to Blocking just to be sure

guizmaii (approved):

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -135,25 +137,10 @@ object Producer {
...
-                         else runtime.unsafe.run(done.succeed(metadata)).getOrThrowFiberFailure(): Unit
-                       ): @nowarn("msg=discarded non-Unit value")
-                       ()
-                     }
-                 }
-               )
-             }
-      } yield done.await
+        _                <- sendQueue.offer((Chunk.single(serializedRecord), done))
+      } yield done.await.map(_.head)
```
guizmaii, line 143: I feel like we don't need all this to just send one event. I'll try to propose something. I was looking at Monix implementation. It's interesting

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -164,40 +151,48 @@ object Producer {
...
+
+              p.send(
+                rec,
+                new Callback {
+                  def onCompletion(metadata: RecordMetadata, err: Exception): Unit =
+                    Unsafe.unsafe { implicit u =>
+                      (if (err != null) runtime.unsafe.run(done.fail(err)).getOrThrowFiberFailure(): Unit
+                       else {
+                         res(idx) = metadata
+                         if (count.incrementAndGet == serializedRecords.length)
```
guizmaii, line 183: What do you think about adding a variable `length` outside of the callback and using it here instead of `serializedRecords.length`?

That'd remove the memory pointer of `serializedRecords` from the callback, potentially allowing the GC to collect the data earlier 🤔

```scala
                    ZIO.attempt {
            val it: Iterator[(ByteRecord, Int)] = serializedRecords.iterator.zipWithIndex
            val length                          = serializedRecords.length
            val res: Array[RecordMetadata]      = new Array[RecordMetadata](length)
            val count: AtomicLong               = new AtomicLong

            while (it.hasNext) {
              val (rec, idx): (ByteRecord, Int) = it.next()

              p.send(
                rec,
                new Callback {
                  def onCompletion(metadata: RecordMetadata, err: Exception): Unit =
                    Unsafe.unsafe { implicit u =>
                      (if (err != null) runtime.unsafe.run(done.fail(err)).getOrThrowFiberFailure(): Unit
                       else {
                         res(idx) = metadata
                         if (count.incrementAndGet == length)
                           runtime.unsafe.run(done.succeed(Chunk.fromArray(res))).getOrThrowFiberFailure(): Unit
                       }): @nowarn("msg=discarded non-Unit value")
                      ()
                    }
                }
              )
            }
          }
```

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -164,40 +151,48 @@ object Producer {
...
-                             }
-                           ): @nowarn("msg=discarded non-Unit value")
-
-                           ()
-                         }
-                     }
-                   )
-                 }
-               }
+          serializedRecords <- ZIO.foreach(records)(serialize(_, keySerializer, valueSerializer))
```
guizmaii, line 154: I'm wondering if it'd be possible to b lazier here and to serialize each event independently only just before sending it, instead of serializing them all and then sending them all 🤔

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -164,40 +151,48 @@ object Producer {
...
+              p.send(
+                rec,
+                new Callback {
+                  def onCompletion(metadata: RecordMetadata, err: Exception): Unit =
+                    Unsafe.unsafe { implicit u =>
+                      (if (err != null) runtime.unsafe.run(done.fail(err)).getOrThrowFiberFailure(): Unit
+                       else {
+                         res(idx) = metadata
+                         if (count.incrementAndGet == serializedRecords.length)
+                           runtime.unsafe.run(done.succeed(Chunk.fromArray(res))).getOrThrowFiberFailure(): Unit
```
guizmaii, line 184:
```suggestion
                         if (count.incrementAndGet == serializedRecords.length) {
                           runtime.unsafe.run(done.succeed(Chunk.fromArray(res))).getOrThrowFiberFailure(): Unit
                         }
```

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
guizmaii: Tested the performance of my proposal and it's actually way slower. See https://github.com/zio/zio-kafka/issues/531#issuecomment-1378274412

The interruptible version doesn't even work (yet)

guizmaii: I made the interruptible version work but it's super slow

guizmaii: > I'm wondering if it'd be possible to be lazier here and to serialize each event independently only just before sending it, instead of serializing them all and then sending them all 🤔

Conclusion: Doesn't seems to be possible without a massive loss of performances. So it's better to keep it this way

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
svroonland: Yes this should definitely be using `attemptBlocking`, as it waits for up to `closeTimeout`

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
svroonland: Sounds like a good improvement

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
svroonland: Always good to experiment :) I do know that `provide` or `provideEnvironment` calls have some overhead, which is why it's recommended to do it at the outer edges of your ZIO workflows.

svroonland:
@guizmaii Looks like I need an approval again after the latest changes. The merge rules for this repo are quite strict it appears 😅

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
guizmaii: > I do know that provide or provideEnvironment calls have some overhead

Yeah they def do have some overhead

guizmaii (approved):

guizmaii:
@svroonland You're good to go :)

PR #582, opened 2023-01-11
[POC] Lazy version that only serialize a Record just before to send it instead of serializing them all and then only to send them all

guizmaii:
(Interruptible version of this POC available here: #583)

The only issue left to fix in this POC is: what do we do if one of the `serialize` calls fails? It's not handled right now in this POC 🤔


guizmaii:
It's actually way slower. See https://github.com/zio/zio-kafka/issues/531#issuecomment-1378274412

PR #602, opened 2023-01-21
Simplify `Diagnostics` code

guizmaii:


svroonland:
I think we should keep the trait, since the abstract class already prescribes some implementation which we don't know will generalize. I used that trait recently here: https://github.com/zio/zio-kafka/pull/591/files#diff-380a3dc268e32f4acb169043f253e408cad37485296566838a4b0176f306a8b6R28

PR #606, opened 2023-01-21
Ease code reading

guizmaii:


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -564,7 +564,13 @@ private[consumer] object Runloop {
   type ByteArrayCommittableRecord = CommittableRecord[Array[Byte], Array[Byte]]
   type ByteArrayConsumerRecord    = ConsumerRecord[Array[Byte], Array[Byte]]
 
-  final case class Request(tp: TopicPartition, cont: Promise[Option[Throwable], Chunk[ByteArrayCommittableRecord]])
+  final case class Request(
+    tp: TopicPartition,
+    private val cont: Promise[Option[Throwable], Chunk[ByteArrayCommittableRecord]]
+  ) {
+    @inline def continue(data: Chunk[ByteArrayCommittableRecord]): UIO[Boolean] = cont.succeed(data)
+    @inline def end(cause: Option[Throwable]): UIO[Boolean]                     = cont.fail(cause)
```
svroonland, line 572: How about a `fail(Throwable) = cont.fail(Some(cause))` and `end = cont.fail(None)`

guizmaii: Good idea! Thanks :)

svroonland (approved):

PR #608, opened 2023-01-21
Prefer `case object Poll`

guizmaii:


svroonland (dismissed):

svroonland:
Scala 3 doesn't like the `final`

PR #620, opened 2023-01-30
Replace key&value tuple with ConsumerRecord in consumeWith

duxet:
Same as #472 PR, but this time making a breaking change instead of adding a new `consumeWith` function.


svroonland:
Thanks! Looks good, let's do this one and close the other PR. Sorry for the long delay.

If you could fix the documentation error, I can approve it.

Since it's a breaking change, I'd like to save it up a bit for a future release with other breaking changes like #217.

svroonland (approved):

PR #643, opened 2023-02-18
Add transactional test

erikvanoosten:
Add test "does not process messages twice for transactional producer, even when rebalancing"

This PR is equivalent to #644 but specific for the branch of #591.


erikvanoosten:
@iravid the review request was automatic. I do not expect a review from you yet. First this PR should be merged in the branch of #591 .

PR #652, opened 2023-02-22
Clean some code

guizmaii:


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -323,9 +323,7 @@ private[consumer] final class Runloop(
   }
 
   private def doPoll(c: ByteArrayKafkaConsumer, requestedPartitions: Set[TopicPartition]) = {
-    val pollTimeout =
-      if (requestedPartitions.nonEmpty) this.pollTimeout.asJava
```
guizmaii, line 327: We don't need to call `.asJava` as the ZIO `Durarion` type is, in fact, an alias of `java.time.Duration`

svroonland: Probably a ZIO 1 remnant when it was its own type.

svroonland (approved):

PR #653, opened 2023-02-22
[POC] More precise error catching in `doPoll`

guizmaii:


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -331,7 +332,7 @@ private[consumer] final class Runloop(
       try
         c.poll(pollTimeout)
       catch {
-        case _: IllegalStateException =>
+        case _: IllegalStateException if isUnsubscribed =>
```
erikvanoosten, line 335: I would guess that you would have the same race condition here. So is `isUnsubscribed` reliable?

guizmaii: Hum 🤔

guizmaii: Should we even try to `c.poll` if we're unsubscribed? 🤔

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -36,7 +37,7 @@ private[consumer] final class Runloop(
   shutdownRef: Ref[Boolean],
   offsetRetrieval: OffsetRetrieval,
   userRebalanceListener: RebalanceListener,
-  subscribedRef: Ref[Boolean],
+  subscribedRef: AtomicBoolean,
```
svroonland, line 40: I would prefer to stick to ZIO Refs here and see if we can find a good solution using that. Somehow feels safer and more zio-idiomatic.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
svroonland: I just noticed that the call to `handlePoll` in `handleShutdown` is not guarded by the `isSubscribed` check, perhaps that was the cause of the race condition..

svroonland:
Thanks for helping along here. I've been thinking, if we want to avoid changing the subscription in the middle of `handlePoll`, we should introduce a Command so we can handle it sequentially. I'd like to investigate that direction

guizmaii:
> we should introduce a Command so we can handle it sequentially. I'd like to investigate that direction

Interesting idea. Please do :)

Thanks for your amazing work! ❤️

svroonland:
@guizmaii @erikvanoosten I implemented the Command idea here: https://github.com/zio/zio-kafka/pull/517/commits/85378e72c6861c754fd613b7de4ccab64dc500ae

PR #654, opened 2023-02-22
Optimization: Do not deserialize the message if the deserializers are `Serde.byteArray` as these deserializers are not doing anything anyway

guizmaii:
Kafka `ByteArrayDeserializer` implementation:

```java
package org.apache.kafka.common.serialization;

public class ByteArrayDeserializer implements Deserializer<byte[]> {

    @Override
    public byte[] deserialize(String topic, byte[] data) {
        return data;
    }
}
```

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
@@ -6,7 +6,7 @@ import zio._
 import zio.kafka.consumer.diagnostics.Diagnostics
 import zio.kafka.consumer.internal.Runloop.ByteArrayCommittableRecord
 import zio.kafka.consumer.internal.{ ConsumerAccess, Runloop }
-import zio.kafka.serde.Deserializer
+import zio.kafka.serde.{ Deserializer, Serde }
 import zio.stream._
 
 import scala.jdk.CollectionConverters._
@@ -242,6 +242,8 @@ object Consumer {
         }).as(newSubscriptions.map(_.toSet).getOrElse(Set.empty))
       }
 
+      val onlyByteArraySerdes: Boolean = (keyDeserializer eq Serde.byteArray) && (valueDeserializer eq Serde.byteArray)
+
       ZStream.unwrapScoped {
         for {
           stream <- ZStream.fromHubScoped(partitionAssignments)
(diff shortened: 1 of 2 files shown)


svroonland (commented):
Can we benchmark if this makes a difference? Is it worth the additional complexity?

guizmaii:
> Can we benchmark if this makes a difference? Is it worth the additional complexity?

I'd prefer not to write these bench because of a lack of time in general.
I think we can easily see the performance gain:
1. `partitionStream` is a `ZStream[Any, Throwable, ByteArrayCommittableRecord]`
2. `mapChunksZIO` emits chunks of `ByteArrayCommittableRecord` to be processed by the passed function
3. For each record consumed, we'll call `_.deserializeWith(keyDeserializer, valueDeserializer)`
4. For each record consumed, we'll instantiate 2 IOs because of the calls of `keyDeserializer.deserialize` and `valueDeserializer.deserialize` + a `.flatMap` call in between these two calls + a final `.map` for the `.copy` call
5. For each record consumed, we'll instantiate a copy of it to "replace" the initial `key` and `value` by the deserialized `key` and `value`, but here, because the `Serde.byteArray` does nothing, the initial `key` and `value` == the deserialized `key` and `value` so we're instantiating a copy of the record for nothing.

Because we're doing this for every record consumed, it means a lot of extra allocations (IO instances + copy of Record and maybe `Chunk` instances) and extra cycles in the ZIO interpreter that can be bypassed thanks to my change.

I think it's super worth it (we're using this `Serde.byteArray` everywhere where I work because we manage the deserialization ourselves)

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -257,7 +259,11 @@ object Consumer {
                   if (settings.perPartitionChunkPrefetch <= 0) partition
                   else partition.bufferChunks(settings.perPartitionChunkPrefetch)
 
-                tp -> partitionStream.mapChunksZIO(_.mapZIO(_.deserializeWith(keyDeserializer, valueDeserializer)))
+                val stream: ZStream[R, Throwable, CommittableRecord[K, V]] =
+                  if (onlyByteArraySerdes) partitionStream.asInstanceOf[ZStream[R, Throwable, CommittableRecord[K, V]]]
+                  else partitionStream.mapChunksZIO(_.mapZIO(_.deserializeWith(keyDeserializer, valueDeserializer)))
```
guizmaii, line 264: Also, I'm wondering why we're doing `.mapChunksZIO(_.mapZIO(...))` and not just:
```scala
partitionStream.mapZIO(_.deserializeWith(keyDeserializer, valueDeserializer))
```

It feels like we're trying to manually do better than what ZIO does with `ZStream#mapZIO`. 
I'm not convinced it's a good idea

guizmaii: Interesting answer from Adam: https://discord.com/channels/629491597070827530/630498701860929559/1078179897354223706

Let's stay with `mapChunksZIO` 🙂

svroonland:
Fair enough, it likely causes a lot less allocations in that case.

svroonland (approved):

Issue #428, opened 2022-02-09
Throughput issue


svroonland:
Reported by `krille` on the [zio-kafka Discord](https://discordapp.com/channels/629491597070827530/629497941719121960/940169418615431182):

> Hi! Experiencing some weirds issues with zio-kafka message throughput and latency. 
>
>Given the follow settings: 
>- `pollTimeout=20000ms`
>- `fetch.max.wait.ms=10000ms` 
>- `fetch.min.bytes=1` 
>- `max.poll.records=10`
>
>Using only the low level confluent java consumer api everything works as expected. 
>Let's say we have a running consumer on a topic - and no messages in lag. 
>Then pushing 100 records into this topic at once results in: 
>  1. the consumer immediately returns from poll with 10 records 
>  2. processes the records and commits 
>  3. calls poll 
>  4. immediately returns with 10 new records.... 
>and this goes on until there is no more lag. 
>
>With zio-kafka the result is totally different. 
>It always seems to call poll for 20 seconds (never returning immediately even if `fetch.max.wait.ms`, `fetch.min.bytes`, `max.poll.records` has been satisfied). 
>I'm struggling to find the cause for this. 
>The code is based on examples from zio-kafka documentation and zio-kafka test code. 
>
>Can provide more detailed code examples if needed. 
>Is there some important considerations to take between setting `pollTimeout` and `pollInterval` (in zio-kafka)? Has anyone ever experienced any issues like this?
>
> More investigation reveals that this weird behaviour is a result of committing the offsets. 
>`offsets.commitOrRetry(Schedule.spaced(100.milliseconds) && Schedule.recurs(10))` seems to blocks for the same amount of time as the pollTimeout.

```scala
ConsumerSettings(
  List(localhost:9092),
  HashMap(
    heartbeat.interval.ms -> 3000, 
    security.protocol -> PLAINTEXT, 
    group.id -> test_blurp_1, 
    request.timeout.ms -> 30000, 
    session.timeout.ms -> 10000, 
    fetch.max.wait.ms -> 10000, 
    fetch.min.bytes -> 1, 
    max.poll.records -> 10, 
    max.poll.interval.ms -> 300000
  ),
[...]

PR #661, opened 2023-02-24
WIP

guizmaii:
The fun fact is that I noticed this issue because it's something we fixed on our app code several months ago :)
But I'm not sure this applies to zio-kafka code, not knowing it very well yet (even if I really start to understand how the `RunLoop` works).
That's why, I still have to write the test but, I'd like your review first @svroonland

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -659,23 +659,23 @@ private[consumer] object Runloop {
       subscribedRef   <- Ref.make(false)
       runtime         <- ZIO.runtime[Any]
       runloop = new Runloop(
-                  runtime,
-                  hasGroupId,
-                  consumer,
-                  pollFrequency,
-                  pollTimeout,
-                  requestQueue,
-                  commitQueue,
-                  lastRebalanceEvent,
-                  partitions,
-                  rebalancingRef,
-                  diagnostics,
-                  shutdownRef,
-                  offsetRetrieval,
-                  userRebalanceListener,
-                  subscribedRef,
-                  restartStreamsOnRebalancing,
-                  currentStateRef
+                  runtime = runtime,
+                  hasGroupId = hasGroupId,
+                  consumer = consumer,
+                  pollFrequency = pollFrequency,
+                  pollTimeout = pollTimeout,
+                  requestQueue = requestQueue,
+                  commitQueue = commitQueue,
+                  lastRebalanceEvent = lastRebalanceEvent,
+                  partitions = partitions,
+                  rebalancingRef = rebalancingRef,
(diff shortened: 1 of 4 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -548,7 +548,7 @@ private[consumer] final class Runloop(
   def run: ZIO[Scope, Nothing, Fiber.Runtime[Throwable, Unit]] =
     ZStream
       .mergeAll(3, 1)(
-        ZStream(Command.Poll).repeat(Schedule.spaced(pollFrequency)),
+        ZStream(Command.Poll).repeat(Schedule.once ++ Schedule.spaced(pollFrequency)),
```
svroonland, line 551: I had to look twice, but I believe `ZStream(x).repeat(schedule)` immediately emits `x` and afterwards according to the `schedule`, which is what we want isn't it?

guizmaii: I'll write a test to be validate the behaviour of `spaced`

guizmaii: FYI @svroonland, I was wrong. The code was indeed already emitting a first `Poll` immediately (my tests are proving this)

But, as mentioned here https://github.com/zio/zio-kafka/issues/428#issuecomment-1445352019, I think we should change this `ZStream(Command.Poll).repeat(Schedule.spaced(pollFrequency))` piece of code by `ZStream.repeatWithSchedule(Command.Poll, Schedule.fixed(pollFrequency))` for 2 reasons:

1. I think that `ZStream.repeatWithSchedule(Command.Poll, ...)` is easier to reason about compared to ` ZStream(Command.Poll).repeat(...)`

In the `ZStream(Command.Poll).repeat(...)` version, it's IMO hard to understand if `ZStream(Command.Poll)` (without the `.repeat(...)` part is emitting this first `Poll` event. And if it does so, if the `.repeat(...)` part triggers an immediate event too? Hard to tell and understand IMO
While the `ZStream.repeatWithSchedule(Command.Poll, ...)` is IMO more explicit. It's trigger this `Command.Poll` according to the passed `Schedule`. No ambiguity.

2. I think that with `Schedule.spaced` we'll always waiting `pollInterval` even if this `pollInterval` has been spend doing something else. With `Schedule.fixed`, if this `pollInterval` amount of time has been spent doing something else, the `Poll` will be triggered immediatly

svroonland: Agreed, `ZStream.repeatWithSchedule(Command.Poll, Schedule.fixed(pollInterval)` is better for now, until we implement the suggestions in #664.

PR #662, opened 2023-02-24
[POC] Move the `ConsumerRebalanceListener` creation to the `RunLoop`

guizmaii:


svroonland (approved):
This makes more sense

PR #668, opened 2023-02-27
Avoid extra latency potentially happening between `Poll` calls

guizmaii:
Related to work done here: 
- https://github.com/zio/zio-kafka/pull/661

I didn't manage to find a way to test that this "extra latency" is not stacked anymore 🤔

@svroonland I wonder if there's not a risk of generating/stacking a lot of `Poll` in the Stream 🤔
I wonder if using a `Queue.dropping[Command.Poll](1)` wouldn't be a good idea.
At any time, I think that we only need to have 1 or 0 Poll command: if we have one, at the next iteration of the loop, it'll consume it from the queue and will execute the corresponding action. If 0, it'll do nothing.
If the queue was containing 1 Poll command and we're emitting a new one, we're replacing the old one with the new one, so we don't waste memory and we have no risks of stacking too many of them (which could potentially lead to an OOM)

WDYT? 🤔

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/diagnostics/DiagnosticEvent.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/diagnostics/DiagnosticEvent.scala
@@ -2,6 +2,7 @@ package zio.kafka.consumer.diagnostics
 
 import org.apache.kafka.clients.consumer.OffsetAndMetadata
 import org.apache.kafka.common.TopicPartition
+import zio.kafka.consumer.internal.Runloop.Command
 
 sealed trait DiagnosticEvent
 object DiagnosticEvent {
@@ -25,4 +26,6 @@ object DiagnosticEvent {
     final case class Assigned(partitions: Set[TopicPartition]) extends Rebalance
     final case class Lost(partitions: Set[TopicPartition])     extends Rebalance
   }
+
+  final case class RunloopEvent(command: Command) extends DiagnosticEvent
 }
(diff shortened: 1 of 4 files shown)


svroonland:
> I wonder if there's not a risk of generating/stacking a lot of Poll in the Stream..
Yes, the same thought had crossed my mind this weekend.

zio-kafka/src/main/scala/zio/kafka/consumer/diagnostics/DiagnosticEvent.scala
```diff
@@ -25,4 +26,6 @@ object DiagnosticEvent {
     final case class Assigned(partitions: Set[TopicPartition]) extends Rebalance
     final case class Lost(partitions: Set[TopicPartition])     extends Rebalance
   }
+
+  final case class RunloopEvent(command: Command) extends DiagnosticEvent
```
svroonland, line 30: Not a huge fan of this, it means exposing one of the Runloop's internals. Shall we at least document that it's interface is unstable, so we can change the Commands a bit more freely if we need?

guizmaii: Agreed. TBH, I'm not clear what this `Diagnostics` thing is for

svroonland (dismissed):

svroonland (approved):

erikvanoosten:
> > I wonder if there's not a risk of generating/stacking a lot of Poll in the Stream..
> 
> Yes, the same thought had crossed my mind this weekend.

I assumed it already worked this way, but now I see it doesn't :)
Has this already been addressed? I can't find this change (if it was done already).

svroonland:
Not yet, I'm experimenting a bit with some changed polling behavior though.

PR #683, opened 2023-03-02
Improve polling behavior

svroonland:
Previously, polling was driven by two triggers: a regular intervals and pending requests from one or more partition streams. This is inefficient for several reasons:
* When not all requests were fulfilled, the runloop would wait until the next scheduled poll to poll again, introducing unneeded latency.
* In between the request-driven polling, there would often be scheduled polls with no pending requests and therefore only paused partitions and no data being fetched. Incoming requests would have to wait for this poll to be complete, introducing extra latency.
* When there were multiple pending requests in the queue, not all of them would be dequeued before initiating a poll. This also lead to more partitions being paused than necessary.
* Commits would often have to wait an extra poll cycle before being dequeued. See #428

The new implementation dequeues all currently enqueued requests and commits and only then performs a poll. If there are no pending requests or commits, we wait until a command can be dequeued. Before we are have created our partition streams (after ininialization or rebalancing), we poll continuously.

Benchmarks show a significant improvement in throughput.


zio-kafka-test-utils/src/main/scala/zio/kafka/KafkaTestUtils.scala
```diff
@@ -80,8 +80,9 @@ object KafkaTestUtils {
           ConsumerConfig.AUTO_OFFSET_RESET_CONFIG        -> "earliest",
           ConsumerConfig.METADATA_MAX_AGE_CONFIG         -> "100",
           ConsumerConfig.SESSION_TIMEOUT_MS_CONFIG       -> "3000",
-          ConsumerConfig.HEARTBEAT_INTERVAL_MS_CONFIG    -> "250",
-          ConsumerConfig.MAX_POLL_RECORDS_CONFIG         -> "10",
```
erikvanoosten, line 84: I suspect that many tests need a low `MAX_POLL_RECORDS_CONFIG` so that pathetic behavior is detected quicker. This is why the transactional spec sets it specifically.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -721,16 +726,16 @@ private[consumer] object Runloop {
   ): ZIO[Scope, Throwable, Runloop] =
     for {
       rebalancingRef     <- Ref.make(false)
-      requestQueue       <- ZIO.acquireRelease(Queue.unbounded[Runloop.Request])(_.shutdown)
-      commandQueue       <- ZIO.acquireRelease(Queue.unbounded[Command])(_.shutdown)
+      requestQueue       <- ZIO.acquireRelease(Queue.bounded[Runloop.Request](1024))(_.shutdown)
+      commandQueue       <- ZIO.acquireRelease(Queue.bounded[Command](1024))(_.shutdown)
```
erikvanoosten, line 727: Someone is going to ask why 1024 is the right number. Better be prepared :)

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -611,21 +602,33 @@ private[consumer] final class Runloop(
...
-      .tapErrorCause(cause => ZIO.logErrorCause("Error in Runloop", cause))
-      .onError(cause => partitions.offer(Take.failCause(cause)))
-      .unit
-      .forkScoped
+  def run: ZIO[Scope, Nothing, Fiber.Runtime[Throwable, Any]] =
+    Promise.make[Nothing, Unit].flatMap { initialized =>
+      (ZStream
+        .mergeAll(3, 1)(
+          // Poll commands repeatedly until we're getting requests to drive the polling
+          ZStream.repeat(Command.Poll).interruptWhen(initialized),
```
erikvanoosten, line 610: Rather than complicating this method, can we extract creating Poll commands? My hope is that it will make it easier to start polling e.g. after a rebalance.

erikvanoosten: How about we introduce a boolean in the state that indicates that we are awaiting a data request? As long as it it true, we need to schedule another poll.

svroonland:
It would be good to have a throughput benchmark and compare before and after this PR.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -611,48 +590,61 @@ private[consumer] final class Runloop(
...
+
+  private def pollLoop(initialized: Boolean, state: State): ZIO[Any, Throwable, State] = {
+    def processCommands(state: State) = for {
+      // Gather available commands or return immediately if nothing in the queue
+      commands   <- commandQueue.takeAll
+      isShutdown <- isShutdown
+      updatedState <- ZIO.foldLeft(commands)(state) { case (s, cmd) =>
+                        (if (isShutdown) handleShutdown(s, cmd) else handleOperational(s, cmd)) <*
+                          diagnostics.emitIfEnabled(DiagnosticEvent.RunloopEvent(cmd))
+                      }
```
erikvanoosten, line 607: `isShutdown` is not going change while looping over the commands.

```suggestion
      isShutdown <- isShutdown
      updatedState <- if (isShutdown) {
        ZIO.foldLeft(commands)(state) { case (s, cmd) => handleShutdown(s, cmd)) <* diagnostics.emitIfEnabled(DiagnosticEvent.RunloopEvent(cmd)) }
      } else {
        ZIO.foldLeft(commands)(state) { case (s, cmd) => handleOperational(s, cmd)) <* diagnostics.emitIfEnabled(DiagnosticEvent.RunloopEvent(cmd)) }
      }
```

Can be further improved by extracting `diagnostics.emitIfEnabled(DiagnosticEvent.RunloopEvent(cmd))` to a helper method:

Something like:

```suggestion
  def emitCmdDiagnostics(cmd: Command) = diagnostics.emitIfEnabled(DiagnosticEvent.RunloopEvent(cmd))

      isShutdown <- isShutdown
      updatedState <- if (isShutdown) {
        ZIO.foldLeft(commands)(state) { case (s, cmd) => handleShutdown(s, cmd)) <* emitCmdDiagnostics(cmd) }
      } else {
        ZIO.foldLeft(commands)(state) { case (s, cmd) => handleOperational(s, cmd)) <* emitCmdDiagnostics(cmd) }
      }
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -611,48 +590,61 @@ private[consumer] final class Runloop(
...
+    } yield updatedState
+
+    def doPollIfPendingActions(state: State) =
+      for {
+        isRebalancing <- isRebalancing
+        isInitialized = initialized || state.pendingRequests.nonEmpty
+        shouldPoll =
+          state.isSubscribed && (state.pendingRequests.nonEmpty || state.pendingCommits.nonEmpty || !isInitialized || isRebalancing)
```
erikvanoosten, line 615: `state.pendingRequests.nonEmpty` is used twice, once in a positive way and once in a negative way. In other words: `shouldPoll` is equal to `state.isSubscribed` :)

```suggestion
        shouldPoll =
          state.isSubscribed && (state.pendingRequests.nonEmpty || state.pendingCommits.nonEmpty || !initialized || isRebalancing)
```

I did not look up what `initialized` does yet. However, from the name I would guess we need:
```suggestion
        shouldPoll =
          state.isSubscribed && initialized && (state.pendingRequests.nonEmpty || state.pendingCommits.nonEmpty || isRebalancing)
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -611,48 +590,61 @@ private[consumer] final class Runloop(
...
+        _ <-
+          ZIO
+            .logTrace(
+              s"Starting poll with ${state.pendingRequests.size} pending requests and ${state.pendingCommits.size} pending commits"
+            )
+            .when(shouldPoll)
+        newState <- if (shouldPoll) handlePoll(state) else ZIO.succeed(state)
```
erikvanoosten, line 622: ZIO.succeed should be replaced with something like ZIO.sleep, otherwise we'll get spinning wait.

svroonland: I thought about that, but if there's queued events coming in in that sleep period, we process them with latency. We could try an optional timeout for the next dequeueing.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
svroonland: Not sure if this makes things more readable..

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -611,48 +590,66 @@ private[consumer] final class Runloop(
...
+  def run: ZIO[Scope, Nothing, Fiber.Runtime[Throwable, Any]] = {
+    def processCommands(state: State, wait: Boolean): Task[State] = for {
+      commands <- if (wait)
+                    commandQueue.takeBetween(1, commandQueueSize).timeoutTo(Chunk.empty)(identity)(pollFrequency)
+                  else commandQueue.takeAll // Gather available commands or return immediately if nothing in the queue
+
+      isShutdown <- isShutdown
+      updatedState <- ZIO.foldLeft(commands)(state) { case (s, cmd) =>
+                        (if (isShutdown) handleShutdown(s, cmd) else handleOperational(s, cmd)) <*
+                          diagnostics.emitIfEnabled(DiagnosticEvent.RunloopEvent(cmd))
```
guizmaii, line 602: What do you guys think about this change:
```suggestion
        isShutdown <- isShutdown
        commands <- if (!wait || isShutdown)
                      commandQueue.takeAll // Gather available commands or return immediately if nothing in the queue
                    else commandQueue.takeBetween(1, commandQueueSize).timeoutTo(Chunk.empty)(identity)(pollFrequency)

        updatedState <- ZIO.foldLeft(commands)(state) { case (s, cmd) =>
                          (if (isShutdown) handleShutdown(s, cmd) else handleOperational(s, cmd)) <*
                            diagnostics.emitIfEnabled(DiagnosticEvent.RunloopEvent(cmd))
                        }
```

So that if we're shutting down, we take all the commands and end them.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -611,48 +590,66 @@ private[consumer] final class Runloop(
...
+          state.isSubscribed && (state.pendingRequests.nonEmpty || state.pendingCommits.nonEmpty || !isInitialized || isRebalancing)
+        _ <-
+          ZIO
+            .logTrace(
+              s"Starting poll with ${state.pendingRequests.size} pending requests and ${state.pendingCommits.size} pending commits"
+            )
+            .when(shouldPoll)
+        newState <-
+          if (shouldPoll) handlePoll(state) else ZIO.succeed(state)
+      } yield (newState.copy(initialized = isInitialized), !shouldPoll)
```
guizmaii, line 620: I want to suggest a simplification for the `initialized` state management:
```scala
    def doPollIfPendingActions(state: State): Task[(State, Boolean)] =
      if (!state.initialized) handlePoll(state).map(state => (state.copy(initialized = true), false))
      else {
        for {
          isRebalancing <- isRebalancing
          shouldPoll =
            state.isSubscribed && (state.pendingRequests.nonEmpty || state.pendingCommits.nonEmpty || isRebalancing)
          _ <-
            ZIO
              .logTrace(
                s"Starting poll with ${state.pendingRequests.size} pending requests and ${state.pendingCommits.size} pending commits"
              )
              .when(shouldPoll)
          newState <- if (shouldPoll) handlePoll(state) else ZIO.succeed(state)
        } yield (newState, !shouldPoll)
      }
```

I think this code is more explicit. We only care once about the `initialized` state, no? 🤔

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -720,25 +724,23 @@ private[consumer] object Runloop {
   ): ZIO[Scope, Throwable, Runloop] =
     for {
       rebalancingRef     <- Ref.make(false)
-      requestQueue       <- ZIO.acquireRelease(Queue.unbounded[Runloop.Request])(_.shutdown)
-      commandQueue       <- ZIO.acquireRelease(Queue.unbounded[Command])(_.shutdown)
+      commandQueue       <- ZIO.acquireRelease(Queue.bounded[Runloop.Command](commandQueueSize))(_.shutdown)
```
guizmaii, line 727: Why not `unbounded`? 
My issue with a bounded queue is that if something goes wrong, the queue will drop/lose commands, giving us no chance to detect this issue. With an unbounded one, if something goes wrong, the program will most likely OOM (or at least consume a lot of RAM), which can be detected by our users and reported to us.

guizmaii, line 727: Why not `unbounded`? 
My issue with a bounded queue is that if something goes wrong, the queue will make the program stop consuming the commands, the program will not do anything anymore. It's hard to detect for a zio-kafka user.
With an unbounded queue, if something goes wrong, the program will most likely OOM (or at least consume a lot of RAM), which can be more easily detected by our users and reported to us.

svroonland: Interesting perspective. I'm not usually a fan of unbounded memory usage in parts of a program. OOMs like these also tend to develop very slowly, over the course of days, which makes them very hard to track. I've seen this with zio-kafka due to some issues in zio core. 

Also, even with an unbounded queue, the run loop could (theoretically) stall for some reason. 

It might be better to detect stalling another way, with a timeout in some places.

svroonland: I implemented a simple timeout mechanism, wdyt?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
svroonland: This shutdown is really a 'gracefully stop fetching new data', so there may still be commits being made. We need to process those commit commands, which may come in over several iterations of the poll loop.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
guizmaii: Well in this case we could still to this:

```scala
def doPollIfPendingActions(state: State): Task[(State, Boolean)] =
  if (!state.initialized) handlePoll(state).map(state => (state.copy(initialized = state.pendingRequests.nonEmpty), false))
  else {
    for {
      isRebalancing <- isRebalancing
      shouldPoll =
        state.isSubscribed && (state.pendingRequests.nonEmpty || state.pendingCommits.nonEmpty || isRebalancing)
      _ <-
        ZIO
          .logTrace(
            s"Starting poll with ${state.pendingRequests.size} pending requests and ${state.pendingCommits.size} pending commits"
          )
          .when(shouldPoll)
        newState <- if (shouldPoll) handlePoll(state) else ZIO.succeed(state)
      } yield (newState, !shouldPoll)
```

No? 🤔

svroonland: I think that's equivalent but in my opinion not easier to read, since there's now two places where `handlePoll` is called.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Solved; waiting for the command queue (as in later commits) is much better than both doing nothing and sleeping.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -98,34 +99,36 @@ private[consumer] final class Runloop(
...
+                case None =>
+                  ZIO.some(Runloop.RebalanceEvent.Revoked(result))
+                case _ =>
+                  ZIO.fail(
+                    new IllegalStateException(
+                      s"onRevoked called on rebalance listener with pending assigned event"
+                    )
                   )
-                )
+              }.unlessZIO(isShutdown).unit
```
erikvanoosten, line 129: onLost needs to be handled differently. When we lose the partition the stream should be interrupted as it doesn't make sense to process the drain queue any more.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -199,7 +202,7 @@ private[consumer] final class Runloop(
    *   New pending requests, new buffered records and active assigned streams
    */
   private def endRevoked(
```
erikvanoosten, line 204:
```suggestion
   * This method is invoked from the rebalance listener.
   */
  private def endRevoked(
```

erikvanoosten: Hmm, that is not correct. It is also invoked in other situations.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -98,34 +99,36 @@ private[consumer] final class Runloop(
...
-            _ => true
-          ).flatMap { result =>
-            lastRebalanceEvent.updateZIO {
-              case None =>
-                ZIO.some(Runloop.RebalanceEvent.Revoked(result))
-              case _ =>
-                ZIO.fail(
-                  new IllegalStateException(
-                    s"onRevoked called on rebalance listener with pending assigned event"
+        ZIO.logDebug("Rebalancing started") *>
```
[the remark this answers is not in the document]
erikvanoosten: Ow I see. This is for restartStreamsOnRebalancing=true...

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Anyway, that is not needed for this PR.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -211,21 +214,28 @@ private[consumer] final class Runloop(
...
     val newBufferedRecords  = BufferedRecords.fromMutableMap(buf)
 
-    revokeAction.as(Runloop.RevokeResult(unfulfilledRequests, newBufferedRecords, assignedStreams))
+    val endRevokedRequests = ZIO.foreachDiscard(requests.filter(req => isRevoked(req.tp))) { req =>
+      ZIO.logTrace(s"Ending request for TP ${req.tp}") *> req.end.unit
+    }
+
+    endRevokedRequests *> revokeAction.as(
+      Runloop.RevokeResult(unfulfilledRequests, newBufferedRecords, assignedStreams)
+    )
```
erikvanoosten, line 238: Not, sure if this is better:
```suggestion
    for {
      _ <- endRevokedRequests
      _ <- revokeAction
    } yield Runloop.RevokeResult(unfulfilledRequests, newBufferedRecords, assignedStreams)
```

erikvanoosten: Or:
```suggestion
    endRevokedRequests *>
    revokeAction *>
    ZIO.succeed(
      Runloop.RevokeResult(unfulfilledRequests, newBufferedRecords, assignedStreams)
    )
```

(I don't like the `as` method as it hinders readability.)

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -235,11 +245,11 @@ private[consumer] final class Runloop(
    *   Remaining pending requests and remaining/new buffered records
    */
   private def fulfillRequests(
-    pendingRequests: Chunk[Runloop.Request],
+    pendingRequests: Chunk[Request],
     bufferedRecords: BufferedRecords,
     records: ConsumerRecords[Array[Byte], Array[Byte]]
```
erikvanoosten, line 250:
```suggestion
    polledRecords: ConsumerRecords[Array[Byte], Array[Byte]]
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -254,7 +264,9 @@ private[consumer] final class Runloop(
       } else {
         val concatenatedChunk = bufferedChunk ++ Chunk.fromJavaIterable(reqRecs)
```
erikvanoosten, line 265:
```suggestion
        val concatenatedChunk = bufferedChunk ++ Chunk.fromJavaIterable(reqRecs)
        // Extend fulfillAction
        fulfillAction = for {
          _ <- fulfillAction
          _ <- ZIO
            .logTrace(s"Fulfilling ${bufferedChunk.size} buffered records")
            .when(bufferedChunk.nonEmpty)
          succeeded <- req.succeed(
            concatenatedChunk.map { record =>
              CommittableRecord(
                record = record,
                commitHandle = commit,
                consumerGroupMetadata = getConsumerGroupMetadataIfAny
              )
            }
          )
        } yield succeeded
        buf -= req.tp
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -461,48 +478,22 @@ private[consumer] final class Runloop(
...
-        }
-        .orElseSucceed(state.addRequests(reqs))
-    )
-
-  private def handleCommit(state: State, cmd: Command.Commit): UIO[State] =
-    ZIO.ifZIO(isRebalancing)(
-      onTrue = ZIO.succeed(state.addCommit(cmd)),
-      onFalse = doCommit(Chunk(cmd)).as(state)
+      newPendingCommits <- ZIO.filter(state.pendingCommits)(_.isPending)
+    } yield state.copy(
```
erikvanoosten, line 492: Using copy instead of building a new State from scratch is tricky. What if State gets a new field? Most probably forgetting to set the new field here will result in a bug...

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: I am not sure I fully understand the logic yet, or perhaps I have a different expectation. I thought that what is called 'initialized' here is not just after startup, but after _every_  change in the list of partitions. Take a rebalance for example. After the rebalance is finished, we might have a new partition for which we did not get a request yet. In the current code we would not poll. Is that correct? Now that I am writing this, I realize that the answer might be: yes :)

Here is a weird case:
Suppose there are 2 partitions, handled by streams A and B. Stream A creates a request so that 'initialized' goes to true, afterwards stream A halts processing due to a bug. Polling now stops while stream B did not send a request yet. How would this work out?

BTW, I have no love for the word `initialized` here. It is not very clear _what_ is initialized. IMHO we should emphasize that we need additional polling to properly initialize the stream (did I say that correctly?). What do you think of `initialPolling` or `extraPollingNeeded`, or ....

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: I'd like to see that. Where is the timeout mechanism?

erikvanoosten: I found it. Looks nice!

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
svroonland: Yeah that's a missing piece of functionality still. Perhaps we can reset `initialized` after revoking partitions. Have to examine what the exact conditions would have to be.

Agreed on the `initialized` name

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Feel free to ignore this...

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
svroonland: > Suppose there are 2 partitions, handled by streams A and B. Stream A creates a request so that 'initialized' goes to true, afterwards stream A halts processing due to a bug. Polling now stops while stream B did not send a request yet. How would this work out?

When stream B eventually sends a request (within the runloop timeout, 30 seconds default), polling will continue, #686 aside.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
svroonland: Obsolete in latest version

svroonland:
@guizmaii @erikvanoosten PR is updated with most of your review comments. If you could approve, we can merge it.

erikvanoosten (approved):
LGTM. BTW I am particularly happy that the isRebalancing ref is gone because I discovered it is not reliable.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
svroonland: My mistake, I was looking at another copy I think. I applied your suggestion

.github/workflows/benchs.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - improve-poll-behavior
```
guizmaii, line 7: To remove?

svroonland: Yes

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -5,6 +5,24 @@ import zio._
...
+ * @param pollTimeout
+ * @param perPartitionChunkPrefetch
+ * @param offsetRetrieval
+ * @param rebalanceListener
+ * @param restartStreamOnRebalancing
+ * @param runloopTimeout
+ *   Internal timeout for each iteration of the command processing and polling loop, use to detect stalling. This should
+ *   be much larger than the pollTimeout and the time it takes to process chunks of records. If your consumer is not
+ *   subscribed for long periods during its lifetime, this timeout should take that into account as well. When the
+ *   timeout expires, the plainStream/partitionedStream/etc will fail with a [[Consumer.RunloopTimeout]].
```
guizmaii, line 24: 🥇 Great doc! Thanks :)

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -27,40 +23,43 @@ private[consumer] final class Runloop(
...
-                   result  <- request.await
-                 } yield result
-               }.interruptWhen(interruptionPromise)
-                 .concat(
-                   ZStream
-                     .fromQueue(drainQueue)
-                     .flattenTake
-                 )
+      stream = ZStream.logAnnotate("topic", tp.topic()) *>
+                 ZStream.logAnnotate("partition", tp.partition().toString) *>
```
guizmaii, line 47: 🟢
```suggestion
      stream =
        ZStream.logAnnotate(LogAnnotation("topic", tp.topic()), LogAnnotation("partition", tp.partition().toString)) *>
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -27,40 +23,43 @@ private[consumer] final class Runloop(
...
+                 ZStream.finalizer(
+                   ZIO.logDebug(s"Partition stream for ${tp.toString} has ended")
+                 ) *> ZStream.repeatZIOChunkOption {
```
guizmaii, line 50: 🟢
```suggestion
      stream =
        ZStream.logAnnotate(LogAnnotation("topic", tp.topic()), LogAnnotation("partition", tp.partition().toString)) *>
          ZStream.finalizer(ZIO.logDebug(s"Partition stream for ${tp.toString} has ended")) *>
          ZStream.repeatZIOChunkOption {
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -235,26 +237,28 @@ private[consumer] final class Runloop(
...
 
       if (bufferedChunk.isEmpty && reqRecs.isEmpty) {
         acc += req
       } else {
         val concatenatedChunk = bufferedChunk ++ Chunk.fromJavaIterable(reqRecs)
 
-        fulfillAction = fulfillAction *> req.succeed(concatenatedChunk.map { record =>
+        fulfillAction = fulfillAction <* ZIO
+          .logTrace(s"Fulfilling ${bufferedChunk.size} buffered records")
+          .when(bufferedChunk.nonEmpty) *> req.succeed(concatenatedChunk.map { record =>
```
guizmaii, line 261: 🌕 Hum, because we're in the `else` clause of `if (bufferedChunk.isEmpty && reqRecs.isEmpty)`, do we need `.when(bufferedChunk.nonEmpty)`? 🤔

It can't be empty, no? 🤔

guizmaii, line 261: 🟢 Why are you logging `bufferedChunk.size` here and not `concatenatedChunk.size`? 🤔

guizmaii, line 261: 🟢 In an effort to reduce to the minimum the amount of `flatMap`, I'd like to suggest this change:
```suggestion
        fulfillAction =
          if (bufferedChunk.nonEmpty)
            ZIO.logTrace(s"Fulfilling ${bufferedChunk.size} buffered records") *> req.succeed(concatenatedChunk)
          else req.succeed(concatenatedChunk)
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -513,44 +497,29 @@ private[consumer] final class Runloop(
...
-      case Command.Requests(reqs) =>
-        ZIO.foreachDiscard(reqs)(_.end).as(state)
-      case cmd @ Command.Commit(_, _) =>
-        handleCommit(state, cmd)
+      case req: Request =>
+        req.end.as(state)
       case r @ Command.ChangeSubscription(_, _, _) =>
         r.succeed.as(state)
+      case _ @Command.Commit(_, cont) =>
+        cont.fail(new KafkaException("Consumer is shutting down")).as(state) // TODO can we just allow it?
```
guizmaii, line 505: 🌕 Do we want to keep this TODO?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -513,44 +497,29 @@ private[consumer] final class Runloop(
...
-          } else ZIO.succeed(state)
+      case req: Request =>
+        if (state.isSubscribed) {
+          ZIO.succeed(state.addRequest(req))
+        } else {
+          req.end.as(state)
         }
       case cmd @ Command.Commit(_, _) =>
-        handleCommit(state, cmd)
+        doCommit(Chunk(cmd)).as(state.addCommit(cmd))
```
guizmaii, line 518:
```suggestion
        doCommit(Chunk.single(cmd)).as(state.addCommit(cmd))
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -611,48 +580,67 @@ private[consumer] final class Runloop(
...
-        ZStream.fromQueue(commandQueue)
-      )
-      .tap(cmd => diagnostics.emitIfEnabled(DiagnosticEvent.RunloopEvent(cmd)))
-      .runFoldZIO(State.initial) { (state, cmd) =>
-        ZIO.ifZIO(isShutdown)(onTrue = handleShutdown(state, cmd), onFalse = handleOperational(state, cmd))
-      }
+  def run: ZIO[Scope, Nothing, Fiber.Runtime[Throwable, Any]] = {
+    def processCommands(state: State, wait: Boolean): Task[State] = for {
+      commands <- if (wait)
+                    commandQueue.takeBetween(1, commandQueueSize).timeoutTo(Chunk.empty)(identity)(pollFrequency)
```
guizmaii, line 586:
```suggestion
                    commandQueue.takeBetween(1, commandQueueSize).timeoutTo(Chunk.empty)(ZIO.identityFn)(pollFrequency)
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -611,48 +580,67 @@ private[consumer] final class Runloop(
...
+          if (shouldPoll) handlePoll(state) else ZIO.succeed(state)
+      } yield (newState, !shouldPoll)
+    }
+
+    def loop(state: State, wait: Boolean): ZIO[Any, Throwable, Nothing] = processCommands(state, wait)
+      .flatMap(doPollIfPendingActions)
+      .timeoutFail(RunloopTimeout)(runloopTimeout)
+      .flatMap { case (state, wait) => loop(state, wait) }
+
+    loop(State.initial, true)
```
guizmaii, line 617:
```suggestion
    loop(State.initial, wait = true)
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -611,48 +580,67 @@ private[consumer] final class Runloop(
...
+      for {
+        _ <-
+          ZIO
+            .logTrace(
+              s"Starting poll with ${state.pendingRequests.size} pending requests and ${state.pendingCommits.size} pending commits"
+            )
+            .when(shouldPoll)
+        newState <-
+          if (shouldPoll) handlePoll(state) else ZIO.succeed(state)
+      } yield (newState, !shouldPoll)
```
guizmaii, line 609: What do you think of:
```suggestion
    def doPollIfPendingActions(state: State): Task[(State, Boolean)] = {
      def logPollStart: UIO[Unit] =
        ZIO
          .logTrace(
            s"Starting poll with ${state.pendingRequests.size} pending requests and ${state.pendingCommits.size} pending commits"
          )

      val shouldPoll =
        state.isSubscribed && (state.pendingRequests.nonEmpty || state.pendingCommits.nonEmpty || state.assignedStreams.isEmpty)
      val wait = !shouldPoll
      if (shouldPoll) logPollStart *> handlePoll(state).map(_ -> wait) else ZIO.succeed(state -> wait)
    }
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -756,21 +752,18 @@ private[consumer] object Runloop {
...
+  pendingRequests: Chunk[Request],
+  pendingCommits: Chunk[Commit],
   bufferedRecords: BufferedRecords,
   assignedStreams: Map[TopicPartition, PartitionStreamControl],
   subscription: Option[Subscription]
 ) {
-  def addCommit(c: Command.Commit): State           = copy(pendingCommits = c +: pendingCommits)
-  def addRequest(c: Runloop.Request): State         = copy(pendingRequests = c +: pendingRequests)
-  def addRequests(c: Chunk[Runloop.Request]): State = copy(pendingRequests = c ++ pendingRequests)
+  def addCommit(c: Commit): State   = copy(pendingCommits = pendingCommits :+ c)
```
guizmaii, line 761: 🟢 This change is interesting. We were prepending the commits, and now we're appending them to the list commits.

1. Why were we prepending them? That seems counter-intuitive 🤔

2. Same question for the `addRequests`, why are we prepending requests instead of appending them? 🤔
IIUC, it means that the `fulfillRequests` loop will take care of the most recent requests first and the oldest last? 🤔

guizmaii (approved):
See suggestions and questions, please :)

Amazing work! As discussed on Discord, the benchs are showing an incredible improvement of the consumer performances! 👏 🔥 💪 🥇

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
svroonland: Ah I missed that todo.

In fact, I now know the answer: we have to process those commits. stopConsumption stops fetching new records, but the Runloop should keep polling and process commits for all messages that are still being processed downstream.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
svroonland: Good one. I traced it back to this PR: https://github.com/zio/zio-kafka/pull/298/files#diff-4bb68de9081869e01fdb038a48581bb121471b3de1ddad2cf0e834e4e3e356fdL456

It used to be `List`, where prepending makes sense, but with `Chunk` I don't see why we should be prepending. There's no bugs resulting from that though, since it doesn't matter in which order pending requests are processed, as far as I know.

I changed it to appending.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
svroonland: I think I like the f.when(..) style. This is executed only for each request, not for each record or something, so I doubt it matters performance-wise. ZIO's internals are quite optimized for synchronous cases like these.

svroonland:
Thanks! Processed the last suggestions.

Here goes.

Issue #664, opened 2023-02-25
repoll immediately when any request is unfulfilled


tnielens:
In `RunLoop.run`
```scala
  def run: ZIO[Scope, Nothing, Fiber.Runtime[Throwable, Unit]] =
    ZStream
      .mergeAll(3, 1)(
        ZStream(Command.Poll).repeat(Schedule.spaced(pollFrequency)),
        ZStream.fromQueue(requestQueue).mapChunks(c => Chunk.single(Command.Requests(c))),
...
```
the `RunLoop` takes a chunk of partition Requests from the `RunLoop.requestQueue` and runs an iteration of the loop in `handleOperational`.  The Requests that aren't fulfilled remain in the `RunLoop.State.pendingRequests`. But a loop iteration will only run if other Requests are present  on the `RunLoop.requestQueue` or upon the next `Command.Poll` occurrence. That can potentially block the processing for a certain time. For example, if the topic has a single partition, the processing can block for up to the full `pollFrequency` when the input records are spaced by more than the `pollTimeout`.

As a suggestion, maybe the unfulfilled requests could be requeued on `RunLoop.requestQueue`. That would immediately trigger an iteration of the loop.

tnielens:
Relates to #656 and #428.

guizmaii:
> The Requests that aren't unfulfilled

To be sure, do you mean "The Requests that aren't fulfilled"? 🤔

tnielens:
Yes! I've edited the description.

guizmaii:
I wonder if this issue isn't related to this: https://github.com/zio/zio-kafka/issues/428#issuecomment-1445352019 🤔

svroonland:
@erikvanoosten  To continue from https://github.com/zio/zio-kafka/issues/656#issuecomment-1455054058:

As car as I can see there are three reasons why we should execute poll()s:
1. To fetch new data for partitions with pending requests
2. At startup, to establish communication with the broker and get a partition assignment and start requests. We should keep polling during (eager) rebalancing as well, because we first revoke partition streams and there are no requests anymore after that. Also manual topic-partition assignment is a special case. 
3. Periodically (at least < `max.poll.interval.ms`) to signal that the consumer is still alive and should not be removed from the consumer group.

For 3, you could argue that if there are no pending requests for a while, the consuming application is not alive. I think that is the behavior we should aim for.

erikvanoosten:
[...]

PR #684, opened 2023-03-03
add test case for the runloop polling behavior

tnielens:
Relates to #664. Add test case demonstrating the expected behavior.

### DIFF
--- a/zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
+++ b/zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
@@ -1034,7 +1034,44 @@ object ConsumerSpec extends ZIOKafkaSpec {
 //          testForPartitionAssignmentStrategy[CooperativeStickyAssignor] // TODO not yet supported
         )
 
-      }: _*) @@ TestAspect.nonFlaky(3)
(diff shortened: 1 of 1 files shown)


svroonland:
@tnielens with #683 merged, what shall we do with this PR?

tnielens:
The test verifies that messages are streamed without delay in a certain poll timeout condition. 
I'll have a look at your pr to see if similar test cases were added. If not, I can finish it.

tnielens:
@svroonland I updated the branch. There are now two test cases. They test the requirements of #664 and currently fail. Could you have a look? I think these cases should pass if #664 is considered implemented.

svroonland:
@tnielens I checked it last night locally, if you update your branch with master, you should find that the tests succeed.

tnielens:
I've adapted the tests to the latest master commit. Feel free to close this PR if the test cases are redundant or provide no value to the project.

svroonland:
> Ah it seems the pollInterval field was removed from the ConsumerSettings. Was that part of release 2.1?
Part of one of the latest merged PRs (#736), not released yet.

svroonland:
We can keep the second test. The first one doesn't really test that it polls immediately, because it doesn't mean there will immediately be any records available in the stream.

svroonland (approved):

PR #689, opened 2023-03-06
Consumer benchmark

svroonland:
Far from perfect, but at least it gives us something to compare future changes to.


.github/workflows/benchs.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - throughput-bbenchmark
```
svroonland, line 7: Will remove this before merging.

zio-kafka-bench/src/main/resources/logback.xml
```diff
@@ -0,0 +1,18 @@
...
+
+    <logger name="org.apache.kafka" level="WARN" />
+    <logger name="state.change.logger" level="WARN" />
+
+    <logger name="kafka" level="WARN" />
+
+    <logger name="org.apache.zookeeper" level="ERROR" />
+
+    <root level="INFO">
+        <appender-ref ref="STDOUT" />
```
guizmaii, line 16: 🌕 Shouldn't we use an async logger here to avoid affecting performance measures?

zio-kafka-bench/src/main/scala/zio/kafka/bench/ConsumerBenchmark.scala
```diff
@@ -0,0 +1,93 @@
...
+import zio.kafka.KafkaTestUtils.{ consumer, produceMany, producer }
+import zio.kafka.bench.ZioBenchmark.randomThing
+import zio.kafka.consumer.{ Consumer, Offset, OffsetBatch, Subscription }
+import zio.kafka.embedded.Kafka
+import zio.kafka.producer.Producer
+import zio.kafka.serde.Serde
+import zio.stream.ZSink
+import zio.{ &, durationInt, Ref, Schedule, ZIO, ZLayer }
+
+object ConsumerBenchmark {
```
guizmaii, line 17: Why do we need this `Main`? 🤔

zio-kafka-bench/src/main/scala/zio/kafka/bench/ConsumerBenchmark.scala
```diff
@@ -0,0 +1,93 @@
...
+  @throws[RunnerException]
+  def main(args: Array[String]): Unit = {
+    val opt = new OptionsBuilder().include(classOf[ConsumerBenchmark].getSimpleName).forks(1).build
+    new Runner(opt).run
+    ()
+  }
+}
```
guizmaii, line 24: Why do we need this `main`? 🤔

svroonland: It was there in the example, I just renamed the object. Probably to run an individual benchmark class?

zio-kafka-bench/src/main/resources/logback.xml
svroonland: Good idea. Do you have an example config?

svroonland: Yeah. I only added it for debugging anyway. Removed it

zio-kafka-bench/src/main/scala/zio/kafka/bench/ConsumerBenchmark.scala
guizmaii: Ok. I think we can remove it TBH

guizmaii (approved):

PR #692, opened 2023-03-08
Simplify code for `newPendingCommits` selection

guizmaii:


svroonland (approved):
Nice, didn't know of ZIO.filter

PR #697, opened 2023-03-11
Performance improvements

guizmaii:
1. Reduce contention by evaluating only once `getConsumerGroupMetadataIfAny`

The `consumer.consumer.groupMetadata()` call uses an `AtomicLong` that generates a lot of contention for each `CommittableRecord` instantiated

2. Reduce allocation by transforming the `commit` function into a `val` containing a function.

This `commit` function is passed in each created `CommittableRecord` and was therefore instantiate for each `CommittableRecord` instantiated


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -270,7 +271,7 @@ private[consumer] final class Runloop(
     fulfillAction.as(Runloop.FulfillResult(unfulfilledRequests, newBufferedRecords))
   }
 
-  private def getConsumerGroupMetadataIfAny: Option[ConsumerGroupMetadata] =
+  private lazy val getConsumerGroupMetadataIfAny: Option[ConsumerGroupMetadata] =
```
guizmaii, line 274: Note that I'm not yet sure this is correct

jdegoes: When in doubt, dig into the implementation...

guizmaii: Yeah, that's what I'm working on here: https://github.com/zio/zio-kafka/pull/699

guizmaii: Well, what I've learnt from my work today is that the aggregation of `Offset`s into a `OffsetBatch` is anyway only keeping the reference of the first `Offset::consumerGroupMetadata` field

This is due to how we do this aggretation.
We `foldLeft(OffsetBatch.empty)(_ add _)` (See https://github.com/zio/zio-kafka/pull/700)
`OffsetBatch.empty.add` returns `offset.batch` which will copy the `Offset::consumerGroupMetadata` field into the returned `OffsetBatch` instance.
All the following calls to `offsetBatch::add` will only add the offset to the offsets map contained in the `OffsetBatch` and will never change anymore the `OffsetBatch::consumerGroupMetadata` field

So, computing only once, for the first offset, should not change the program's behaviour.

guizmaii:
Closing because this is too hacky. I need to think about it.

We probably need to find a way to compute this `consumerGroupMetatada` only when it's value in the consumer changes

erikvanoosten:
Today I heard a story that a Kafka cluster at my company once became heavily overloaded because one consumer was requesting metadata too often. So it seems a good idea to call this api as little as possible.

guizmaii:
Good to know! Thanks @erikvanoosten 🙏

PR #700, opened 2023-03-12
Rename `OffsetBatch::merge(Offset)` to `OffsetBatch::add(Offset)` to improve readability

guizmaii:


erikvanoosten:
Method `merge` is part of the API of zio-kafka. IMHO you can't change the name just like that. It will have to go through a deprecation cycle at least.
Besides all that, I find the name `merge` quite intuitive (though odd at first). However, renaming it to `add` does IMHO not improve readability.

guizmaii:
> Besides all that, I find the name merge quite intuitive (though odd at first). However, renaming it to add does IMHO not improve readability.

The thing I don't like is that you have 2 `merge` functions. One that is indeed merging 2 `OffsetBatch` and one that adds an `Offset` to the `OffsetBatch`. 
Having these 2 functions named similarly is confusing at the use site, especially because it's mostly used in a `foldLeft` function this way: `_ merge _`.
It's too complicated to know which version of the `merge` function is used in such code. I want code that reads fast, not code in which I need to guess/look for which version of a function is used

> Method merge is part of the API of zio-kafka. IMHO you can't change the name just like that. It will have to go through a deprecation cycle at least.

Hum indeed. Should this function be accessible to zio-kafka users tho? 🤔

svroonland (approved):

PR #701, opened 2023-03-12
Compute `getConsumerGroupMetadataIfAny` as few as possible

guizmaii:


guizmaii:
Closing because this is too hacky. I need to think about it.

We probably need to find a way to compute this `consumerGroupMetatada` only when it's value in the consumer changes

PR #710, opened 2023-03-16
Simplify `endRevoked` code

guizmaii:


svroonland (commented):
I'm in favour of this more immutable style, but I do believe that it was done this way for performance reasons. Could you check if it helps?

Also, can we do this as a separate PR? I would prefer not to pile code improvements on top of the other one if they're related to existing code.

guizmaii:
> I'm in favour of this more immutable style, but I do believe that it was done this way for performance reasons. Could you check if it helps?

To me these changes are not about immutable or mutable style. For zio-kafka code, we just need to peek the most performant one. Here, it's more performant (and simpler) this way as we're only iterating once over the lists of `requests` and we're don't need the buffer anymore.

PR #711, opened 2023-03-16
Make as much computation as soon as possible in `fulfillRequests` code

guizmaii:


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -254,17 +254,18 @@ private[consumer] final class Runloop(
       if (bufferedChunk.isEmpty && reqRecs.isEmpty) {
         acc += req
       } else {
-        val concatenatedChunk = bufferedChunk ++ Chunk.fromJavaIterable(reqRecs)
+        val concatenatedChunk =
+          (bufferedChunk ++ Chunk.fromJavaIterable(reqRecs)).map { record =>
+            CommittableRecord(
+              record = record,
+              commitHandle = commit,
+              consumerGroupMetadata = getConsumerGroupMetadataIfAny
```
svroonland, line 262: Can `getConsumerGroupMetadataIfAny` be computed once instead of for every record?

svroonland (commented):
I would prefer this as a separate PR

PR #712, opened 2023-03-16
`fulfillRequests`: prefer using a ZIO while loop over a long chain of `flatMap`

guizmaii:


erikvanoosten:
@guizmaii I though this was interesting. Why did you close it again? Was the performance not good?

guizmaii:
@erikvanoosten I closed it because Adam told me that it would probably not change anything, see https://discord.com/channels/629491597070827530/630498701860929559/1085913117965431004, and because I wasn't seeing anything in the benchmarks anyway

erikvanoosten:
I see. Well, if there is no down side, I found the way you structured the code in this PR more readable than what was there before. It took me some time before I realized that there was a `var` being mutated...

adamgfraser:
Yes may still make code easier to read and reason about!

guizmaii:
Ok, let me reopen this PR then
Thanks for the feedback  :)

guizmaii:
See https://github.com/zio/zio-kafka/pull/728

PR #713, opened 2023-03-16
Optimise `Runloop::run` code

guizmaii:


zio-kafka/src/main/scala/zio/kafka/consumer/diagnostics/DiagnosticEvent.scala
```diff
@@ -2,42 +2,29 @@ package zio.kafka.consumer.diagnostics
...
 
-  sealed trait Rebalance extends StableDiagnosticEvent
+  sealed trait Rebalance extends DiagnosticEvent
   object Rebalance {
     final case class Revoked(partitions: Set[TopicPartition])  extends Rebalance
     final case class Assigned(partitions: Set[TopicPartition]) extends Rebalance
     final case class Lost(partitions: Set[TopicPartition])     extends Rebalance
   }
 
-  final case class RunloopEvent(command: Command) extends InternalDiagnosticEvent
```
guizmaii, line 42: We don't need it anymore as we remove the tests for the `Runloop`

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -495,16 +490,16 @@ private[consumer] final class Runloop(
...
-  private def handleShutdown(state: State, cmd: Command): Task[State] =
-    cmd match {
-      case req: Request =>
-        req.end.as(state)
-      case r @ Command.ChangeSubscription(_, _, _) =>
-        r.succeed.as(state)
-      case _ @Command.Commit(_, cont) =>
-        cont.fail(new KafkaException("Consumer is shutting down")).as(state) // TODO can we just allow it?
-
-    }
```
guizmaii, line 507: Here, because I don't have the `.as(state)` anymore, I remove a lot of `.map(_ => state)` calls in the IOs chain

PR #719, opened 2023-03-17
Simplify `endRevoked` code

guizmaii:


svroonland (commented):
As talked about, I'd like to see the benchmark results on this

erikvanoosten:
I doubt this would impact performance but I am curious as well :)

erikvanoosten:
Note: this change collides with #727 since there we no longer need to end requests for revoked streams.

PR #721, opened 2023-03-17
Optimise `Runloop::run` code

guizmaii:


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -160,32 +166,21 @@ private[consumer] final class Runloop(
...
-  private def makeOffsetCommitCallback(
-    onSuccess: Task[Unit],
-    onFailure: Exception => Task[Unit]
-  ): OffsetCommitCallback =
-    new OffsetCommitCallback {
-      override def onComplete(offsets: util.Map[TopicPartition, OffsetAndMetadata], exception: Exception): Unit =
-        Unsafe.unsafe { implicit u =>
-          runtime.unsafe.run(if (exception eq null) onSuccess else onFailure(exception)).getOrThrowFiberFailure()
-        }
-    }
```
guizmaii, line 187: I didn't like this method. It's not reused and not doing enough things to justify extracting this code in a method IMO.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -138,18 +138,24 @@ private[consumer] final class Runloop(
       _ <- p.await
     } yield ()
 
-  private def doCommit(cmds: Chunk[Commit]): UIO[Unit] = {
-    val offsets   = aggregateOffsets(cmds)
-    val cont      = (e: Exit[Throwable, Unit]) => ZIO.foreachDiscard(cmds)(_.cont.done(e))
+  private def doCommit(cmd: Commit): UIO[Unit] = {
+    val offsets   = aggregateOffsets(cmd)
```
svroonland, line 142: I missed it in #683 , but `doCommit` is now only called for a single commit. There's no need to aggregate anymore.

The old poll behavior would batch up commits before executing a poll, we now execute them directly. This will execute send the commit to kafka directly, it's up to the user to do the aggregation on their side using `Consumer.offsetBatches` for example.

svroonland (changes requested):
We can get rid of aggregateOffsets altogether

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
guizmaii: Can we merge this PR and you make a PR to fix that please? I'm not sure to know how to fix that.

guizmaii: Made the change

svroonland (approved):

PR #725, opened 2023-03-18
Safer handling of commandQueue.takeBetween interruption

svroonland:


erikvanoosten:
If I read this correctly, `takeN` can still be interrupted, so we can still loose partially taken data.

erikvanoosten:
> If I read this correctly, `takeN` can still be interrupted, so we can still loose partially taken data.

Ah, the queue is restored. Tricky stuff.

zio-kafka-bench/src/main/scala/zio/kafka/bench/ConsumersComparisonBenchmark.scala
```diff
@@ -98,6 +98,10 @@ class ConsumersComparisonBenchmark extends ZioBenchmark[Env] {
       Consumer
         .plainStream(Subscription.topics(topic1), Serde.byteArray, Serde.byteArray)
         .take(nrMessages.toLong)
+        .tapErrorCause(e =>
+          ZIO.debug(s"Error in bench: ${e.prettyPrint}")
+        ) // It's weird but this prevents an Interrupt error somehow..
         .runDrain
+        .provideLayer(simpleConsumer())
```
svroonland, line 105: For some reason we need the `provideLayer` here, otherwise the stream won't start. Also the `tapErrorCause` is necessary to prevent something weird related to finalizer order, which seems like a bug in ZIO I'd like to further investigate separately..

erikvanoosten:
@adamgfraser Would you agree when I say that the methods added in this PR (`Queue.takeAll(timeout: Duration)` and `Queue.takeBetween(min: Int, max: Int, timeout: Duration)`) belong in zio-core?

erikvanoosten (commented):
This is not really an issue in zio-kafka since there is only 1 reader at all times. When there are multiple readers, I suspect that `DequeueWithTimeout` might lead to elements being retrieved twice.

I am think of the scenario where a timeout occurred and immediately afterwards 2 readers try to get elements. I suspect they will both get the items of the aborted dequeue.

svroonland:
> This is not really an issue in zio-kafka since there is only 1 reader at all times. When there are multiple readers, I suspect that DequeueWithTimeout might lead to elements being retrieved twice.
I think there's a misunderstanding here. This is not about multiple readers. The issue is that racing a `take` and a `timeout` may lead to the `take` fiber being interrupted shortly after having taken an item from the queue, which is then lost.

erikvanoosten:
> > This is not really an issue in zio-kafka since there is only 1 reader at all times. When there are multiple readers, I suspect that DequeueWithTimeout might lead to elements being retrieved twice.
> 
> I think there's a misunderstanding here. This is not about multiple readers. The issue is that racing a `take` and a `timeout` may lead to the `take` fiber being interrupted shortly after having taken an item from the queue, which is then lost.

Yes, I understand that. I was just thinking ahead; what would happen when we move the logic to ZStream in zio core? Then multiple readers could be a problem.

adamgfraser:
I think this implementation feels somewhat special purpose so belongs here. Also with `Queue` being an open interface now our ability to add new methods to it is limited. This does mean that you could just extend `Queue` instead of wrapping it.

This also feels very similar to `aggregateAsync`. Perhaps another approach would be to use streams here and have a stream of commands?

svroonland:
Marked as draft while I troubleshoot the benchmark getting stuck a bit more

zio-kafka/src/main/scala/zio/kafka/consumer/internal/DequeueWithTimeout.scala
```diff
@@ -0,0 +1,55 @@
+package zio.kafka.consumer.internal
+import zio.{ Chunk, Dequeue, Duration, Fiber, Ref, UIO, ZIO }
+
+/**
+ * Avoids race conditions between dequeueing and timeout, which would lead to lost dequeued elements, by storing the
+ * interrupted dequeue action and finishing it before starting a new dequeue
+ */
+class DequeueWithTimeout[A](q: Dequeue[A], previousDequeue: Ref[Option[Fiber[Nothing, Chunk[A]]]]) {
```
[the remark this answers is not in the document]
svroonland: Well, it replaces `commandQueue.takeBetween.timeoutTo(..)`, so it's consistent with that..

svroonland:
@adamgfraser I see what you mean about the similarity with `aggregateAsync`, but I'm not quite sure how to express our poll trigger logic in a stream, since it's not just driven by commands from the queue. FYI we also just moved a way from a stream of commands to this poll loop.

zio-kafka-bench/src/main/scala/zio/kafka/bench/ConsumersComparisonBenchmark.scala
guizmaii: This is weird, isn't it? It was working fine previously 🤔

zio-kafka/src/main/scala/zio/kafka/consumer/internal/DequeueWithTimeout.scala
```diff
@@ -0,0 +1,73 @@
+package zio.kafka.consumer.internal
+import zio.{ Chunk, Dequeue, Duration, Fiber, Ref, Scope, Semaphore, UIO, ZIO }
+
+/**
+ * Avoids race conditions between dequeueing and timeout, which would lead to lost dequeued elements, by storing the
+ * interrupted dequeue action and finishing it before starting a new dequeue
+ */
+class DequeueWithTimeout[A](
```
guizmaii, line 8:
```suggestion
final class DequeueWithTimeout[A](
```

zio-kafka-bench/src/main/scala/zio/kafka/bench/ConsumersComparisonBenchmark.scala
svroonland: It might have been introduced when adding the RunloopTimeout. I think the plain java consumer benchmark is running beyond that timeout. Note that the benchmark got stuck a lot of the time after merging #683

zio-kafka/src/main/scala/zio/kafka/consumer/internal/DequeueWithTimeout.scala
guizmaii, line 1: @adamgfraser Could you please review this code? It is using quite advanced ZIO things, I'd like to get your eyes on it

zio-kafka/src/main/scala/zio/kafka/consumer/internal/DequeueWithTimeout.scala
```diff
@@ -0,0 +1,73 @@
...
+                  // The forking is necessary to prevent interrupted exceptions
+                  permit
+                    .withPermit(
+                      awaitAction
+                        .forkIn(scope)
+                        .tap(fib => previousDequeue.set(Some(fib)))
+                    )
+                    .flatMap { awaitFib =>
+                      awaitFib.join
+                        .raceWith[Any, Nothing, Nothing, Chunk[A], Chunk[A]](
```
guizmaii, line 51: Here, the `raceWith` will fork (create a new fiber) based on `awaitFib.join` (which is already a Fiber) 🤔

@adamgfraser Is there a better/simpler way to do that? 🤔

svroonland:
Superseded by #736

PR #730, opened 2023-03-18
Feature/bench java client with prefetch

tnielens:
Add a java consumer benchmark with chunk prefetch to compare to baseline without prefetch. Interesting because zio-kafka prefetches partition chunks.

### DIFF
--- a/zio-kafka-bench/src/main/scala/zio/kafka/bench/ConsumersComparisonBenchmark.scala
+++ b/zio-kafka-bench/src/main/scala/zio/kafka/bench/ConsumersComparisonBenchmark.scala
@@ -10,7 +10,8 @@ import zio.kafka.consumer.{ Consumer, ConsumerSettings, Subscription }
 import zio.kafka.embedded.Kafka
 import zio.kafka.producer.Producer
 import zio.kafka.serde.Serde
-import zio.{ ULayer, ZIO, ZLayer }
+import zio.stream.ZStream
+import zio.{ Chunk, ULayer, ZIO, ZLayer }
 
 import java.util.concurrent.TimeUnit
 import scala.jdk.CollectionConverters._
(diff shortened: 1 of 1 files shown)


erikvanoosten (approved):
Nice!

zio-kafka-bench/src/main/scala/zio/kafka/bench/ConsumersComparisonBenchmark.scala
```diff
@@ -91,6 +92,25 @@ class ConsumersComparisonBenchmark extends ZioBenchmark[Env] {
...
+      for {
+        settings <- ZIO.service[ConsumerSettings]
+        consumer <- ZIO.service[LowLevelKafka]
+        _        <- ZIO.attemptBlocking(consumer.subscribe(java.util.Arrays.asList(topic1)))
+        _ <- ZStream
+               .repeatZIO(ZIO.attemptBlocking(consumer.poll(settings.pollTimeout)))
+               .map(records => Chunk.fromIterable(records.asScala))
+               .bufferChunks(if (settings.perPartitionChunkPrefetch > 1) settings.perPartitionChunkPrefetch else 16)
+               .flattenChunks
+               .take(nrMessages.toLong)
```
guizmaii, line 108: This will produce better performances:

```suggestion
        _ <- ZStream
               .repeatZIOChunk(ZIO.attemptBlocking(Chunk.fromIterable(consumer.poll(settings.pollTimeout).asScala)))
               .bufferChunks(if (settings.perPartitionChunkPrefetch > 1) settings.perPartitionChunkPrefetch else 16)
               .take(nrMessages.toLong)
```

guizmaii (commented):
See comment :)

tnielens:
Given that the KafkaConsumer fetch latency is probably low and the downstream work is negligible (a simple `.take(N).runDrain`, I don't think this case is very relevant. But if there was a series of cases with these aspects in action, it would be nice to demonstrate the effect of prefetching record batches before processing. Also the `max.poll.records` is misleading because the size of the batches fetched is not constrained by it, but instead by [fetch.max.bytes](https://kafka.apache.org/documentation/#brokerconfigs_fetch.max.bytes). It might very well be that the underlying KafkaConsumer fetches the entire topic content in a single fetch in the existing benchmark scenarios.

PR #731, opened 2023-03-19
Uninterruptable subscribe/unsubscribe

svroonland:
Fixes a race condition


erikvanoosten:
Since `updateZIO` guarantees atomic updates, do we need this to be uninterruptable?

svroonland:
Closed because it doesn't fix the benchmarks like I expected.

`modifyZIO` is implemented as `semaphore.withPermit(get.flatMap(f).flatMap { case (b, a) => ref.set(a).as(b) })`

It only guarantees that there is no concurrent execution, there's no protection against interruption.

PR #736, opened 2023-03-20
ZStream based implementation of poll loop

svroonland:
This avoids a race condition between commandQueue.takeAll and timeout, which could result in lost commands and therefore timeouts or stuck partition streams (data loss was not a risk).

Also, to fix flaky benchmark runs:

* Later confirmation of change subscription command
* Cleaner shutdown of runloop (is diferent from the graceful shutdown)
* Add logging (off by default) to benchmark for troubleshooting purposes

Note that before #683 we already had a stream-based poll loop, now we do again but with more optimized behavior.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
@@ -10,8 +10,7 @@ import zio.kafka.security.KafkaCredentialStore
  * @param properties
  * @param closeTimeout
  * @param pollInterval
- *   When there are no pending requests from partition streams, the frequency of polling for liveness and getting
- *   partition assignments.
+ *   Deprecated, no longer used.
  * @param pollTimeout
  * @param perPartitionChunkPrefetch
  * @param offsetRetrieval
@@ -27,6 +26,7 @@ case class ConsumerSettings(
   bootstrapServers: List[String],
   properties: Map[String, AnyRef],
   closeTimeout: Duration,
+  @deprecated("Is no longer used", "2.1.3")
   pollInterval: Duration,
   pollTimeout: Duration,
   perPartitionChunkPrefetch: Int,
@@ -70,6 +70,7 @@ case class ConsumerSettings(
   def withPerPartitionChunkPrefetch(prefetch: Int): ConsumerSettings =
     copy(perPartitionChunkPrefetch = prefetch)
 
+  @deprecated("no longer used", "2.1.3")
   def withPollInterval(interval: Duration): ConsumerSettings =
     copy(pollInterval = interval)
 
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
@@ -26,8 +26,6 @@ case class ConsumerSettings(
   bootstrapServers: List[String],
   properties: Map[String, AnyRef],
   closeTimeout: Duration,
-  @deprecated("Is no longer used", "2.1.3")
-  pollInterval: Duration,
   pollTimeout: Duration,
   perPartitionChunkPrefetch: Int,
   offsetRetrieval: OffsetRetrieval = OffsetRetrieval.Auto(),
@@ -70,10 +68,6 @@ case class ConsumerSettings(
   def withPerPartitionChunkPrefetch(prefetch: Int): ConsumerSettings =
     copy(perPartitionChunkPrefetch = prefetch)
 
-  @deprecated("no longer used", "2.1.3")
-  def withPollInterval(interval: Duration): ConsumerSettings =
-    copy(pollInterval = interval)
-
   def withPollTimeout(timeout: Duration): ConsumerSettings =
     copy(pollTimeout = timeout)
 
@@ -102,7 +96,6 @@ object ConsumerSettings {
       bootstrapServers = bootstrapServers,
       properties = Map.empty,
       closeTimeout = 30.seconds,
-      pollInterval = 50.millis,
       pollTimeout = 50.millis,
       perPartitionChunkPrefetch = 2,
       offsetRetrieval = OffsetRetrieval.Auto(),
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
@@ -9,8 +9,6 @@ import zio.kafka.security.KafkaCredentialStore
  * @param bootstrapServers
  * @param properties
  * @param closeTimeout
- * @param pollInterval
- *   Deprecated, no longer used.
  * @param pollTimeout
  * @param perPartitionChunkPrefetch
  * @param offsetRetrieval
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -552,36 +552,29 @@ private[consumer] final class Runloop(
       state.isSubscribed && (state.pendingRequests.nonEmpty || state.pendingCommits.nonEmpty || state.assignedStreams.isEmpty)
 
     ZStream
-      .fromZIO(Ref.make(State.initial))
-      .flatMap { stateRef =>
-        ZStream
-          .fromQueue(commandQueue)
-          .merge(ZStream.repeatWithSchedule(Command.PeriodicPoll, Schedule.spaced(pollFrequency)))
-          .aggregateAsync(ZSink.collectAllN[Command](commandQueueSize))
-          .mapZIO { commands =>
-            for {
-              state      <- stateRef.get
-              _          <- ZIO.logTrace(s"Processing ${commands.size} commands: ${commands.mkString(",")}")
-              isShutdown <- isShutdown
-              handleCommand = if (isShutdown) handleShutdown _ else handleOperational _
-              updatedState <- ZIO.foldLeft(commands)(state)(handleCommand)
+      .fromQueue(commandQueue)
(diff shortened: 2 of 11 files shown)


erikvanoosten (commented):
I am not sure. If I read this correctly this no longer calls poll continuously.

erikvanoosten:
> I am not sure. If I read this correctly this no longer calls poll continuously.

Nope, I read it incorrectly.

svroonland:
You are right, can't believe I missed that. I implemented a fix for that

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -541,41 +541,43 @@ private[consumer] final class Runloop(
...
+
+          updatedStateAfterPoll <- if (shouldPoll(updatedState))
+                                     logPollStart(updatedState) *> handlePoll(updatedState)
+                                   else ZIO.succeed(updatedState)
+          _ <-
+            commandQueue
+              .offer(Command.Poll)
+              .when(
+                shouldPoll(updatedStateAfterPoll)
+              ) // Immediately poll again, after processing all new queued commands
```
erikvanoosten, line 575: 😆 Yes, it can be that simple!

BTW, not sure what you think about this. I always prefer comments on a line before the statement.
```suggestion
           // Immediately poll again, after processing all new queued commands
           _ <-
            commandQueue
              .offer(Command.Poll)
              .when(shouldPoll(updatedStateAfterPoll))
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -541,41 +541,43 @@ private[consumer] final class Runloop(
...
-            } yield ()
-          }
-          .takeUntilZIO(_ => stop.get)
+    def shouldPoll(state: State): Boolean =
+      state.isSubscribed && (state.pendingRequests.nonEmpty || state.pendingCommits.nonEmpty || state.assignedStreams.isEmpty)
+
+    ZStream
+      .fromQueue(commandQueue)
+      .timeoutFail[Throwable](RunloopTimeout)(runloopTimeout)
+      .merge(ZStream.repeatWithSchedule(Command.Poll, Schedule.spaced(pollFrequency)))
```
erikvanoosten, line 557: Do we need this? Is the Command.Poll added in line 571 not enough?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -541,41 +541,43 @@ private[consumer] final class Runloop(
...
-              shouldPoll =
-                updatedState.isSubscribed && (updatedState.pendingRequests.nonEmpty || updatedState.pendingCommits.nonEmpty || updatedState.assignedStreams.isEmpty)
-              updatedStateAfterPoll <- if (shouldPoll) logPollStart(updatedState) *> handlePoll(updatedState)
-                                       else ZIO.succeed(updatedState)
-              _ <- stateRef.set(updatedStateAfterPoll)
-            } yield ()
-          }
-          .takeUntilZIO(_ => stop.get)
+    def shouldPoll(state: State): Boolean =
+      state.isSubscribed && (state.pendingRequests.nonEmpty || state.pendingCommits.nonEmpty || state.assignedStreams.isEmpty)
```
erikvanoosten, line 552: This can be moved to State. (I'll assume this is done in a next comment.)

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten, line 575: We can move calling handlePoll (including logging) into the 2 handle* methods. That would simplify this piece even further.

```suggestion
          updatedState <- ZIO.foldLeft(commands)(state)(handleCommand)

          // Immediately poll again, after processing all new queued commands
          _ <- commandQueue
              .offer(Command.Poll)
              .when(updatedState.shouldPoll)
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -530,7 +530,7 @@ private[consumer] final class Runloop(
    * Poll behavior:
    *   - Run until stop is set to true
    *   - Process commands as soon as they are queued, unless in the middle of polling
-   *   - Process all currently queued commands instead of one by one
+   *   - Process all currently queued commands before polling instead of one by one
    *   - Immediately after polling, if there are available commands, process them instead of waiting until some periodic
    *     trigger
    *   - Poll only when subscribed (leads to exceptions from the Apache Kafka Consumer if not)
```
erikvanoosten, line 536: Lines 536-538 can move along with `shouldPoll` (see next comment).

erikvanoosten (dismissed):
Amazingly simple. This is even more true when we compare this to the previous attemp (#725).

I think we simplify it even further by moving shouldPoll to State, and move handling the Poll command to the 2 handle* methods (where the other commands are handled).

erikvanoosten:
I would like to repeat the following (from #725 but a bit more worked out).

In normal circumstance the poll stream is continuously occupying a fiber. In particular the poll command blocks the thread for some time. I do not for sure how the ZIO runtime reacts to ZIOs that look like blocking, nor how it detects them. Anyways, if it moves the polling ZIO to a dedicated thread pool ([documentation suggests this](https://zio.dev/reference/core/zio/#blocking-synchronous-side-effects)), then we should run the entire stream on the blocking thread pool (or a dedicated thread pool) so that we're not switching between thread pools for blocking ZIOs and non-blocking ZIOs. This switching gives a high overhead.

We can ask an expert on whether this is worth the effort.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
svroonland: We want to process all commands besides poll first to update the state, and only then do a poll. If we handle the Poll command in handleOperational/handleShutdown, we haven't processed all other commands yet.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
svroonland: True but I think having the full poll behavior described here is useful

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -702,10 +693,10 @@ private[consumer] object Runloop {
...
                   currentStateRef
                 )
-      _    <- ZIO.logDebug("Starting Runloop").withFinalizer(_ => ZIO.logDebug("Shut down Runloop"))
-      stop <- Ref.make(false)
-      fib  <- runloop.run(stop).forkScoped
-      _    <- ZIO.addFinalizer(ZIO.logTrace("Shutting down Runloop") *> stop.set(true) *> fib.join.orDie)
+      _ <- ZIO.logDebug("Starting Runloop").withFinalizer(_ => ZIO.logDebug("Shut down Runloop"))
+      // Run the entire loop on the blocking thread pool to avoid executor shifts
+      fib <- ZIO.blocking(runloop.run).forkScoped
+      _ <- ZIO.addFinalizer(ZIO.logTrace("Shutting down Runloop") *> commandQueue.offer(StopRunloop) *> fib.join.orDie)
```
erikvanoosten, line 699: The suggestion removes double logging.

I was confused about the use of the *> operator. I though: this is not needed in a for comprehension. However, only then I noticed they are all within the finalizer block.
Perhaps some more newlines make this clearer:

```suggestion
      _ <- ZIO.logDebug("Starting Runloop")
      // Run the entire loop on the blocking thread pool to avoid executor shifts
      fib <- ZIO.blocking(runloop.run).forkScoped
      _ <- ZIO.addFinalizer(
         ZIO.logTrace("Shutting down Runloop") *>
         commandQueue.offer(StopRunloop) *>
         fib.join.orDie
      )
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: That is not correct. Poll is only added to the commandqueue when it is empty (unless more is added concurrently). Anyway, the effect will be the same: first the commandqueue is emptied, then a poll is added, and then the next loop picks up the poll command and executes it.
The advantage of the current implementation is that it is more explicit about the order of things (and I almost always like explicit more).

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -10,8 +10,7 @@ import zio.kafka.security.KafkaCredentialStore
  * @param properties
  * @param closeTimeout
  * @param pollInterval
- *   When there are no pending requests from partition streams, the frequency of polling for liveness and getting
- *   partition assignments.
+ *   Deprecated, no longer used.
```
erikvanoosten, line 13: If we remove `withPollInterval`, we might as well remove this parameter as well.

In either case, this breaks binary compatibility.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -468,27 +469,31 @@ private[consumer] final class Runloop(
...
+          .tapBoth(e => cmd.fail(e), _ => cmd.succeed)
+          .uninterruptible
+      case _: Command.Control => ZIO.succeed(state)
     }
 
   private def handleChangeSubscription(
-    state: State,
     command: Command.ChangeSubscription
-  ): Task[State] =
+  ): Task[Any] =
```
erikvanoosten, line 496: Not `Task[Unit]` ?

erikvanoosten (dismissed):

erikvanoosten (approved):
LGTM

PR #727, opened 2023-03-18
Feed streams through queue instead of promise

erikvanoosten:
Using a queue has the potential to simplify the runloop, and it unlocks new features:
 - runloop no longer needs to 'end' a request when there is no more data (e.g. when we're shutting down),
 - we can remove the concept of 'buffered data',
 - we can proactively get more data when we detect that a stream is paused every 2nd poll.
 
Of these, only the first is implemented. The other points are for future PRs.

This PR is the first step towards solving #706.

PR #744, opened 2023-03-25
Remove buffered records

erikvanoosten:
Previously we had to buffer records because the target stream was not created yet, or because it had not requested any data yet. Since #727 we no longer need to wait for a data request. Secondly, we now always create the stream _before_ we try to offer records to it.
As a consequence we no longer need to buffer records.

Streams for manually assigned partitions are now also stored in the runloop state. This was needed to find them for offering new data. A nice side effect is that they now properly end during a shutdown.

TODO: restore diagnostics around line 341?

Also:
- Rename `withConsumerM` to `withConsumerZIO` following ZIO 2 naming conventions.
- Make members private where possible.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -13,7 +13,6 @@ import zio.kafka.consumer.{ CommittableRecord, RebalanceConsumer, RebalanceListe
 import zio.stream._
 
 import java.util
-import scala.collection.mutable
 import scala.jdk.CollectionConverters._
 import scala.util.control.NonFatal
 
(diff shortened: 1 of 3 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -156,87 +155,71 @@ private[consumer] final class Runloop(
...
     currentAssignedStreams: Seq[PartitionStreamControl],
     isRevoked: TopicPartition => Boolean
   ): UIO[Runloop.RevokeResult] = {
     val (revokedStreams, assignedStreams) =
       currentAssignedStreams.partition(control => isRevoked(control.tp))
 
-    val revokeAction: UIO[Unit] =
-      ZIO.foreachDiscard(revokedStreams) { control =>
-        val remaining = bufferedRecords.recs.getOrElse(control.tp, Chunk.empty)
+    val pendingRequests = requests.filter(req => !isRevoked(req.tp))
```
svroonland, line 170: Maybe just inline it

svroonland (approved):
This is a really nice simplification, eliminating the concept of buffered records. 

Can you get the diagnostics event back?

svroonland (approved):

PR #745, opened 2023-03-26
Review of PR #727

guizmaii:
See #727 

Some minor code changes

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
@@ -63,7 +63,7 @@ private[internal] object PartitionStreamControl {
       interruptionPromise <- Promise.make[Throwable, Unit]
       completedPromise    <- Promise.make[Throwable, Unit]
       dataQueue           <- Queue.unbounded[Take[Throwable, ByteArrayCommittableRecord]]
-      requestData =
+      requestAndAwaitData =
         for {
           _     <- commandQueue.offer(Request(tp, dataQueue))
           _     <- diagnostics.emitIfEnabled(DiagnosticEvent.Request(tp))
@@ -81,7 +81,7 @@ private[internal] object PartitionStreamControl {
                  ZStream.repeatZIOChunk {
                    // First try to take all records that are available right now.
                    // When no data is available, request more data and await its arrival.
-                   dataQueue.takeAll.flatMap(data => if (data.isEmpty) requestData else ZIO.succeed(data))
+                   dataQueue.takeAll.flatMap(data => if (data.isEmpty) requestAndAwaitData else ZIO.succeed(data))
                  }.flattenTake
                    .interruptWhen(interruptionPromise)
     } yield new PartitionStreamControl(tp, stream, dataQueue, interruptionPromise, completedPromise)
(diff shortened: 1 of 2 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -74,14 +81,7 @@ private[internal] object PartitionStreamControl {
...
                    // When no data is available, request more data and await its arrival.
-                   dataQueue.takeAll
-                     .filterOrElse(_.nonEmpty) {
-                       for {
-                         _     <- commandQueue.offer(Request(tp, dataQueue)).unit
-                         _     <- diagnostics.emitIfEnabled(DiagnosticEvent.Request(tp))
-                         taken <- dataQueue.takeBetween(1, Int.MaxValue)
-                       } yield taken
-                     }
+                   dataQueue.takeAll.flatMap(data => if (data.isEmpty) requestData else ZIO.succeed(data))
```
guizmaii, line 84: I think that the code is simpler to read this way. The `.filterOrElse(_.nonEmpty) ...` operator is quite complex to grasp IMO.

WDYT?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -161,7 +161,7 @@ private[consumer] final class Runloop private (
    */
   private def endRevoked(
     requests: Chunk[Request],
-    currentAssignedStreams: Seq[PartitionStreamControl],
+    currentAssignedStreams: Chunk[PartitionStreamControl],
```
guizmaii, line 164: @erikvanoosten Why did you use a `Seq` here? 

It's usually considered a bad practice to use `Seq` as it's just an interface and can be backed by finite or infinite (streams) collections, so we don't really know what you're using here and your give the possibility to use an infinite collections to call your function here while it might not be what you're expecting

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
erikvanoosten: I have always found filterOrElse quite intuitive and I am disappointed if I am working with other effect systems and see its not there (cats-effects).
However, I have no problems with this approach.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -63,6 +63,13 @@ private[internal] object PartitionStreamControl {
       interruptionPromise <- Promise.make[Throwable, Unit]
       completedPromise    <- Promise.make[Throwable, Unit]
       dataQueue           <- Queue.unbounded[Take[Throwable, ByteArrayCommittableRecord]]
+      requestData =
```
erikvanoosten, line 66:
```suggestion
      requestAndAwaitData =
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: 🤷 I did not consider it. Chunk is fine by me.

erikvanoosten (dismissed):
LGTM

guizmaii:
Thanks for your review @erikvanoosten 🙏

PR #747, opened 2023-03-26
Review of PR #727: Remove unused `Request::dataQueue`

guizmaii:


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -53,7 +53,7 @@ private[consumer] final class Runloop private (
   def awaitShutdown: UIO[Unit] =
     for {
       state <- currentState.get
-      _     <- ZIO.foreachDiscard(state.assignedStreams)(_.awaitCompleted().ignore)
```
guizmaii, line 56: We don't need to `.ignore` anymore here as the Promise is now a `Promise[Nothing, Unit]`

svroonland (dismissed):

erikvanoosten (dismissed):
Ah yes, very good. This way to get the `dataQueue` is no longer used.

PR #749, opened 2023-03-26
Improve performances: Remove the `aggregateAsync` call

guizmaii:
To replicate the benchmark results:
```bash
sbt "zioKafkaBench/Jmh/run -wi 10 -i 10 -r 1 -w 1 -t 1 -f 5 -foe true zio.kafka.bench.ConsumersComparisonBenchmark"
```

Before (`master` branch):
```scala
[info]   2347.283 ±(99.9%) 495.618 ms/op [Average]
[info]   (min, avg, max) = (1143.906, 2347.283, 4427.192), stdev = 1001.173
[info]   CI (99.9%): [1851.665, 2842.901] (assumes normal distribution)
[info] # Run complete. Total time: 00:07:59
```

After:
```scala
[info]   1796.335 ±(99.9%) 161.723 ms/op [Average]
[info]   (min, avg, max) = (1169.957, 1796.335, 2808.878), stdev = 326.688
[info]   CI (99.9%): [1634.612, 1958.057] (assumes normal distribution)
[info] # Run complete. Total time: 00:06:36
```

IMO, these results are (I could be wrong, ofc) quite counter-intuitive/surprising.
This `aggregateAsync` call seems to be an optimisation: It reduces the number of operations for each message, some of these operations being done once per batch instead of once per message.
Therefore, IMO, it's quite surprising that removing this call has such a positive impact on the performances. 🤔

@adamgfraser Would you expect such results? Is there a `runFoldChunkZio` that could allow us to keep the chunking?

Cc @svroonland @erikvanoosten 

Edit:
FYI, I made an experiment here to keep the chunking: https://github.com/zio/zio-kafka/pull/751
The results are surprising too. The peak performance is better but the variability in the results is quite high

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -495,13 +495,12 @@ private[consumer] final class Runloop private (
       .fromQueue(commandQueue)
       .timeoutFail[Throwable](RunloopTimeout)(runloopTimeout)
       .takeWhile(_ != StopRunloop)
-      .aggregateAsync(ZSink.collectAllN[Command](commandQueueSize))
-      .runFoldZIO(State.initial) { case (state, commands) =>
+      .runFoldZIO(State.initial) { case (state, command) =>
         for {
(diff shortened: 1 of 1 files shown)


guizmaii:
Closing in favour of https://github.com/zio/zio-kafka/pull/751

PR #758, opened 2023-03-27
Direct access to underlying consumer when already on a blocking threadpool

svroonland:
* More direct access to ZIO consumer when already on a blocking threadpool (in the poll loop)

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -145,7 +145,7 @@ private[consumer] final class Runloop private (
           }
       }
 
-    consumer.withConsumerZIO { c =>
+    consumer.withConsumerZIONoFork { c =>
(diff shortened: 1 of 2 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -222,6 +224,7 @@ private[consumer] final class Runloop private (
             )
         }
       }
+      .fork // Do not await fulfilling of results
```
erikvanoosten, line 227: We have to be careful about this. Can we still guarantee that only fiber is offering data to the streams? I think not. For example the next poll loop can start offering records while we're still offering data from the previous loop. Records might be re-ordered.

We need some way to keep offering sequentially.

svroonland: Yeah, you're probably right about the sequentiality.

svroonland:
Obsolete by #802

PR #777, opened 2023-04-02
Commit immediately instead of waiting for poll to complete

svroonland:
Given a situation with a large poll timeout, no consumer lag and no new messages on the topic, when a commit is made, it will have to wait for an in-progress poll to complete before `KafkaConsumer#commitAsync` is called. It will then again have to wait for the duration of the poll timeout before the commit is confirmed. This is up to `2 * pollTimeout`. 

By committing immediately (during the poll), we can reduce the wait duration to max `1 * pollTimeout`.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -134,7 +134,7 @@ private[consumer] final class Runloop private (
     val onFailure: Throwable => UIO[Unit] = {
       case _: RebalanceInProgressException =>
         ZIO.logDebug(s"Rebalance in progress, retrying commit for offsets $offsets") *>
-          commandQueue.offer(cmd).unit
+          commandQueue.offer(cmd).unit // TODO: retry, but do not add to command queue again
       case err =>
         cont(Exit.fail(err)) <* diagnostics.emitIfEnabled(DiagnosticEvent.Commit.Failure(offsets, err))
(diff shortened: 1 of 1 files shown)


svroonland:
This might also have the benefit of shorter delays between poll calls, because we have to perform less operations to handle the commands between polling.

erikvanoosten:
I was building an argument for doing this as well. However, your argument is more immediate and much more compelling :)

erikvanoosten:
This PR is not complete yet. Some loose remarks:

I suggest we we rename the Commit command to AddPendingCommit or AddCommit.


```scala
      case _: RebalanceInProgressException =>
        ZIO.logDebug(s"Rebalance in progress, retrying commit for offsets $offsets") *>
commandQueue.offer(cmd).unit
```

Putting it on the command queue again doesn't make sense anymore (the commit is already in `pendingCommits`). More interesting question is: when/how do we do retries?


```scala
    consumer.withConsumerZIO { c =>
      ZIO.attempt(c.commitAsync(offsets.asJava, callback))
    }
```

I was working on a proposal to _not_ use `ConsumerAccess.withConsumerZIO` here. Consumer.commit can always be invoked from any thread. Also when a poll is in progress (so when the consumer is already in use).

svroonland:
> I was working on a proposal to not use ConsumerAccess.withConsumerZIO here. 
Ah you are right, this is essential. We can't get the permit while poll is in progress..

erikvanoosten:
Working on a different approach. Closing for now.

PR #778, opened 2023-04-02
Simpler shutdown handling

svroonland:
Just to recap: graceful shutdown is a mechanism to stop streaming but allow the consuming application to finish processing 'in-flight' records and commit them. Poll calls are still done to get the confirmation of these commits.

Graceful shutdown already ends all active partition streams, so there is no need for `handleShutdown` anymore.

This incidentally improves the `throughput` benchmark by 33%.. From 300 ms to less than 200 ms..

```
[info] Result "zio.kafka.bench.ConsumerBenchmark.throughput":
[info]   187,258 ±(99.9%) 2,681 ms/op [Average]
[info]   (min, avg, max) = (180,438, 187,258, 193,896), stdev = 3,580
[info]   CI (99.9%): [184,577, 189,940] (assumes normal distribution)
```

Also:
* fixes a race condition where the `currentState` ref was not up to date, leading to not all partition streams being ended.
* Because 'shutdown' is no longer a state, after all active streams have been ended, the consumer can resubscribe and use `stopConsumption` again.

Future followup: stopConsumption is now global over all (multiple) subscriptions, make this stoppable per subscription.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -25,28 +25,18 @@ private[consumer] final class Runloop private (
   lastRebalanceEvent: Ref.Synchronized[Option[Runloop.RebalanceEvent]],
   val partitions: Queue[Take[Throwable, (TopicPartition, Stream[Throwable, ByteArrayCommittableRecord])]],
   diagnostics: Diagnostics,
-  shutdownRef: Ref[Boolean],
   offsetRetrieval: OffsetRetrieval,
   userRebalanceListener: RebalanceListener,
   restartStreamsOnRebalancing: Boolean,
   currentState: Ref[State]
 ) {
-  private val isShutdown = shutdownRef.get
 
   private def newPartitionStream(tp: TopicPartition): UIO[PartitionStreamControl] =
     PartitionStreamControl.newPartitionStream(tp, commandQueue, diagnostics)
 
   /** Initiate a graceful shutdown. */
   def gracefulShutdown: UIO[Unit] =
-    ZIO
-      .whenZIO(shutdownRef.getAndSet(true).negate) {
-        for {
-          state <- currentState.get
-          _     <- ZIO.foreachDiscard(state.assignedStreams)(_.end())
-          _     <- partitions.offer(Take.end)
-        } yield ()
-      }
-      .unit
+    commandQueue.offer(Command.StopFetching).unit
 
   /** Wait until graceful shutdown completes. */
   def awaitShutdown: UIO[Unit] =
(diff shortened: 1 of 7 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -502,10 +465,8 @@ private[consumer] final class Runloop private (
       .takeWhile(_ != StopRunloop)
       .runFoldChunksDiscardZIO(State.initial) { (state, commands) =>
         for {
-          _          <- ZIO.logTrace(s"Processing ${commands.size} commands: ${commands.mkString(",")}")
-          isShutdown <- isShutdown
-          handleCommand = if (isShutdown) handleShutdown _ else handleOperational _
-          updatedState <- ZIO.foldLeft(commands)(state)(handleCommand)
+          _            <- ZIO.logTrace(s"Processing ${commands.size} commands: ${commands.mkString(",")}")
+          updatedState <- ZIO.foldLeft(commands)(state)(handleCommand _)
```
erikvanoosten, line 469: Can we do this?
```suggestion
          updatedState <- ZIO.foldLeft(commands)(state)(handleCommand)
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -575,8 +536,9 @@ private[consumer] object Runloop {
     // Used for internal control of the runloop
     sealed trait Control extends Command
 
-    case object Poll        extends Control
-    case object StopRunloop extends Control
+    case object Poll         extends Control
+    case object StopRunloop  extends Control
+    case object StopFetching extends Control
```
erikvanoosten, line 541: 'Fetching' is not a word that we use yet. I suggest we stick to the terms that we have. E.g. something like:

```suggestion
    case object StopAllStreams extends Control
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -639,7 +599,6 @@ private[consumer] object Runloop {
 
       _ <- ZIO.addFinalizer(
              ZIO.logTrace("Shutting down Runloop") *>
```
erikvanoosten, line 601: We should also end the streams here!

erikvanoosten (commented):
I absolutely love this; doing the same thing with less code are always the best changes.

Some nitpicking in the comments.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
svroonland: Apparently, yes :)

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Can be done in a separate PR.

erikvanoosten (approved):

PR #781, opened 2023-04-03
Backport Conduktor's improvements

guizmaii:


zio-kafka/src/main/scala/zio/kafka/admin/AdminClient.scala
```diff
@@ -191,15 +192,15 @@ trait AdminClient {
    */
   def listConsumerGroupOffsets(
     groupSpecs: Map[String, ListConsumerGroupOffsetsSpec]
-  ): Task[Map[TopicPartition, OffsetAndMetadata]]
+  ): Task[Map[String, Map[TopicPartition, OffsetAndMetadata]]]
```
erikvanoosten, line 195: This is not an obvious change. What are the keys in the outer map representing?

The type signature is quite complex, can we somehow simplify this with case classes? Does it make sense to return a high overhead type like a Map, can we return a List[pair] instead?

This is a public API so we should consider retaining the original method.

erikvanoosten: I see from the code that the outer map contains groupIds as keys. This should be documented in the scaladoc.

zio-kafka/src/main/scala/zio/kafka/admin/AdminClient.scala
```diff
@@ -627,33 +628,43 @@ object AdminClient {
...
           adminClient
             .listConsumerGroupOffsets(groupSpecs.map { case (groupId, offsetsSpec) =>
               (groupId, offsetsSpec.asJava)
             }.asJava)
-            .partitionsToOffsetAndMetadata()
+            .all()
         )
       }
-        .map(_.asScala.filter { case (_, om) => om ne null }.toMap.bimap(TopicPartition(_), OffsetAndMetadata(_)))
+        .map(
```
erikvanoosten, line 641:
```suggestion
        .map {
```

zio-kafka/src/main/scala/zio/kafka/admin/AdminClient.scala
```diff
@@ -627,33 +628,43 @@ object AdminClient {
...
             )
-            .partitionsToOffsetAndMetadata()
+            .all()
         )
       }
-        .map(_.asScala.filter { case (_, om) => om ne null }.toMap.bimap(TopicPartition(_), OffsetAndMetadata(_)))
+        .map(
+          _.asScala.map { case (groupId, offsets) =>
+            groupId -> offsets.asScala.filter { case (_, om) => om ne null }.toMap
+              .bimap(TopicPartition(_), OffsetAndMetadata(_))
```
erikvanoosten, line 665: If efficiency is very important here, we should do the transformation (done by bimap) before we convert the collection into a map.

In that case it could be as simple as:
```suggestion
            groupId -> offsets.asScala
              .filter { case (_, om) => om ne null }
              .bimap(TopicPartition(_), OffsetAndMetadata(_))
              .toMap
```
(And we need to define a new bimap that works on `List[(A,B)]`.)

zio-kafka/src/main/scala/zio/kafka/admin/AdminClient.scala
```diff
@@ -627,33 +628,43 @@ object AdminClient {
...
             }.asJava)
-            .partitionsToOffsetAndMetadata()
+            .all()
         )
       }
-        .map(_.asScala.filter { case (_, om) => om ne null }.toMap.bimap(TopicPartition(_), OffsetAndMetadata(_)))
+        .map(
+          _.asScala.map { case (groupId, offsets) =>
+            groupId -> offsets.asScala.filter { case (_, om) => om ne null }.toMap
+              .bimap(TopicPartition(_), OffsetAndMetadata(_))
```
erikvanoosten, line 644: Same as below.

zio-kafka-test/src/test/scala/zio/kafka/OOMSpecXmx300m.scala
```diff
@@ -0,0 +1,88 @@
...
+  override def spec: Spec[Kafka, Any] =
+    suite("OOM check")(
+      test("producer should fail with ssl check") {
+        for {
+          result <- (for {
+                      topic <- randomTopic
+                      _     <- Producer.produce(new ProducerRecord(topic, "boo", "baa"), Serde.string, Serde.string)
+                    } yield ()).provideSomeLayer(KafkaTestUtils.producer).exit
+        } yield assert(result)(
+          fails(
```
erikvanoosten, line 29: The whole fail expression is the same in each test (right?). It, or a part of it, could be extracted to a variable.

zio-kafka-test/src/test/scala/zio/kafka/OOMSpecXmx300m.scala
```diff
@@ -0,0 +1,88 @@
...
+import zio.test._
+
+/**
+ * This test checks the fix for the issue https://issues.apache.org/jira/browse/KAFKA-4090
+ */
+object OOMSpecXmx300m extends ZIOSpecWithSslKafka {
+
+  override val kafkaPrefix: String = "oom-spec"
+  override def spec: Spec[Kafka, Any] =
+    suite("OOM check")(
```
erikvanoosten, line 21: We're not testing for OOM so this name is misleading.
```suggestion
    suite("Plain connection on TLS port")(
```

zio-kafka-test/src/test/scala/zio/kafka/OOMSpecXmx300m.scala
```diff
@@ -0,0 +1,88 @@
...
+      test("producer should fail with ssl check") {
```
erikvanoosten, line 22:
```suggestion
      test("producer should fail due to ssl check") {
```

zio-kafka-test/src/test/scala/zio/kafka/OOMSpecXmx300m.scala
```diff
@@ -0,0 +1,88 @@
...
+                _.getCause.getMessage,
+                equalTo(
+                  "Received an unexpected SSL packet from the server. Please ensure the client is properly configured with SSL enabled"
+                )
+              )
+            )
+          )
+        )
+      },
+      test("consumer should fail with ssl check") {
```
erikvanoosten, line 42:
```suggestion
      test("consumer should fail due to ssl check") {
```

zio-kafka-test/src/test/scala/zio/kafka/OOMSpecXmx300m.scala
```diff
@@ -0,0 +1,88 @@
...
+                _.getCause.getMessage,
+                equalTo(
+                  "Received an unexpected SSL packet from the server. Please ensure the client is properly configured with SSL enabled"
+                )
+              )
+            )
+          )
+        )
+      },
+      test("admin client should fail with ssl check") {
```
erikvanoosten, line 68:
```suggestion
      test("admin client should fail due to ssl check") {
```

zio-kafka-test/src/test/scala/zio/kafka/OOMSpecXmx300m.scala
```diff
@@ -0,0 +1,88 @@
...
+import zio.kafka.producer.Producer
+import zio.kafka.serde.Serde
+import zio.test.Assertion._
+import zio.test.TestAspect._
+import zio.test._
+
+/**
+ * This test checks the fix for the issue https://issues.apache.org/jira/browse/KAFKA-4090
+ */
+object OOMSpecXmx300m extends ZIOSpecWithSslKafka {
```
erikvanoosten, line 17: Where is the name of this class coming from?

zio-kafka-test-utils/src/main/scala/zio/kafka/KafkaTestUtils.scala
```diff
@@ -11,8 +11,15 @@ import zio.kafka.embedded.Kafka
 import zio.kafka.producer._
 import zio.kafka.serde.{ Deserializer, Serde, Serializer }
 
+import java.io.File
+import java.nio.file.Paths
+
 object KafkaTestUtils {
 
+  val trustStoreFile: File = Paths.get(this.getClass.getResource("/truststore/kafka.truststore.jks").toURI).toFile
```
erikvanoosten, line 19: Does this also work when the resource is in a jar on the classpath?

zio-kafka-test-utils/src/main/resources/keystore/kafka.keystore.jks
erikvanoosten, line 1: In order to call this open source we need to have documentation on how to create these binary files. A readme.md with instructions in this directory would be nice, a bash script would be amazing. Please also state the purpose of these files.

erikvanoosten (approved):
I see a lot of complexity but I understand that is worth it. Having good failures is important for dev UX.

Some nitpicking: I usually use {} instead of () when an expression spans multiple lines.

zio-kafka/src/main/scala/zio/kafka/admin/AdminClient.scala
guizmaii: > The type signature is quite complex, can we somehow simplify this with case classes? Does it make sense to return a high overhead type like a Map, can we return a List[pair] instead?

I don't think so. It's a mapping. It can't be expressed with another type than a Map without losing information.

For example, a `List[(String, Map[TopicPartition, OffsetAndMetadata])]` is not equivalent.

Therefore, I'm not sure to see what you have in mind 🤔
Can you give me an example, please?

guizmaii: > This is not an obvious change. What are the keys in the outer map representing?

> I see from the code that the outer map contains groupIds as keys. This should be documented in the scaladoc.

Well, I agree that I can improve the doc, I think disagree about the complexity.
The user has to pass a `groupSpecs: Map[String, ListConsumerGroupOffsetsSpec]` to call this function and the function is called `listConsumerGroupOffsets` so it's quite obvious that the Strings in the result are corresponding to the Strings in that passed parameter

zio-kafka/src/main/scala/zio/kafka/admin/AdminClient.scala
guizmaii: Done

zio-kafka/src/main/scala/zio/kafka/admin/AdminClient.scala
guizmaii: Done

zio-kafka-test/src/test/scala/zio/kafka/OOMSpecXmx300m.scala
guizmaii: Fixed

zio-kafka/src/main/scala/zio/kafka/admin/AdminClient.scala
erikvanoosten: To be honest, this is one of the things I dislike about ZIO documentation. It is quite terse. This is a problem because what is obvious can be totally different for different people. I really like this aspect of the scala community guidelines for inclusitivity where they tell you to avoid words like 'just' and 'simply'. What those mean is different for everyone and when it is not obvious for someone that person may feel excluded.

erikvanoosten: There is no strong need to change the type, I was only wondering if we could make it more explicit. Adding a bit more documentation can't hurt much.

erikvanoosten (commented):
Lets work together on the script that creates the binary files.

zio-kafka-test-utils/src/main/scala/zio/kafka/KafkaTestUtils.scala
guizmaii: Should be fixed now :)

Thanks for spotting this!

zio-kafka/src/main/scala/zio/kafka/admin/AdminClient.scala
guizmaii: I updated the documentation. Please let me know what you think about it :)

erikvanoosten:
Could you please put this PR in draft until it is ready for review again? Or at least let us know when it is ready?

zio-kafka-test-utils/src/main/resources/keystore/kafka.keystore.jks
guizmaii: I added README file next to these files to explain how they've been generated

guizmaii:
@erikvanoosten I took all your comments into account, I think. The PR is ready to be reviewed (finally! 😅)

zio-kafka-test-utils/src/main/resources/README.md
```diff
@@ -0,0 +1,8 @@
+# Keystore and truststore files generation
+
+The files:
+ - `keystore/kafka.keystore.jks`
+ - `truststore/kafka.truststore.jks`
+ - `truststore/ca-key`
+
+were generated thanks to https://github.com/confluentinc/confluent-platform-security-tools/blob/master/kafka-generate-ssl.sh
```
erikvanoosten, line 8: Can you change it something like the following? The script requires many parameters and it is not clear which.

```suggestion
were generated with https://github.com/confluentinc/confluent-platform-security-tools/blob/master/kafka-generate-ssl.sh using the following input:

'''
Do you need to generate a trust store and associated private key? [yn] --> y
Password for the private key --> ...
Your name, company and CN -> Any value you like
the trust store's password (labeled 'keystore') -> ...
... etc ...
'''
```

zio-kafka-test-utils/src/main/scala/zio/kafka/KafkaTestUtils.scala
```diff
@@ -11,8 +11,27 @@ import zio.kafka.embedded.Kafka
...
 
+import java.io.File
+import java.nio.file.{ Files, StandardCopyOption }
+
 object KafkaTestUtils {
 
+  private def readResourceFile(file: String, tmpFileName: String, tmpFileSuffix: String): File =
+    try {
+      val tmpFile = Files.createTempFile(tmpFileName, tmpFileSuffix)
+      Files.copy(getClass.getClassLoader.getResourceAsStream(file), tmpFile, StandardCopyOption.REPLACE_EXISTING)
```
erikvanoosten, line 22: Nice.

erikvanoosten, line 22:
```suggestion
      Files.copy(getClass.getClassLoader.getResourceAsStream(file), tmpFile, StandardCopyOption.REPLACE_EXISTING)
      tmpFile.deleteOnExit()
```

zio-kafka-test-utils/src/main/scala/zio/kafka/ZIOSpecWithKafka.scala
erikvanoosten, line 1: You can't just delete this file. All my company's zio-kafka projects will start to fail if you do that!

Okay, okay, if there is a good reason, you could. But then we need good documentation on how to write tests. WDYT?

zio-kafka/src/main/scala/zio/kafka/admin/AdminClient.scala
erikvanoosten: Thanks. I guess people that need this feature will be able to find and use it with this description.

erikvanoosten (approved):
LGTM. Can we keep `ZIOSpecWithKafka.scala` for now?

zio-kafka-test-utils/src/main/resources/README.md
guizmaii: Everything is explained in the Confluent repo, I don't want us to repeat the instructions

zio-kafka-test-utils/src/main/scala/zio/kafka/ZIOSpecWithKafka.scala
guizmaii: Ok, I'll put it back, but I'll deprecate it. 

That's not the correct way to bring this Kafka instance into your tests. That creates a lot of issues when zio-test runs tests in parallel.

erikvanoosten (approved):

PR #784, opened 2023-04-05
Simplify

erikvanoosten:
Remove double logging, remove unnecessary match statement.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -156,10 +156,7 @@ private[consumer] final class Runloop private (
       assignedStreams.partition(control => isRevoked(control.tp))
 
     ZIO
-      .foreachDiscard(revokedStreams) { control =>
-        ZIO.logDebug(s"Revoking topic-partition ${control.tp}") *>
-          control.end()
-      }
+      .foreachDiscard(revokedStreams)(control => control.end())
       .as(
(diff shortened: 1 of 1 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -267,21 +264,11 @@ private[consumer] final class Runloop private (
             val records = doPoll(c)
 
             val currentAssigned = c.assignment().asScala.toSet
+            val newlyAssigned   = currentAssigned -- prevAssigned
```
guizmaii, line 267: Doesn't this change the behaviour when `rebalanceEvent` is `Assigned` or `RevokedAndAssigned`? 🤔

erikvanoosten: I don't think so. The match expression is a performance optimization that only applies when you enabled `restartStreamsOnRebalancing` which is normally not the case. In fact we slowing down the use case that should be fast while we're marginally speeding up the use case where a bit extra delay is acceptable.

erikvanoosten: 😄  The unit tests do not seem to agree with my argument...

erikvanoosten: I see now. We need to revoke all partitions... This PR is broken.

Issue #706, opened 2023-03-14
Rewrite runloop


erikvanoosten:
Zio-kafka has a very nice programming model but has a few design choices that make it only suitable for high throughput use cases where duplicating or losing a few messages is not an issue.

This issue tries to summarize the problems and gives a rough design for a new runloop that will fix these problems.

# Problems

~*Problem 1 - commits can be arbitrarily delayed*~

Currently commits are delayed during a rebalance. However, there is no good way to know when a rebalance is happening. Unfortunately, using the rebalance listener for that is not reliable. For example when the only change in the rebalance is that a partition is revoked. In this case we do not see the end of a rebalance because only onRevoke will be called, and onAssigned will not be.

**Update:** ✅ With #683 commits are no longer blocked by a rebalance. Instead, they are directly sent to the underlying Kafka client.

*Problem 2 - message duplication after rebalance*

A clean rebalance does not lead to message duplication when a revoked partition gets the chance to finish its work and commit everything before the rebalance starts. This is the purpose of the onRevoked callback of the rebalance listener. Currently, we do not wait inside the callback until the streams that handle the revoked partitions are complete.

**Update:** This is implemented in https://github.com/zio/zio-kafka/pull/788.

~*Problem 3 - polling delays*~

Polling frequency is determined by the poll delay parameter. Better latency and throughput is achieved when poll is called continuously. An experiment by @svroonland (see #683) showed that calling poll continuously gives a ridiculous improvement in throughput.

**Update:** ✅ #683 which implemented this idea has been merged!

*Problem 4 - transactional producer is not well supported*

It is currently quite difficult to use a transactional producer with zio-kafka. It requires the use of a custom rebalance listener and the _restartStreamOnRebalancing_ mode. With the design proposed in this issue working with transactional producers becomes as easy as regular producers.

Another problem is that when a stream needs to be restarted, that the new stream can already start processing while the old stream is still processing older messages of the same partition. That means that all ordering guarantees are off and that commits might be executed in the wrong order (which can lead to duplicates).

# Related issues

This issue is a duplicate of/is related with:

#590: duplicates after rebalance
#664 and #428: optimize polling behavior
#479 is no longer needed since we'll only close the correct streams

What follows here is a rough design for a new run loop.

--------------------------

# Runloop

_NOTE: This document describes an alternative Runloop implementation, not the currently implemented Runloop_

[...]

PR #788, opened 2023-04-08
Wait for stream end in rebalance listener

erikvanoosten:
This PR adds the setting `rebalanceSafeStreamEnd`. When set to true rebalances are hold up until the streams that handle the revoked partitions have ended. This prevents duplicate processing.

While waiting for the streams to end we need to continuously poll the broker. We do this by calling `commitAsync`. By inspecting the Java consumer code I found that when the passed Map of offsets is empty, the Java consumer will still poll the broker which is exactly what we need.

Commits are send to a separate queue (separate from commands) so that it is easier to access them while waiting for the streams to end.

The rebalance listener no longer calculates the next state. Instead, it only records what happened. The caller of poll then handles this data and calculates the next state. In case the rebalance listener was not invoked, we can skip the handling for a small performance improvement.

This fixes most of #706 

Tasks:

- [x] avoid the confusing terminology 'before rebalance'
- [x] find a nicer way to pass the consumer to the rebalance listener
- [x] remove methods on RebalanceConsumer (note, did not remove RebalanceConsumer itself)
- [x] add timeout in the commit-loop inside rebalance listener
- [ ] experiment with alternative approach (using only command-queue plus buffered commands)
- [x] extract ConsumerSpec.scala to a separate PR
- [ ] reduce logging on INFO and DEBUG level
- [ ] run all tests on ConsumerSpec and SubscriptionSpec twice: with rebalanceSafeStreamEnd true and false (note: locally all tests pass for both)

Alternative for the last bullet: we could decide to not support `rebalanceSafeStreamEnd` = false.

For a later PR:
 * configure zio-kafka via use-case profiles
 * experiment with removing transactionalRebalanceListener from consumerSpec
 * see if we can support the CooperativeStickyAssignor

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
@@ -14,6 +14,8 @@ import zio.kafka.security.KafkaCredentialStore
  * @param offsetRetrieval
  * @param rebalanceListener
  * @param restartStreamOnRebalancing
+ *   When `true` _all_ streams are restarted during a rebalance, including those streams that are not revoked. The
+ *   default is `false`.
  * @param runloopTimeout
  *   Internal timeout for each iteration of the command processing and polling loop, use to detect stalling. This should
  *   be much larger than the pollTimeout and the time it takes to process chunks of records. If your consumer is not
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
@@ -17,17 +17,25 @@ import zio.kafka.security.KafkaCredentialStore
  *   When `true` _all_ streams are restarted during a rebalance, including those streams that are not revoked. The
  *   default is `false`.
  *
- * Set both `restartStreamOnRebalancing` and `endRevokedStreamsBeforeRebalance` to `true` for transactional producing.
- * @param endRevokedStreamsBeforeRebalance
- *   When `true` (the default) streams that need to end because the partition has been revoked, will be ended before the
- *   rebalance starts. The consumer that takes over this partition will continue from the committed offset. However, it
- *   is not possible to commit during a rebalance. So holding up the rebalance until the stream has ended (and done its
- *   commits) will prevent duplicate processing.
+ * Set both `restartStreamOnRebalancing` and `rebalanceSafeStreamEnd` to `true` for transactional producing.
+ * @param rebalanceSafeStreamEnd
+ *   When `true` (the default) we hold up a rebalance until ending streams have completed. This is the recommended
+ *   setting.
  *
- * Set this to `false` when your streams does need commits, or when you need the extra performance and do not care for
- * duplicate processing.
+ * When a partition is revoked from this consumer due to a rebalance, we allow the stream to complete so that we know
+ * that any offset commits are done. The consumer that takes over this partition will then continue from the committed
+ * offset. Therefore, by holding up the rebalance, we ensure that the new consumer will start consuming from the correct
+ * offset.
  *
- * Set both `restartStreamOnRebalancing` and `endRevokedStreamsBeforeRebalance` to `true` for transactional producing.
+ * When `false`, streams for revoked partitions may continue to run to completion but the rebalance is not held up. Any
+ * offset commits from such a stream have a high chance of being delayed (commits are not possible during some phases of
+ * a rebalance). The consumer that takes over the partition will likely not see these delayed commits and will start
+ * from an earlier offset. The result is that some messages are processed twice and concurrently.
+ *
+ * Use `false` _only_ when your streams does not do commits, or when it is okay to have messages processed twice
+ * concurrently and you cannot afford the performance hit during a rebalance.
+ *
+ * Set both `restartStreamOnRebalancing` and `rebalanceSafeStreamEnd` to `true` for transactional producing.
  * @param runloopTimeout
  *   Internal timeout for each iteration of the command processing and polling loop, use to detect stalling. This should
  *   be much larger than the pollTimeout and the time it takes to process chunks of records. If your consumer is not
@@ -43,7 +51,7 @@ case class ConsumerSettings(
   offsetRetrieval: OffsetRetrieval = OffsetRetrieval.Auto(),
   rebalanceListener: RebalanceListener = RebalanceListener.noop,
   restartStreamOnRebalancing: Boolean = false,
-  endRevokedStreamsBeforeRebalance: Boolean = true,
+  rebalanceSafeStreamEnd: Boolean = true,
   runloopTimeout: Duration = ConsumerSettings.defaultRunloopTimeout
 ) {
   private[this] def autoOffsetResetConfig: Map[String, String] = offsetRetrieval match {
@@ -99,8 +107,8 @@ case class ConsumerSettings(
   def withRestartStreamOnRebalancing(value: Boolean): ConsumerSettings =
     copy(restartStreamOnRebalancing = value)
 
-  def withEndRevokedStreamsBeforeRebalance(value: Boolean): ConsumerSettings =
-    copy(endRevokedStreamsBeforeRebalance = value)
+  def withRebalanceSafeStreamEnd(value: Boolean): ConsumerSettings =
+    copy(rebalanceSafeStreamEnd = value)
 
   def withCredentials(credentialsStore: KafkaCredentialStore): ConsumerSettings =
     withProperties(credentialsStore.properties)
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
@@ -16,6 +16,18 @@ import zio.kafka.security.KafkaCredentialStore
  * @param restartStreamOnRebalancing
  *   When `true` _all_ streams are restarted during a rebalance, including those streams that are not revoked. The
  *   default is `false`.
+ *
+ * Set both `restartStreamOnRebalancing` and `endRevokedStreamsBeforeRebalance` to `true` for transactional producing.
+ * @param endRevokedStreamsBeforeRebalance
+ *   When `true` (the default) streams that need to end because the partition has been revoked, will be ended before the
+ *   rebalance starts. The consumer that takes over this partition will continue from the committed offset. However, it
+ *   is not possible to commit during a rebalance. So holding up the rebalance until the stream has ended (and done its
+ *   commits) will prevent duplicate processing.
+ *
+ * Set this to `false` when your streams does need commits, or when you need the extra performance and do not care for
+ * duplicate processing.
+ *
+ * Set both `restartStreamOnRebalancing` and `endRevokedStreamsBeforeRebalance` to `true` for transactional producing.
  * @param runloopTimeout
  *   Internal timeout for each iteration of the command processing and polling loop, use to detect stalling. This should
  *   be much larger than the pollTimeout and the time it takes to process chunks of records. If your consumer is not
@@ -31,6 +43,7 @@ case class ConsumerSettings(
   offsetRetrieval: OffsetRetrieval = OffsetRetrieval.Auto(),
   rebalanceListener: RebalanceListener = RebalanceListener.noop,
   restartStreamOnRebalancing: Boolean = false,
+  endRevokedStreamsBeforeRebalance: Boolean = true,
   runloopTimeout: Duration = ConsumerSettings.defaultRunloopTimeout
 ) {
   private[this] def autoOffsetResetConfig: Map[String, String] = offsetRetrieval match {
@@ -85,6 +98,9 @@ case class ConsumerSettings(
 
   def withRestartStreamOnRebalancing(value: Boolean): ConsumerSettings =
     copy(restartStreamOnRebalancing = value)
+
+  def withEndRevokedStreamsBeforeRebalance(value: Boolean): ConsumerSettings =
+    copy(endRevokedStreamsBeforeRebalance = value)
 
   def withCredentials(credentialsStore: KafkaCredentialStore): ConsumerSettings =
     withProperties(credentialsStore.properties)
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/RebalanceListener.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/RebalanceListener.scala
@@ -6,7 +6,7 @@ import zio.{ Runtime, Task, Unsafe, ZIO }
 import scala.jdk.CollectionConverters._
 
 /**
- * ZIO wrapper around Kafka's `ConsumerRebalanceListener` to work with Scala collection types and ZIO effects
+ * ZIO wrapper around Kafka's `ConsumerRebalanceListener` to work with Scala collection types and ZIO effects.
  */
 final case class RebalanceListener(
   onAssigned: (Set[TopicPartition], RebalanceConsumer) => Task[Unit],
@@ -32,21 +32,27 @@ final case class RebalanceListener(
       override def onPartitionsRevoked(
         partitions: java.util.Collection[TopicPartition]
       ): Unit = Unsafe.unsafe { implicit u =>
-        runtime.unsafe.run(onRevoked(partitions.asScala.toSet, consumer)).getOrThrowFiberFailure()
+        runtime.unsafe
+          .run(onRevoked(partitions.asScala.toSet, consumer))
+          .getOrThrowFiberFailure()
         ()
       }
 
       override def onPartitionsAssigned(
         partitions: java.util.Collection[TopicPartition]
       ): Unit = Unsafe.unsafe { implicit u =>
-        runtime.unsafe.run(onAssigned(partitions.asScala.toSet, consumer)).getOrThrowFiberFailure()
+        runtime.unsafe
+          .run(onAssigned(partitions.asScala.toSet, consumer))
+          .getOrThrowFiberFailure()
         ()
       }
 
       override def onPartitionsLost(
         partitions: java.util.Collection[TopicPartition]
       ): Unit = Unsafe.unsafe { implicit u =>
-        runtime.unsafe.run(onLost(partitions.asScala.toSet, consumer)).getOrThrowFiberFailure()
+        runtime.unsafe
+          .run(onLost(partitions.asScala.toSet, consumer))
+          .getOrThrowFiberFailure()
         ()
       }
     }
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/RebalanceListener.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/RebalanceListener.scala
@@ -7,6 +7,9 @@ import scala.jdk.CollectionConverters._
 
 /**
  * ZIO wrapper around Kafka's `ConsumerRebalanceListener` to work with Scala collection types and ZIO effects.
+ *
+ * Note that the given ZIO effects are executed directly on the Kafka poll thread. Fork and shift to another executor
+ * when this is not desired.
  */
 final case class RebalanceListener(
   onAssigned: (Set[TopicPartition], RebalanceConsumer) => Task[Unit],
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/ConsumerAccess.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/ConsumerAccess.scala
@@ -19,7 +19,7 @@ private[consumer] final class ConsumerAccess(
   def withConsumerZIO[R, A](f: ByteArrayKafkaConsumer => RIO[R, A]): RIO[R, A] =
     access.withPermit(withConsumerNoPermit(f))
 
-  private[consumer] def withConsumerNoPermit[R, A](
+  private def withConsumerNoPermit[R, A](
     f: ByteArrayKafkaConsumer => RIO[R, A]
   ): RIO[R, A] =
     ZIO
@@ -31,10 +31,17 @@ private[consumer] final class ConsumerAccess(
       .flatMap(fib => fib.join.onInterrupt(ZIO.succeed(consumer.wakeup()) *> fib.interrupt))
 
   /**
-   * Do not use this method outside of the Runloop
+   * Use this method only from Runloop.
    */
   private[internal] def runloopAccess[R, E, A](f: ByteArrayKafkaConsumer => ZIO[R, E, A]): ZIO[R, E, A] =
     access.withPermit(f(consumer))
+
+  /**
+   * Use this method ONLY from the rebalance handler.
+   */
+  private[internal] def rebalanceHandlerAccess[R, A](f: ByteArrayKafkaConsumer => RIO[R, A]): RIO[R, A] =
+    withConsumerNoPermit(f)
+
 }
 
 private[consumer] object ConsumerAccess {
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/ConsumerAccess.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/ConsumerAccess.scala
@@ -37,9 +37,9 @@ private[consumer] final class ConsumerAccess(
     access.withPermit(f(consumer))
 
   /**
-   * Use this method ONLY from the rebalance handler.
+   * Use this method ONLY from the rebalance listener.
    */
-  private[internal] def rebalanceHandlerAccess[R, A](f: ByteArrayKafkaConsumer => RIO[R, A]): RIO[R, A] =
+  private[internal] def rebalanceListenerAccess[R, A](f: ByteArrayKafkaConsumer => RIO[R, A]): RIO[R, A] =
     withConsumerNoPermit(f)
 
 }
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
@@ -43,10 +43,6 @@ private[internal] final class PartitionStreamControl private (
   def isRunning: ZIO[Any, Nothing, Boolean] =
     isCompleted.negate
 
-  /** Wait till the stream is done. */
-  def awaitCompleted(): ZIO[Any, Nothing, Unit] =
-    completedPromise.await
-
   val tpStream: (TopicPartition, ZStream[Any, Throwable, ByteArrayCommittableRecord]) =
     (tp, stream)
 }
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
@@ -11,8 +11,10 @@ private[internal] final class PartitionStreamControl private (
   val tp: TopicPartition,
   stream: ZStream[Any, Throwable, ByteArrayCommittableRecord],
   dataQueue: Queue[Take[Throwable, ByteArrayCommittableRecord]],
-  interruptPromise: Promise[Throwable, Unit],
-  completedPromise: Promise[Nothing, Unit]
+  startedPromise: Promise[Nothing, Unit],
+  endedPromise: Promise[Nothing, Unit],
+  completedPromise: Promise[Nothing, Unit],
+  interruptPromise: Promise[Throwable, Unit]
 ) {
 
   private val logAnnotate = ZIO.logAnnotate(
@@ -32,16 +34,27 @@ private[internal] final class PartitionStreamControl private (
   def end(): ZIO[Any, Nothing, Unit] =
     logAnnotate {
       ZIO.logTrace(s"Partition ${tp.toString} ending") *>
-        dataQueue.offer(Take.end).unit
+        ZIO
+          .whenZIO(endedPromise.succeed(())) {
+            dataQueue.offer(Take.end)
+          }
+          .unit
     }
 
-  /** Returns true when the stream is done. */
-  def isCompleted: ZIO[Any, Nothing, Boolean] =
-    completedPromise.isDone
+  /** Returns true when the stream accepts new data. */
+  def acceptsData: ZIO[Any, Nothing, Boolean] =
+    for {
+      ended       <- endedPromise.isDone
+      completed   <- completedPromise.isDone
+      interrupted <- interruptPromise.isDone
+    } yield !(ended || completed || interrupted)
 
-  /** Returns true when the stream is running. */
-  def isRunning: ZIO[Any, Nothing, Boolean] =
-    isCompleted.negate
+  /** Returns true when the stream is done (or when it didn't even start). */
+  def isCompletedAfterStart: ZIO[Any, Nothing, Boolean] =
+    for {
+      started   <- startedPromise.isDone
+      completed <- completedPromise.isDone
+    } yield !started || completed
 
   val tpStream: (TopicPartition, ZStream[Any, Throwable, ByteArrayCommittableRecord]) =
     (tp, stream)
@@ -56,8 +69,10 @@ private[internal] object PartitionStreamControl {
   ): ZIO[Any, Nothing, PartitionStreamControl] =
     for {
       _                   <- ZIO.logTrace(s"Creating partition stream ${tp.toString}")
-      interruptionPromise <- Promise.make[Throwable, Unit]
+      startedPromise      <- Promise.make[Nothing, Unit]
+      endedPromise        <- Promise.make[Nothing, Unit]
       completedPromise    <- Promise.make[Nothing, Unit]
+      interruptionPromise <- Promise.make[Throwable, Unit]
       dataQueue           <- Queue.unbounded[Take[Throwable, ByteArrayCommittableRecord]]
       requestAndAwaitData =
         for {
@@ -74,12 +89,21 @@ private[internal] object PartitionStreamControl {
                    completedPromise.succeed(()) <*
                      ZIO.logDebug(s"Partition stream ${tp.toString} has ended")
                  ) *>
+                 ZStream.fromZIO(startedPromise.succeed(())) *>
                  ZStream.repeatZIOChunk {
                    // First try to take all records that are available right now.
                    // When no data is available, request more data and await its arrival.
                    dataQueue.takeAll.flatMap(data => if (data.isEmpty) requestAndAwaitData else ZIO.succeed(data))
                  }.flattenTake
                    .interruptWhen(interruptionPromise)
-    } yield new PartitionStreamControl(tp, stream, dataQueue, interruptionPromise, completedPromise)
+    } yield new PartitionStreamControl(
+      tp,
+      stream,
+      dataQueue,
+      startedPromise,
+      endedPromise,
+      completedPromise,
+      interruptionPromise
+    )
 
 }
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -45,12 +45,14 @@ import scala.jdk.CollectionConverters._
  * Partitions can be assigned, revoked or lost.
  *
  * When a partition is revoked, the stream that handles it will be ended (signal the stream that no more data will be
- * available). Processing however, might continue.
+ * available). Even though there is no more data, the stream still needs to complete processing the data is already got.
+ * Depending on the the `rebalanceSafeStreamEnd` setting, we either hold up the rebalance until the stream is completed
+ * (or not). (See scaladoc in `ConsumerSettings`.)
  *
  * ### Rebalance listener - Commit-loop
  *
- * When `endRevokedStreamsBeforeRebalance` is `true` (the default), we wait for the stream to complete running inside
- * the rebalance listener. This gives the stream a chance to commit offsets before its partition is given to another
+ * When `rebalanceSafeStreamEnd` is `true` (the default), we wait for the stream to complete running inside the
+ * rebalance listener. This gives the stream a chance to commit offsets before its partition is given to another
  * consumer.
  *
  * While the rebalance listener is waiting for streams to complete, we need to continue sending commits. In addition we
@@ -61,7 +63,7 @@ import scala.jdk.CollectionConverters._
  *
  * ## The command-queue and the commit-queue
  *
- * TODO: document more here ...
+ * TODO: Move this document to a central place.
  */
 // Disable zio-intellij's inspection `SimplifyWhenInspection` because its suggestion is not
 // equivalent performance-wise.
@@ -80,7 +82,7 @@ private[consumer] final class Runloop private (
   offsetRetrieval: OffsetRetrieval,
   userRebalanceListener: RebalanceListener,
   restartStreamsOnRebalancing: Boolean,
-  endRevokedStreamsBeforeRebalance: Boolean, // TODO: rename to something like 'completeRevokedStreamsDuringRebalance'
+  rebalanceSafeStreamEnd: Boolean,
   currentState: Ref[State]
 ) {
 
@@ -118,8 +120,8 @@ private[consumer] final class Runloop private (
           state          <- currentState.get
           streamsToEnd = if (restartStreamsOnRebalancing && !rebalanceEvent.wasInvoked) state.assignedStreams
                          else Chunk.empty
-          pendingCommits <- consumer.rebalanceHandlerAccess { consumer =>
-                              endStreams(consumer, streamsToEnd, endRevokedStreamsBeforeRebalance)
+          pendingCommits <- consumer.rebalanceListenerAccess { consumer =>
+                              endStreams(consumer, streamsToEnd, rebalanceSafeStreamEnd)
                             }
           _ <- rebalanceListenerEvent.set(rebalanceEvent.onAssigned(assigned, pendingCommits, streamsToEnd))
           _ <- ZIO.logTrace("onAssigned done")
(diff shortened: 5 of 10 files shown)


erikvanoosten:
I realized that there are some problems with how commits are handled.
1. When there are no rebalances for some time, the commitQueue fills up. It needs to be emptied also during normal operations (fixable).
2. Commits handled in the rebalance listener are not yet added to the state as pendingCommits.

erikvanoosten:
Okay, the problems with the commit queue should be fixed.

svroonland:
Didn't take a very detailed look, but is it possible to just take from the command queue and push back on it whatever we don't use?

zio-kafka/src/main/scala/zio/kafka/consumer/RebalanceListener.scala
```diff
@@ -2,17 +2,22 @@ package zio.kafka.consumer
...
+import zio.internal.ExecutionMetrics
+import zio.{ Executor, Runtime, RuntimeFlag, RuntimeFlags, Task, Trace, Unsafe, ZIO, ZLayer }
+
 import scala.jdk.CollectionConverters._
 
 /**
- * ZIO wrapper around Kafka's `ConsumerRebalanceListener` to work with Scala collection types and ZIO effects
+ * ZIO wrapper around Kafka's `ConsumerRebalanceListener` to work with Scala collection types and ZIO effects.
+ *
+ * Note that the given ZIO effects are executed directly on the Kafka poll thread.
```
svroonland, line 13: Can we instruct the user how to avoid this if not desired?

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -14,6 +14,20 @@ import zio.kafka.security.KafkaCredentialStore
  * @param offsetRetrieval
  * @param rebalanceListener
  * @param restartStreamOnRebalancing
+ *   When `true` _all_ streams are restarted during a rebalance, including those streams that are not revoked. The
+ *   default is `false`.
+ *
+ * Set both `restartStreamOnRebalancing` and `endRevokedStreamsBeforeRebalance` to `true` for transactional producing.
+ * @param endRevokedStreamsBeforeRebalance
+ *   When `true` (the default) streams that need to end because the partition has been revoked, will be ended before the
+ *   rebalance starts. The consumer that takes over this partition will continue from the committed offset. However, it
```
svroonland, line 23: starts -> completes. Revoking is already the start of rebalancing.

Note that for cooperative rebalancing this all doesn't work.

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -14,6 +14,20 @@ import zio.kafka.security.KafkaCredentialStore
...
+ *   is not possible to commit during a rebalance. So holding up the rebalance until the stream has ended (and done its
```
svroonland, line 24: > is not possible to commit during a rebalance
Isn't that what this PR implements..?

zio-kafka/src/main/scala/zio/kafka/consumer/RebalanceConsumer.scala
```diff
@@ -11,11 +11,14 @@ import scala.jdk.CollectionConverters._
  */
 trait RebalanceConsumer {
   def commit(offsets: Map[TopicPartition, OffsetAndMetadata]): Task[Unit]
+  def commitAsync(offsets: Map[TopicPartition, OffsetAndMetadata], callback: OffsetCommitCallback): Task[Unit]
```
svroonland, line 14: Should we have both methods or just keep commitAsync (and name it commit again)?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -22,12 +22,15 @@ private[consumer] final class Runloop private (
   pollTimeout: Duration,
   runloopTimeout: Duration,
   commandQueue: Queue[Command],
-  lastRebalanceEvent: Ref.Synchronized[Option[Runloop.RebalanceEvent]],
+  commitQueue: Queue[Commit],
+  rebalanceListenerInvoked: Ref.Synchronized[Boolean],
```
svroonland, line 26: Can both just be plain Refs.

Synchronized is only needed when you perform effects during the update.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -57,114 +60,147 @@ private[consumer] final class Runloop private (
...
+  private def endRevokedStreamsAndAwaitEnd(
+    rebalanceConsumer: RebalanceConsumer,
+    streamsToEnd: Chunk[PartitionStreamControl]
+  ): Task[Chunk[Commit]] =
+    for {
+      _ <- ZIO.logTrace(s"Ending ${streamsToEnd.size} partitions")
+      _ <- ZIO.foreachDiscard(streamsToEnd)(_.end())
+      pendingCommits <-
+        if (streamsToEnd.nonEmpty && endRevokedStreamsBeforeRebalance) {
+          ZStream
```
svroonland, line 117: Do we need the aggregateAsync? Could it be simpler like:

```scala
ZStream.fromQueue(commitQueue).tapChunk(doCommits.).takeUntilZIO(..).runCollect
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -57,114 +60,147 @@ private[consumer] final class Runloop private (
...
+
+  /** This is the implementation behind the user facing api `Offset.commit`. */
   private val commit: Map[TopicPartition, Long] => Task[Unit] =
     offsets =>
       for {
+        s <- Promise.make[Nothing, Unit]
         p <- Promise.make[Throwable, Unit]
-        _ <- commandQueue.offer(Commit(offsets, p)).unit
+        c = Commit(offsets, s, p)
+        _ <- commandQueue.offer(c).unit
```
svroonland, line 198: Don't you get duplicates this way..?

svroonland (commented):
Will probably need one or two more review passes to understand all the details, but some early feedback

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
erikvanoosten: > Revoking is already the start of rebalancing.

Actually, the rebalance starts once onPartitionRevoke method completes.

zio-kafka/src/main/scala/zio/kafka/consumer/RebalanceListener.scala
erikvanoosten: Interesting. I didn't expect that being something that is desired :)
Definitely possible to support this, though it will require a new setting. Note that using the RebalanceConsumer requires that is called from the calling thread. Ideally, we only execute tasks that use the RebalanceConsumer on the calling thread. But I am not sure this is possible.

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
erikvanoosten: Nope, this PR implements comitting just _before_ the rebalance.

zio-kafka/src/main/scala/zio/kafka/consumer/RebalanceConsumer.scala
erikvanoosten: This method is part of the public API. On the other hand, it can not possibly have worked for anyone because it has never been executed on the thread of callback. I am fine with removing `commit`. Perhaps there are use cases for synchronous commit as well?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Didn't know that. Will change it.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: I am not sure but I think we do. By using aggregate we can combine all outstanding commits into one. This should be a huge performance boost.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: We won't.
BTW, this is my least favorite part of this PR.

Every commit has a new Promise that acts as a gate keeper. Only one task can use the commit, either the poll-loop, or the commit-loop inside the rebalance listener. First come, first to act. The second will ignore already send commits.

erikvanoosten:
There is an alternative to adding the commits to both queues: only add a placeholder command (e.g. CommitAvailable) on the command-queue. This is enough to start the poll-loop where the commit-queue is handled.

This alternative removes the need for checking who executed a commit and should be a bit easier to understand. I'll work on it :)

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
erikvanoosten: BTW, it could be this terminology is not consistently used in the Kafka world. In that we should change the wording here...

erikvanoosten:
Only 1 test failing now: `Consumer Streaming - should close old stream during rebalancing under load`

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: No longer relevant since this code changed completely.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Fixed in https://github.com/zio/zio-kafka/pull/788/commits/e3acebd98395439a2a8a0a6cf3a81e6af7610deb

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: See PR discussion.

erikvanoosten:
I have implemented the idea from https://github.com/zio/zio-kafka/pull/788#issuecomment-1500931734 so that we only add a control value (CommitAvailable) to the command queue.

> is it possible to just take from the command queue and push back on it whatever we don't use?

Yes, I guess that would also be possible. I am a bit worried about ordering though. We could set the unused commands aside and handle them once handlePoll returns. I think we need to experiment with both to see which approach is simpler.

erikvanoosten:
I figured out why the test fails. All partition can be revoked and immediately assigned again. This can not be detected outside the rebalance listener, yet we need to know. Will fix it later.

svroonland:
> I figured out why the test fails. All partition can be revoked and immediately assigned again. This can not be detected outside the rebalance listener, yet we need to know. Will fix it later.
That is the `RevokedAndAssigned` case, isn't it?

erikvanoosten:
> That is the `RevokedAndAssigned` case, isn't it?

Yes, I should not have removed that 😄

erikvanoosten:
Tests are passing now.

Still todo:
- add timeout in the commit-loop inside rebalance listener
- experiment with removing transactionalRebalanceListener from consumerSpec
- run all tests on ConsumerSpec twice: with `endRevokedStreamsBeforeRebalance` true and false
- see if we can support the `CooperativeStickyAssignor`
- experiment with alternative approach (using only command-queue plus buffered commands)

erikvanoosten:
That is not nice. The tests are passing on my laptop, but not in CI...

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -14,6 +14,20 @@ import zio.kafka.security.KafkaCredentialStore
...
+ *   commits) will prevent duplicate processing.
+ *
+ * Set this to `false` when your streams does need commits, or when you need the extra performance and do not care for
+ * duplicate processing.
+ *
+ * Set both `restartStreamOnRebalancing` and `endRevokedStreamsBeforeRebalance` to `true` for transactional producing.
```
guizmaii, line 30: I usually dislike this kind of settings where we need to know that we need to set this parameter to a certain value and that other parameter to another value to the get the desired behaviour.
It often leads to possible invalid binaisons.

Here we have 2 booleans so 4 possibilities. 3 of them are valid but the `false, false` combinaison are no real meaning.

Also, this kind of configuration is not declarative. It doesn't explicit the behaviour that will be activated by the possible combination.

IMO, having 1 parameter which is an ADT or an enum that explicit the possible cases is better.
In our case, instead of having these 2 booleans, we could have something like this:
```scala
sealed trait RebalanceBehaviour
object RebalanceBehaviour {
  case object NoRestart extends RebalanceBehaviour // correspond to `restartStreamOnRebalancing = false`
  case object RestartAllStreams extends RebalanceBehaviour // correspond to `restartStreamOnRebalancing = true, endRevokedStreamsBeforeRebalance = false`
  case object RestartRevokedSteams extends RebalanceBehaviour // correspond to `restartStreamOnRebalancing = true, endRevokedStreamsBeforeRebalance = true`
}

case class ConsumerSettings(
  ....
  rebalanceBehaviour:  RebalanceBehaviour = RebalanceBehaviour.RestartAllStreams
)
```

zio-kafka/src/main/scala/zio/kafka/consumer/RebalanceConsumer.scala
```diff
@@ -11,11 +11,14 @@ import scala.jdk.CollectionConverters._
...
 }
 
 object RebalanceConsumer {
   final case class Live(consumer: JConsumer[Array[Byte], Array[Byte]]) extends RebalanceConsumer {
     def commit(offsets: Map[TopicPartition, OffsetAndMetadata]): Task[Unit] =
       ZIO.attemptBlocking(consumer.commitSync(offsets.asJava))
+    def commitAsync(offsets: Map[TopicPartition, OffsetAndMetadata], callback: OffsetCommitCallback): Task[Unit] =
+      ZIO.attempt(consumer.commitAsync(offsets.asJava, callback))
```
guizmaii, line 22: This could be misleading to our users, DYT?
Because the `Task[Unit]` doesn't wait for the `commitAsync` to be done to return. 
Shouldn't we use a callback to incorporate the `commitAsync` into the returned `Task`? Something like this:
```scala
    def commitAsync(offsets: Map[TopicPartition, OffsetAndMetadata], callback: OffsetCommitCallback): Task[Unit] =
      ZIO.async(cb =>
        consumer.commitAsync(
          offsets.asJava,
          (map: util.Map[TopicPartition, OffsetAndMetadata], e: Exception) =>
            if (e eq null) cb(ZIO.attempt(callback.onComplete(map, e)))
            else cb(ZIO.fail(e))
        )
      )
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -32,17 +33,25 @@ private[internal] final class PartitionStreamControl private (
...
+          .unit
     }
 
+  /** Returns true when the stream accepts new data. */
+  def acceptsData: ZIO[Any, Nothing, Boolean] =
+    for {
+      ended       <- endedPromise.isDone
+      completed   <- completedPromise.isDone
+      interrupted <- interruptPromise.isDone
+    } yield !ended && !completed && !interrupted
```
guizmaii, line 49:
```suggestion
    } yield !(ended || completed || interrupted)
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/package.scala
```diff
@@ -0,0 +1,28 @@
+package zio.kafka.consumer
+
+import zio._
+import zio.internal.ExecutionMetrics
+
+package object internal {
+
+  private[internal] val SameThreadRuntimeLayer: ZLayer[Any, Nothing, Unit] = {
```
guizmaii, line 8: Can we add a comment here to explicit that this code was provided by Adam in Discord: https://discord.com/channels/629491597070827530/630498701860929559/1094279123880386590

That'd avoid people spending time trying to see if this is correct

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -505,46 +506,55 @@ private[consumer] object Runloop {
...
+    def apply(records: ConsumerRecords[Array[Byte], Array[Byte]]): PollResult =
+      PollResult(Chunk.empty, Set.empty, records, Set.empty)
+  }
+
   private final case class FulfillResult(
     pendingRequests: Chunk[Request]
   )
 
-  private sealed trait RebalanceEvent
+  private case class RebalanceEvent(
```
guizmaii, line 527:
```suggestion
  private final case class RebalanceEvent(
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -505,46 +506,55 @@ private[consumer] object Runloop {
...
     records: ConsumerRecords[Array[Byte], Array[Byte]],
     ignoreRecordsForTps: Set[TopicPartition]
   )
-  private final case class RevokeResult(
-    pendingRequests: Chunk[Request],
-    assignedStreams: Chunk[PartitionStreamControl]
-  )
+  private object PollResult {
+    def apply(records: ConsumerRecords[Array[Byte], Array[Byte]]): PollResult =
+      PollResult(Chunk.empty, Set.empty, records, Set.empty)
```
guizmaii, line 520:
```suggestion
            PollResult(
        newCommits = Chunk.empty,
        startingTps = Set.empty,
        records = records,
        ignoreRecordsForTps = Set.empty
      )
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -57,113 +62,152 @@ private[consumer] final class Runloop private (
...
+   */
+  private def doCommitsFromRebalanceListener(
+    rebalanceConsumer: RebalanceConsumer
+  )(commits: Chunk[Commit]): Task[Unit] = {
+    val (offsets, callback, onFailure) = asyncCommitParameters(commits)
+    ZIO.logTrace(s"Async commit of ${offsets.size} offsets for ${commits.size} commits") *>
+      rebalanceConsumer
+        .commitAsync(offsets, callback)
+        .catchAll(onFailure)
+        .unit
```
guizmaii, line 152: It's already returning `Unit`
```suggestion
      rebalanceConsumer
        .commitAsync(offsets, callback)
        .catchAll(onFailure)
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -57,113 +62,152 @@ private[consumer] final class Runloop private (
...
+   * Handle commits while waiting for revoked streams to end.
+   *
+   * We need to ensure the revoked streams can end. This is only possible if the commits these streams started complete.
+   * The commits complete when the callback is invoked. The callback is invoked when the underlying consumer polls the
+   * broker. This can be achieved by invoking `commitAsync`. Even when we pass no offsets, the broker will be polled and
+   * callbacks will be called.
+   */
+  private def doCommitsFromRebalanceListener(
+    rebalanceConsumer: RebalanceConsumer
+  )(commits: Chunk[Commit]): Task[Unit] = {
```
guizmaii, line 146:
```suggestion
  )(commits: Chunk[Commit]): UIO[Unit] = {
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -57,113 +62,152 @@ private[consumer] final class Runloop private (
...
+  }
+
+  private def asyncCommitParameters(
+    commits: Chunk[Commit]
+  ): (Map[TopicPartition, OffsetAndMetadata], OffsetCommitCallback, Throwable => UIO[Unit]) = {
+    val offsets = commits.foldLeft(Map.empty[TopicPartition, Long]) { case (acc, commit) =>
+      commit.offsets.foldLeft(acc) { case (acc, (tp, offset)) =>
+        acc.updated(tp, acc.get(tp).map(_ max offset).getOrElse(offset))
+      }
+    }
```
guizmaii, line 174:
```suggestion
    val offsets =
      commits
        .foldLeft(mutable.Map.empty[TopicPartition, Long]) { case (acc, commit) =>
          commit.offsets.foreach { case (tp, offset) =>
            acc += (tp, acc.get(tp).fold(offset)(_ max offset))
          }
          acc
        }
        .toMap
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -57,113 +62,152 @@ private[consumer] final class Runloop private (
...
+      .catchAll(onFailure)
+      .unit
+      .as(state.addCommits(commits))
+  }
+
+  private def asyncCommitParameters(
+    commits: Chunk[Commit]
+  ): (Map[TopicPartition, OffsetAndMetadata], OffsetCommitCallback, Throwable => UIO[Unit]) = {
+    val offsets = commits.foldLeft(Map.empty[TopicPartition, Long]) { case (acc, commit) =>
```
guizmaii, line 170: This value is only useful when the diagnostic is enable, so most probably useless in prod for our users
``` suggestion
    def offsets = commits.foldLeft(Map.empty[TopicPartition, Long]) { case (acc, commit) =>
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -57,113 +62,152 @@ private[consumer] final class Runloop private (
...
+      commit.offsets.foldLeft(acc) { case (acc, (tp, offset)) =>
+        acc.updated(tp, acc.get(tp).map(_ max offset).getOrElse(offset))
+      }
+    }
+    val offsetsWithMetaData = offsets.map { case (tp, offset) => tp -> new OffsetAndMetadata(offset + 1) }
```
guizmaii, line 175: This value is only useful when the diagnostic is enabled, so most probably useless in prod for our users
```suggestion
    def offsetsWithMetaData = offsets.map { case (tp, offset) => tp -> new OffsetAndMetadata(offset + 1) }
```

guizmaii, line 175:
```suggestion
    val offsetsWithMetaData = offsets.view.mapValues(offset => new OffsetAndMetadata(offset + 1)).toMap
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -243,90 +286,60 @@ private[consumer] final class Runloop private (
...
   }
 
   private def handlePoll(state: State): Task[State] =
     for {
+      _ <-
+        ZIO.logTrace(
+          s"Starting poll with ${state.pendingRequests.size} pending requests and ${state.pendingCommits.size} pending commits"
+        )
       _ <- currentState.set(state)
+      _ <- rebalanceListenerEvent.set(RebalanceEvent.None)
```
guizmaii, line 302: Why do we reset the `rebalanceListenerEvent` each time we call `handlePoll`? 🤔

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -57,113 +62,152 @@ private[consumer] final class Runloop private (
...
+  private def handleCommits(state: State, commits: Chunk[Commit]): Task[State] = {
+    val (offsets, callback, onFailure) = asyncCommitParameters(commits)
+    consumer.withConsumerZIO { c =>
+      // We don't wait for the completion of the commit here, because it
+      // will only complete once we poll again.
+      ZIO.attempt(c.commitAsync(offsets.asJava, callback))
+    }
+      .catchAll(onFailure)
+      .unit
+      .as(state.addCommits(commits))
```
guizmaii, line 164:
```suggestion
    consumer.withConsumerZIO { c =>
      // We don't wait for the completion of the commit here, because it
      // will only complete once we poll again.
      ZIO.attempt(c.commitAsync(offsets.asJava, callback))
    }
      .catchAll(onFailure)
      .as(state.addCommits(commits))
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -57,113 +62,152 @@ private[consumer] final class Runloop private (
...
+        .commitAsync(offsets, callback)
+        .catchAll(onFailure)
+        .unit
+  }
 
-  private def doCommit(cmd: Commit): UIO[Unit] = {
-    val offsets   = cmd.offsets.map { case (tp, offset) => tp -> new OffsetAndMetadata(offset + 1) }
-    val cont      = (e: Exit[Throwable, Unit]) => cmd.cont.done(e).asInstanceOf[UIO[Unit]]
-    val onSuccess = cont(Exit.unit) <* diagnostics.emitIfEnabled(DiagnosticEvent.Commit.Success(offsets))
+  private def handleCommits(state: State, commits: Chunk[Commit]): Task[State] = {
```
guizmaii, line 155:
```suggestion
  private def handleCommits(state: State, commits: Chunk[Commit]): UIO[State] = {
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -57,113 +62,152 @@ private[consumer] final class Runloop private (
...
+    val (offsets, callback, onFailure) = asyncCommitParameters(commits)
+    consumer.withConsumerZIO { c =>
+      // We don't wait for the completion of the commit here, because it
+      // will only complete once we poll again.
+      ZIO.attempt(c.commitAsync(offsets.asJava, callback))
+    }
+      .catchAll(onFailure)
+      .unit
+      .as(state.addCommits(commits))
+  }
```
guizmaii, line 165: private def handleCommits(state: State, commits: Chunk[Commit]): Task[State] = {
    val (offsets, callback, onFailure) = asyncCommitParameters(commits)
    val newState                       = state.addCommits(commits)
    consumer.withConsumerZIO { c =>
      ZIO.attempt(c.commitAsync(offsets.asJava, callback))
    }
      .catchAll(onFailure)
      .as(newState)
  }

guizmaii, line 165:
```suggestion
  private def handleCommits(state: State, commits: Chunk[Commit]): Task[State] = {
    val (offsets, callback, onFailure) = asyncCommitParameters(commits)
    val newState                       = state.addCommits(commits)
    consumer.withConsumerZIO { c =>
      // We don't wait for the completion of the commit here, because it
      // will only complete once we poll again.
      ZIO.attempt(c.commitAsync(offsets.asJava, callback))
    }
      .catchAll(onFailure)
      .as(newState)
  }
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -448,37 +454,32 @@ private[consumer] final class Runloop private (
...
       .fromQueue(commandQueue)
       .timeoutFail[Throwable](RunloopTimeout)(runloopTimeout)
       .takeWhile(_ != StopRunloop)
       .runFoldChunksDiscardZIO(State.initial) { (state, commands) =>
         for {
-          _            <- ZIO.logTrace(s"Processing ${commands.size} commands: ${commands.mkString(",")}")
-          updatedState <- ZIO.foldLeft(commands)(state)(handleCommand)
+          commits <- commitQueue.takeAll
+          _ <- ZIO.logTrace(s"Processing ${commits.size} commits, ${commands.size} commands: ${commands.mkString(",")}")
+          stateAfterCommits  <- handleCommits(state, commits)
```
guizmaii, line 472: Let's avoid triggering all the `handleCommits` if there are no commits
```suggestion
          stateAfterCommits  <- if (commits.isEmpty) ZIO.succeed(state) else handleCommits(state, commits)
```

guizmaii:
I added a few comments but still need to finish my review.

I'm wondering why you need to add this new `CommitAvailable` command, but you never manage it in the run loop. 🤔

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
erikvanoosten: I like your thinking.

> Here we have 2 booleans, so 4 possibilities. Are they all valid? Are they all making sense?

They are all valid, but perhaps not all making sense.

| `restartStream-OnRebalancing` | `endRevokedStreams-BeforeRebalance` | Description |
| --- | --- | --- |
| `false` | `false` | The most performant option, but rebalances cause duplicates. |
| `false` | `true` | Should be the default option, no more duplicates, but less performance during rebalance. |
| `true` | `false` | Was introduced to make transactional producing possible. Still useful for backward compatibility. |
| `true` | `true` | Should be the default for transactional producing. Note, we still need more work to make easy transactional producing possible. |

We also need to think about whether we want to support so many combinations, perhaps 1 is enough! IMHO zio-kafka should be an allround zio library so my answer would be yes (as long as there are sensible defaults).

zio-kafka/src/main/scala/zio/kafka/consumer/RebalanceConsumer.scala
erikvanoosten: Yes, that is a good point. Actually, I am wondering if we should expose these methods at all. The stream can do the committing, what use case is there for the user to do commits in the rebalance listener?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: I have read that views are usually slower than direct operations. That's why I went for this approach.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: No this is not good, in fact this is explicitly necessary. By doing this we enforce a poll to the broker so that previously send commits can complete. Please also see the comments around this code.

erikvanoosten:
`SubscriptionsSpec - can resume a stream for the same subscription` fails on CI but not on my laptop. Any ideas? On CI the rebalance listener is waiting for the streams to end.

Could this be because `val hubCapacity = 32` in Consumer (https://github.com/zio/zio-kafka/blob/master/zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala#L354..L362) is too low?

zio-kafka/src/main/scala/zio/kafka/consumer/RebalanceListener.scala
erikvanoosten: Added the text "Fork and shift to another executor when this is not desired.".

erikvanoosten:
> `SubscriptionsSpec - can resume a stream for the same subscription` fails on CI but not on my laptop. Any ideas? On CI the rebalance listener is waiting for the streams to end.

Turns out that in this test streams do not even get a chance to start. This means that will never be able to end either. To fix this, in https://github.com/zio/zio-kafka/pull/788/commits/daa928619c1a53a15f968080c541b80b32374c4a we detect stream start as well.

erikvanoosten:
Tests are now also passing in CI 😄 

Still todo:

* add timeout in the commit-loop inside rebalance listener
* run all tests on ConsumerSpec and SubscriptionSpec twice: with endRevokedStreamsBeforeRebalance true and false
* experiment with alternative approach (using only command-queue plus buffered commands)

For a later PR:
 * configure zio-kafka via use case profile
 * experiment with removing transactionalRebalanceListener from consumerSpec
 * see if we can support the CooperativeStickyAssignor

guizmaii:
FYI,

### Before

```scala
[info] Benchmark                                  Mode  Cnt     Score      Error  Units
[info] ConsumersComparisonBenchmark.kafkaClients  avgt   20   653.895 ±   15.343  ms/op
[info] ConsumersComparisonBenchmark.zioKafka      avgt   20  2029.388 ± 1029.933  ms/op
```

### After

```scala
[info] ConsumersComparisonBenchmark.kafkaClients  avgt   20   647.766 ±   5.359  ms/op
[info] ConsumersComparisonBenchmark.zioKafka      avgt   20  1045.803 ± 235.863  ms/op
```

🎉

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -14,6 +14,28 @@ import zio.kafka.security.KafkaCredentialStore
...
+ *
+ * When `false`, streams for revoked partitions may continue to run to completion but the rebalance is not hold up. Any
+ * offset commits from such a stream have a high chance of being delayed (commits are not possible during some phases of
+ * a rebalance). The consumer that takes over the partition will likely not see these delayed commits and will start
+ * from an earlier offset. The result is that some messages are processed twice and concurrently.
+ *
+ * Use `false` _only_ when your streams does not do commits, or when it is okay to have messages processed twice
+ * concurrently and you cannot afford the performance hit during a rebalance.
+ *
+ * Set both `restartStreamOnRebalancing` and `rebalanceSafeStreamEnd` to `true` for transactional producing.
```
guizmaii, line 38: 🥇 Great documentation! ❤️

zio-kafka/src/main/scala/zio/kafka/consumer/internal/ConsumerAccess.scala
```diff
@@ -31,10 +31,17 @@ private[consumer] final class ConsumerAccess(
...
+   * Use this method only from Runloop.
    */
   private[internal] def runloopAccess[R, E, A](f: ByteArrayKafkaConsumer => ZIO[R, E, A]): ZIO[R, E, A] =
     access.withPermit(f(consumer))
+
+  /**
+   * Use this method ONLY from the rebalance handler.
+   */
+  private[internal] def rebalanceHandlerAccess[R, A](f: ByteArrayKafkaConsumer => RIO[R, A]): RIO[R, A] =
+    withConsumerNoPermit(f)
```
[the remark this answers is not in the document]
erikvanoosten: Yes we can. I did it this way for two reasons:
- it is the minimal change
- to indicate that this method is really only for use from the rebalance listener

Oops, I realize now that the correct name is `rebalanceListenerAccess` 😄

erikvanoosten:
> FYI,
> ### Before
> 
> ```scala
> [info] Benchmark                                  Mode  Cnt     Score      Error  Units
> [info] ConsumersComparisonBenchmark.kafkaClients  avgt   20   653.895 ±   15.343  ms/op
> [info] ConsumersComparisonBenchmark.zioKafka      avgt   20  2029.388 ± 1029.933  ms/op
> ```
> 
> ### After
> 
> ```scala
> [info] Benchmark                                  Mode  Cnt     Score     Error  Units
> [info] ConsumersComparisonBenchmark.kafkaClients  avgt   20   647.766 ±   5.359  ms/op
> [info] ConsumersComparisonBenchmark.zioKafka      avgt   20  1045.803 ± 235.863  ms/op
> ```

Wait, what??
This was not intended as a performance PR at all. This is intended to increase reliability and correctness, not performance. Are you sure you measured the correct PR??

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -7,27 +7,82 @@ import zio._
...
   commandQueue: Queue[Command],
-  lastRebalanceEvent: Ref.Synchronized[Option[Runloop.RebalanceEvent]],
+  commitQueue: Queue[Commit],
+  rebalanceListenerEvent: Ref[RebalanceEvent],
   val partitions: Queue[Take[Throwable, (TopicPartition, Stream[Throwable, ByteArrayCommittableRecord])]],
   diagnostics: Diagnostics,
   offsetRetrieval: OffsetRetrieval,
   userRebalanceListener: RebalanceListener,
   restartStreamsOnRebalancing: Boolean,
+  rebalanceSafeStreamEnd: Boolean, // TODO: rename to something like 'completeRevokedStreamsDuringRebalance'
```
guizmaii, line 85: Do we want to rename it now?

erikvanoosten: Thanks for spotting that.
It already has been renamed. The TODO needs to be removed.

erikvanoosten: Fixed.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/ConsumerAccess.scala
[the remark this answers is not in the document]
erikvanoosten: I changed the name.

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
erikvanoosten: Thanks.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
guizmaii: > By doing this we enforce a poll to the broker so that previously send commits can complete.

I'm not sure to see where this is enforced 🤔

If we look at `handleCommits`,

First, it computes these variables:
```
(offsets, callback, onFailure) = asyncCommitParameters(commits)
```
where:
 - `offsets` will be an empty list
 - `callback` will have a `onSuccess` that does nothing and a `onFailure` that requeues an empty list of commits
 - `onFailure` is the same as the one in the callback
 
Then, it computes the new state by adding these commits to the state, but because the commits is empty, we just re-allocating a new state with the same data inside.

Finally, it calls `Consumer::commitAsync` with an empty list of commits.
If we dig into the code of the `KafkaConsumer`, it seems that calling `commitAsync` will do nothing else than triggering our `callback.onSuccess`, which is a noOp:
Step 1:
Screenshot 2023-04-17 at 12 51 04 pm
Step 2:
Screenshot 2023-04-17 at 12 51 23 pm
Step 3:
Screenshot 2023-04-17 at 12 51 34 pm
Step 4:
Screenshot 2023-04-17 at 12 53 47 pm


So, I'm still unsure why we need to call `handleCommits` with an empty list of commits.

🤔

guizmaii: Hum ok I think I see.
It's because of this `client.pollNoWakeup` call in `commitOffsetsAsync`:
image
?

🤔

svroonland (commented):
Some preliminary comments. Still need to look into the core parts of this change

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Yeah, it took me long time as well. But indeed, yes, `client.pollNoWakeup` is the place where the broker is polled and callbacks are invoked.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/package.scala
```diff
@@ -0,0 +1,28 @@
...
+import zio._
+import zio.internal.ExecutionMetrics
+
+package object internal {
+
+  /**
+   * A runtime layer that can be used to run everything on the thread of the caller.
+   *
+   * Provided by Adam Fraser in Discord:
+   * https://discord.com/channels/629491597070827530/630498701860929559/1094279123880386590 but with cooperative
```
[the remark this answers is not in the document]
erikvanoosten: Currently we don't need it as we only have a single fiber in the code called from the rebalance listener. As soon as there are multiple fibers (that still need to run on the same single thread), we need the fibers to yield so that the other fibers get a chance to run.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -57,114 +105,175 @@ private[consumer] final class Runloop private (
...
-            }
-          }
+        for {
+          _              <- ZIO.logDebug(s"${assigned.size} partitions are assigned")
+          rebalanceEvent <- rebalanceListenerEvent.get
+          state          <- currentState.get
+          streamsToEnd = if (restartStreamsOnRebalancing && !rebalanceEvent.wasInvoked) state.assignedStreams
+                         else Chunk.empty
+          pendingCommits <- consumer.rebalanceListenerAccess { consumer =>
+                              endStreams(consumer, streamsToEnd, rebalanceSafeStreamEnd)
```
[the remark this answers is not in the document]
erikvanoosten: The 3rd parameter of endStreams is called `awaitEnd` and can be `true` or `false` independent of the value of `rebalanceSafeStreamEnd`.

erikvanoosten:
There is a problem with this PR: stream end is harder to detect than I thought. When streams are merged (e.g. with `ZStream.mergeAllUnbounded`), the underlying stream is ended but the merged stream is still processing data. Since it processing data, it is _not done_.

Any ideas for more robust detection that all messages up to some point have been processed?

erikvanoosten:
Closing, this PR is superseded by #803.

PR #791, opened 2023-04-10
Optimize `offerRecordsToStreams` code

guizmaii:
The goal here is to:
 1. make all the iterations to compute the `committableRecords` into 1 `IO` instead of `n` (being the size of the `streams`) 
 2. to allocate the ChunkBuilders only when necessary
 3. Do the minimum number of iterations in the `ZIO.foreachDiscard`

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -177,6 +177,8 @@ private[consumer] final class Runloop private (
     ignoreRecordsForTps: Set[TopicPartition],
     polledRecords: ConsumerRecords[Array[Byte], Array[Byte]]
   ): UIO[Runloop.FulfillResult] = {
+    type Record = CommittableRecord[Array[Byte], Array[Byte]]
+
     // The most efficient way to get the records from [[ConsumerRecords]] per
     // topic-partition, is by first getting the set of topic-partitions, and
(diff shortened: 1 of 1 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -189,26 +191,33 @@ private[consumer] final class Runloop private (
     else {
       for {
         consumerGroupMetadata <- getConsumerGroupMetadataIfAny
+        committableRecords = {
+          val acc             = ChunkBuilder.make[(PartitionStreamControl, Chunk[Record])](streams.size)
+          val streamsIterator = streams.iterator
+          while (streamsIterator.hasNext) {
+            val streamControl = streamsIterator.next()
```
erikvanoosten, line 198:
```suggestion
          streams.foreach { streamControl =>
```

erikvanoosten (approved):
Smart! This optimizes for the normal case where a poll returns records for a single or low number of partitions.

With this change the number of ZIOs is o(number of partitions in poll) instead of o(number of streams).

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
guizmaii: I wrote it with an iterator on purpose. It gives a better understanding of things when profiling code

Issue #800, opened 2023-04-12
Optimistic prefetching


svroonland:
Zio-kafka pauses and resumes partitions for backpressure purposes, when messages cannot be processed in the user stream at the speed they are retrieved in the poll calls.

When uncommenting the `pause()` calls, we see a significant difference in the `ConsumersComparison` benchmark.

With pausing:
```
[info] Benchmark                              Mode  Cnt     Score     Error  Units
[info] ConsumersComparisonBenchmark.zioKafka  avgt   25  2151,998 ± 748,781  ms/op
```

Without pausing:
```
[info] Benchmark                              Mode  Cnt    Score    Error  Units
[info] ConsumersComparisonBenchmark.zioKafka  avgt   25  925,993 ± 83,150  ms/op
```

So there is potential to improve throughput by having a better backpressure mechanism.

Idea from @erikvanoosten from #706 :
> A partition can be paused and un-paused every 2 cycles of the poll loop. When this observed twice, we should automatically delay the pause by 1 poll loop.

svroonland:
Implemented in #809

PR #802, opened 2023-04-12
Run the Runloop on a dedicated single-threaded thread pool

guizmaii:

PR #803, opened 2023-04-12
Improve prefetch

svroonland:
See #800

This PR improves prefetching by fixing a race condition between the fulfilling of Requests and enqueueing new ones in time for the next poll. 

Instead of waiting for incoming Requests, we now continue polling for a partition until it's queue size exceeds some threshold, in which case we pause it. Replaces the `perPartitionChunkPrefetch` setting, which was always hard to translate into an estimated nr of records, with a `maxPartitionQueueSize` setting which is a number of records.

Master:
```
[info] ConsumersComparisonBenchmark.kafkaClients  avgt   25   657,527 ±   7,726  ms/op
[info] ConsumersComparisonBenchmark.zioKafka      avgt   25  2144,102 ± 755,787  ms/op
```

This branch:

```
[info] Benchmark                                  Mode  Cnt    Score    Error  Units
[info] ConsumersComparisonBenchmark.kafkaClients  avgt   25  659,675 ± 10,947  ms/op
[info] ConsumersComparisonBenchmark.zioKafka      avgt   25  821,853 ± 73,123  ms/op
```

* [x] Think of recommendation for value of new setting
* [x] Remove Request mechanism
* [x] Effect on duplicate message processing when rebalancing..?


svroonland:
With #802 I'm getting this:

```
[info] Benchmark                              Mode  Cnt    Score    Error  Units
[info] ConsumersComparisonBenchmark.zioKafka  avgt   25  794,420 ± 85,576  ms/op
```

svroonland:
I reverted the removal of Requests. They are needed for the case where the consumer fills all partition streams' queues above the threshold and before they are drained it looks at the queue sizes to see if it should poll again and finds that all of them are 'full', so won't poll again, resulting in a stuck runloop.

svroonland:
@erikvanoosten Regarding the effect of this extra buffering on duplicate messages when rebalancing: do you think that if we can wait for streams to finish in `onRevoked` it doesn't matter what size the buffer is (within reason)?

svroonland:
@erikvanoosten @guizmaii 
Regarding a recommendation for this new setting, here's my thoughts which hopefully also capture our discussion about it yesterday:

* Very large values of this setting effectively disable backpressure but will lead to unbounded memory usage. 
* A value of zero will enforce maximum backpressure: all messages need to have been dequeued before a next batch will be polled. This will lead to bad throughput.
* On every poll we expect approximately `max.poll.records / nr of partitions` records to be placed on each partition's queue, but the actual value will be anywhere between 0 and `max.poll.records`. 
* If the value is less than `max.poll.records`, the next poll will often pause the partition when the partition stream hasn't yet dequeued the last offered values. This is a race condition between offering the records to the queue, the partition stream's dequeueing and checking the queue sizes for the next poll. 
* The larger the value, the more messages will be processed duplicately when rebalancing, at least until we get #788 to work.
* If you have large variance in your downstream message processing rate, to avoid starvation it's best to have a value that is several times `max.poll.records`. The more buffer memory you can spare the less chance of stravation.

Our default setting should balance throughput and memory usage somewhat, so perhaps 2 times `max.poll.records`. Users are free to decrease or increase according to their needs and the above information.

erikvanoosten:
> @erikvanoosten Regarding the effect of this extra buffering on duplicate messages when rebalancing: do you think that if we can wait for streams to finish in `onRevoked` it doesn't matter what size the buffer is (within reason)?

Yes, that is my reasoning as well.

> Regarding a recommendation for this new setting, here's my thoughts which hopefully also capture our discussion about it yesterday:

One thing that is missing in this analysis is that we shouldn't look at how much work the streams can process separately, but how much they can process together. This is necessary because we typically have a limited number of CPUs that can be used by the streams. For example, on a quad-core machine, even with 100 streams, only 4 are effectively progressing concurrently. In short: it doesn't make sense to buffer for a stream that is waiting on CPU anyway.

svroonland:
> One thing that is missing in this analysis is that we shouldn't look at how much work the streams can process separately, but how much they can process together. This is necessary because we typically have a limited number of CPUs that can be used by the streams. For example, on a quad-core machine, even with 100 streams, only 4 are effectively progressing concurrently.

I'm not sure what that implies for the recommended values for the new setting. Do you have a suggestion? The important thing IMO is what we find some value that does not cause unnecessary delays, so it would rather be too large than too small.

> ..it doesn't make sense to buffer for a stream that is waiting on CPU anyway.

The way I see it always makes sense to buffer a bit in zio-kafka, because the moment that a stream has finished processing and it signals that it wants more data, there is unneeded latency until the next records are fetched.

erikvanoosten:
> so it would rather be too large than too small.

Yeah, that is my problem with this approach. It will either use more memory than needed, or there is no prefetching. Even worse, we are going to give the responsibility to configure this to the user!

> The way I see it always makes sense to buffer a bit in zio-kafka, because the moment that a stream has finished processing and it signals that it wants more data, there is unneeded latency until the next records are fetched.

I know. I think the trick is to fetch more data _just_ before it is needed. I suspect that the only way to achieve that is with auto-tuning. So runtime prediction of when data is needed and then fetch data just a bit earlier. (This is the underlying idea of #809.)

Auto-tuning should be possible with the approach of this PR as well!

svroonland:
@guizmaii Done

svroonland:
Closed for #809

svroonland:
@guizmaii @erikvanoosten Okay to merge?

erikvanoosten (approved):

PR #804, opened 2023-04-12
Remove `Set` instance from the Runloop

guizmaii:


svroonland (approved):

erikvanoosten:
Was it not merged because it is not possible, or because we didn't find the right approach yet?

PR #806, opened 2023-04-12
Fix #799

### LINKED ISSUE #799: Consumer block the app from shutting down
After updating ZIO from 2.0.10 to 2.0.11, we noticed that consumer stream cannot be interrupted when the process receives SIGTERM or SIGINT, so the app refused to be shutdown normally. This bug still happens with the latest commit from master branch.

Example code:

```scala
object UserApp extends ZIOAppDefault {

    val stream = Consumer.plainStream(Subscription.topics("test"), Serde.byteArray.asTry, Serde.byteArray.asTry)
    val consumerLayer = {
      val consumerSettings =
        ConsumerSettings(List("localhost:9092")).withGroupId("test")
      val dianostic = Diagnostics.NoOp
      (ZLayer.succeed(consumerSettings) ++ ZLayer.succeed(dianostic)) >>> Consumer.live
    }
    stream.runDrain.provideLayer(consumerLayer)
  }

}
```

guizmaii:
The `fib.join` is probably creating a deadlock preventing the finalizer made by the `forkScoped` to be be called

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -473,7 +473,10 @@ private[consumer] final class Runloop private (
     ZStream
       .fromQueue(commandQueue)
       .timeoutFail[Throwable](RunloopTimeout)(runloopTimeout)
-      .takeWhile(_ != StopRunloop)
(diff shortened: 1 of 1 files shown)


guizmaii:
The issue comes from ZIO. See #807

guizmaii:
FYI https://github.com/zio/zio/issues/8016

PR #807, opened 2023-04-12
Add a minimal example app, useful to tests things manually if necessary

### LINKED ISSUE #799: Consumer block the app from shutting down

guizmaii:
Reproducer for #799 

The issue comes from ZIO 2.0.11.

To reproduce here are the steps:

# First run the app with 2.0.10

To run the app
```bash
sbt "clean;zioKafkaExample/stage"
./zio-kafka-example/target/universal/stage/bin/zio-kafka-example
```

Then when the Consumer has started, hit CTRL+C to interrupt the app. 
The app will be interrupted

# Then run the app with 2.0.11

In the `build.sbt` file, replace the line 149: `"dev.zio" %% "zio" % "2.0.10",` with `"dev.zio" %% "zio" % "2.0.11",` and run the app again:
```bash
sbt "clean;zioKafkaExample/stage"
./zio-kafka-example/target/universal/stage/bin/zio-kafka-example
```

Then when the Consumer has started, hit CTRL+C to interrupt the app. 
The app will be **never** interrupted

----- 
_Edit:_ ZIO 2.0.13 is fixing #799

### DIFF
--- a/zio-kafka-example/src/main/scala/zio/kafka/example/Main.scala
+++ b/zio-kafka-example/src/main/scala/zio/kafka/example/Main.scala
@@ -1,11 +1,8 @@
 package zio.kafka.example
 
-import io.github.embeddedkafka.EmbeddedKafka
 import zio._
-import zio.kafka.KafkaTestUtils.{ produceMany, producer }
 import zio.kafka.consumer.diagnostics.Diagnostics
 import zio.kafka.consumer.{ Consumer, ConsumerSettings, Subscription }
-import zio.kafka.embedded.Kafka
 import zio.kafka.serde.Serde
 import zio.logging.backend.SLF4J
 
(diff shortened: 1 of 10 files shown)


svroonland (approved):

zio-kafka-example/src/main/scala/zio/kafka/example/Main.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  /**
+   * See `zio-logging` documentation: https://zio.github.io/zio-logging/docs/overview/overview_slf4j
+   */
+  override val bootstrap: ZLayer[ZIOAppArgs, Any, Any] =
+    zio.Runtime.removeDefaultLoggers >>> SLF4J.slf4j
+
+  private val topic = "test-topic"
+
+  private def consumerLayer(kafka: Kafka): ZLayer[Any, Throwable, Consumer] = {
```
erikvanoosten, line 44: Can we move the `Kafka` dependency to the environment?

erikvanoosten (commented):
Good idea!
Just 1 small comment.

PR #809, opened 2023-04-13
Optimistically resume partitions early

erikvanoosten:
Optimistically resume partitions that are likely to need more data before the next poll.

Solves #800.

Benchmark reports:

master:
```
[info] ConsumerBenchmark.throughput               avgt   25   465.958 ± 100.053  ms/op
[info] ConsumerBenchmark.throughputWithCommits    avgt   25   433.205 ±  85.996  ms/op
[info] ConsumersComparisonBenchmark.kafkaClients  avgt   25   664.730 ±  12.088  ms/op
[info] ConsumersComparisonBenchmark.zioKafka      avgt   25  1999.502 ± 717.557  ms/op
```

This branch (commit https://github.com/zio/zio-kafka/pull/809/commits/379b490c8144e077cbac7361cbb6e176af5feae5):
```
[info] ConsumerBenchmark.throughput               avgt   25   226.624 ±  19.807  ms/op
[info] ConsumerBenchmark.throughputWithCommits    avgt   25   316.707 ±  29.479  ms/op
[info] ConsumersComparisonBenchmark.kafkaClients  avgt   25   654.995 ±  10.792  ms/op
[info] ConsumersComparisonBenchmark.zioKafka      avgt   25  1132.004 ± 174.654  ms/op
```

1.8 times faster!

After removing pre-fetching (commit https://github.com/zio/zio-kafka/pull/809/commits/ca4e88cb536026b3ddd88afa9810481b89112fc5):

```
[info] ConsumerBenchmark.throughput               avgt   25   298.363 ±  57.858  ms/op
[info] ConsumerBenchmark.throughputWithCommits    avgt   25   300.736 ±  23.548  ms/op
[info] ConsumersComparisonBenchmark.kafkaClients  avgt   25   654.316 ±  15.199  ms/op
[info] ConsumersComparisonBenchmark.zioKafka      avgt   25  1002.338 ± 101.240  ms/op
```

2 times faster! 🚀


erikvanoosten:
Benchmark reports:

master:
```
[info] ConsumerBenchmark.throughput               avgt   25   465.958 ± 100.053  ms/op
[info] ConsumerBenchmark.throughputWithCommits    avgt   25   433.205 ±  85.996  ms/op
[info] ConsumersComparisonBenchmark.kafkaClients  avgt   25   664.730 ±  12.088  ms/op
[info] ConsumersComparisonBenchmark.zioKafka      avgt   25  1999.502 ± 717.557  ms/op
```

This branch:
```
[info] ConsumerBenchmark.throughput               avgt   25   396.172 ±  68.890  ms/op
[info] ConsumerBenchmark.throughputWithCommits    avgt   25   407.719 ±  52.321  ms/op
[info] ConsumersComparisonBenchmark.kafkaClients  avgt   25   746.998 ±  28.032  ms/op
[info] ConsumersComparisonBenchmark.zioKafka      avgt   25  2029.001 ± 538.249  ms/op
```

All lower numbers but also all within the error-margin.

erikvanoosten:
I made a mistake with the patterns. Fixing them...

erikvanoosten:
Patterns fixed and benchmark updated.

erikvanoosten:
The earlier benchmark contains a feedback loop that does a lot more resuming than it should.

With the latest change we get:

```
[info] ConsumerBenchmark.throughput               avgt   25   524.119 ±  85.142  ms/op
[info] ConsumerBenchmark.throughputWithCommits    avgt   25   415.531 ±  65.929  ms/op
[info] ConsumersComparisonBenchmark.kafkaClients  avgt   25   666.154 ±  17.329  ms/op
[info] ConsumersComparisonBenchmark.zioKafka      avgt   25  1793.227 ± 635.461  ms/op
```

It is faster but all in the error margin. This is not very encouraging.

erikvanoosten:
Yes, this is better:

```
[info] ConsumerBenchmark.throughput               avgt   25   240.436 ±  23.397  ms/op
[info] ConsumerBenchmark.throughputWithCommits    avgt   25   324.053 ±  28.001  ms/op
[info] ConsumersComparisonBenchmark.kafkaClients  avgt   25   660.852 ±  10.127  ms/op
[info] ConsumersComparisonBenchmark.zioKafka      avgt   25  1197.165 ± 223.556  ms/op
```

svroonland:
Ran a comparison again on my machine (2019 2,4 GHz 8-Core Intel Core i9)

This branch:

```
[info] Benchmark                                  Mode  Cnt    Score     Error  Units
[info] ConsumerBenchmark.throughput               avgt   25  189,833 ±   7,312  ms/op
[info] ConsumerBenchmark.throughputWithCommits    avgt   25  287,570 ±   2,914  ms/op
[info] ConsumersComparisonBenchmark.kafkaClients  avgt   25  676,153 ±  21,006  ms/op
[info] ConsumersComparisonBenchmark.zioKafka      avgt   25  863,424 ± 122,444  ms/op
```

#803 
```
[info] Benchmark                                  Mode  Cnt    Score    Error  Units
[info] ConsumerBenchmark.throughput               avgt   25  189,069 ± 6,276   ms/op
[info] ConsumerBenchmark.throughputWithCommits    avgt   25  289,007 ± 3,590   ms/op
[info] ConsumersComparisonBenchmark.kafkaClients  avgt   25  693,354 ± 31,611  ms/op
[info] ConsumersComparisonBenchmark.zioKafka      avgt   25  838,962 ± 70,050  ms/op
```

I'd say performance is virtually equal. I do think #803 is simpler to understand and explain..

erikvanoosten:
> I'd say performance is virtually equal. I do think #803 is simpler to understand and explain..

I am happy about that because I found it hard to explain why one would be faster than the other.

I do not find this an argument given that Runloop is already pretty complicated. For the case you disagree, I think both have about equal complexity, perhaps 803 is a tiny bit more complex because it still needs the Request mechanism while 809 embraces it.

Anyways, IMHO we should look at memory usage. My hypothesis is that 809 can reach the same performance with less memory usage. I hope we can test this with the benchmarks as well.

@guizmaii is this command suitable for comparing memory usage as well?
```
zioKafkaBench/Jmh/run -wi 10 -i 10 -r 1 -w 1 -t 1 -f 5 -foe true -prof async:event=alloc
```

svroonland:
@erikvanoosten A few thoughts:

1. There is some initial learning period for this feature to 'kick in' right? So for the second Request of a partition we would not have yet prefetched the data, only for the third request we would 
2. Let's say we get a Request every `N` poll iterations. We will prefetching for that partition at iteration `N-1`. The Request itself may arrive anytime during this poll iteration and will have to wait until the poll completes. That would add some unneeded (in the sense that we could optimize it away) latency.
3. We're not guaranteed to get data for every partition that we resume, right? So we may still sometimes not be on time for the Request. 

That being said, this method improves benchmarks by a lot and more than #803 does and it's a great feature.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/OptimisticResume.scala
```diff
@@ -0,0 +1,83 @@
...
+   * Warning! Be aware of the feedback loop:
+   *
+   * If a pattern matches, the partition will be resumed (adding a `1` to the poll history). If in the next poll,
+   * another pattern matches, the partition will be resumed again. This is not desirable because the stream might not
+   * more data already.
+   *
+   * As a consequence, if pattern `p` is included, `p1` may _not_ be included.
+   */
+  private val OptimisticResumePollPatterns: Seq[(Int, Int)] =
+    Seq(
```
svroonland, line 49: Maybe use a Chunk here

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -49,6 +52,20 @@ private[internal] final class PartitionStreamControl private (
 
   val tpStream: (TopicPartition, ZStream[Any, Throwable, ByteArrayCommittableRecord]) =
     (tp, stream)
+
+  def pollHistory(): PollHistory = pollResumedHistory
```
svroonland, line 56: No need for the ()

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -15,6 +16,8 @@ private[internal] final class PartitionStreamControl private (
   completedPromise: Promise[Nothing, Unit]
 ) {
 
+  private var pollResumedHistory: PollHistory = 0
```
svroonland, line 19: I'm a bit uncomfortable with using `vars` in functional scala code. A `Ref` would probably not hinder performance and is more idiomatic.

erikvanoosten: The consequence of making this a Ref is a lot of code restructuring. We even need to rewrite method `handlePoll` in `Runloop` because method `resumeAndPausePartitions` will become effectful. It is al doable of course.

`handlePoll` currently freely accesses the consumer (no wrapping with zio.attempt), if we're going to make this idiomatic, I would like to change that as well.

WDYT?

erikvanoosten:
@svroonland 

>     1. There is some initial learning period for this feature to 'kick in' right? So for the second Request of a partition we would not have yet prefetched the data, only for the third request we would

Yes, that is correct. Currently the shortest pattern is 3 polls, so 3 polls (150ms with default settings) is the lowest learning time. After the first 3 polls more pre-fatch patterns can match so more polls become eligible for pre-fetching.

>     2. Let's say we get a Request every `N` poll iterations. We will prefetching for that partition at iteration `N-1`. The Request itself may arrive anytime during this poll iteration and will have to wait until the poll completes. That would add some unneeded (in the sense that we could optimize it away) latency.

Correct. However, since the previous approach had the same issue, a pre-fetch still reduces the average waiting time with 1 full poll cycle.

>     3. We're not guaranteed to get data for every partition that we resume, right? So we may still sometimes not be on time for the Request.

Correct. Same is true for previous approach. So a pre-fetch still lowers the average waiting time with 1 poll cycle.

The algorithm will also sometimes predict a pre-fetch when it is not yet necessary (reducing waiting time though waiting for the next poll would have been better), or fail to predict one when it is necessary (increasing wait time). Combined this somewhat lowers the effectiveness of the current algorithm. More fancy algorithms should be possible.

For a more accurate estimate of the waiting time reduction, you can probably iterate all possibilities or do a montecarlo simulation for different situations. My math skills are not up to that task though 😄 

I was thinking about adding some metrics to see how well the algorithm is doing. However, we can only detect failure to predict pre-fetches. We cannot detect whether a predicted pre-fetch was correct (though we could when we track more information). To make meaningful statements about the pre-fetching algorithm we need both metrics.

guizmaii:
Performances measured on my computer:
```scala
[info] Result "zio.kafka.bench.ConsumersComparisonBenchmark.kafkaClients":
[info]   641.297 ±(99.9%) 2.210 ms/op [Average]
[info]   (min, avg, max) = (634.237, 641.297, 659.168), stdev = 4.465
[info]   CI (99.9%): [639.087, 643.508] (assumes normal distribution)
...
[info]   734.035 ±(99.9%) 18.438 ms/op [Average]
[info]   (min, avg, max) = (674.275, 734.035, 875.752), stdev = 37.247
[info]   CI (99.9%): [715.596, 752.473] (assumes normal distribution)
...
[info] Benchmark                                  Mode  Cnt    Score    Error  Units
[info] ConsumersComparisonBenchmark.kafkaClients  avgt   50  641.297 ±  2.210  ms/op
[info] ConsumersComparisonBenchmark.zioKafka      avgt   50  734.035 ± 18.438  ms/op
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/OptimisticResume.scala
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * A sequence of partition statuses (resumed or paused) before a poll. We use one bit per poll, where value 1
+   * indicates that the partition was resumed and value 0 indicates it was paused.
+   *
+   * The most recent poll is in the least significant bit, the oldest poll is in the most significant bit.
+   */
+  type PollHistory = Int
+
+  implicit class PollHistoryOps(val pollHistory: PollHistory) extends AnyVal {
```
guizmaii, line 18:
```suggestion
  implicit final class PollHistoryOps(val pollHistory: PollHistory) extends AnyVal {
```

erikvanoosten: Done in https://github.com/zio/zio-kafka/pull/809/commits/f40255954ad1275e1ea6eb9adc1826795e2ae693

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
guizmaii: I think that we can use a `var` like this but it needs to be encapsulated and shouldn't not leak (nor its value) outside of the encapsulation.

I need more time to review but, at first view, I'd say that there's maybe a missing abstraction in this PR.
This abstraction would encapsulate this `var` and would only expose operators on it.

erikvanoosten: `toResume` is properly encapsulated, however, `(pollHistory & 1) == 0` is not. I have moved it to a method in https://github.com/zio/zio-kafka/pull/809/commits/ff1f2c24a66296c906f6fa0638b668925b47614f

guizmaii: > toResume is properly encapsulated

I think that the `pollHistory` value should not be exposed. That's what I mean by saying that the `toResume` logic is leaky

erikvanoosten: Okay, like in this commit? https://github.com/zio/zio-kafka/pull/809/commits/ff1f2c24a66296c906f6fa0638b668925b47614f

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -317,7 +313,7 @@ object Consumer {
       for {
         r <- ZIO.environment[R & R1]
         _ <- partitionedStream(subscription, keyDeserializer, valueDeserializer)
-               .flatMapPar(Int.MaxValue, bufferSize = settings.perPartitionChunkPrefetch) { case (_, partitionStream) =>
+               .flatMapPar(Int.MaxValue, bufferSize = 2) { case (_, partitionStream) =>
```
guizmaii, line 316: Why `2`? 🤔

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -233,17 +234,34 @@ private[consumer] final class Runloop private (
...
+      val toResume = requestedPartitions.contains(tp) ||
+        pollHistory.optimisticResume
+
+      if (toResume) {
+        if (!pollHistory.latestWasResumed) resumeTps.add(tp)
+      } else {
+        if (pollHistory.latestWasResumed) pauseTps.add(tp)
+      }
+
+      stream.addPollHistory(toResume)
```
guizmaii, line 260: Isn't this incorrect when `toResume == true && pollHistory.latestWasResumed == true`?

Shouldn't it be:
```
      if (toResume) {
        if (!pollHistory.latestWasResumed)
           resumeTps.add(tp)
           stream.addPollHistory(true)
        }
      } else {
        if (pollHistory.latestWasResumed) {
          pauseTps.add(tp)
          stream.addPollHistory(false)
        }
      }
```

?? 🤔

erikvanoosten: To get a good view of the poll history, we need to capture the state for _every_ poll, so `addPollHistory` needs to be invoked regardless of the previous state.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
erikvanoosten: @guizmaii Shall we resolve this conversation?

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
erikvanoosten: @svroonland do you have an idea what bufferSize value would be appropriate?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
guizmaii: I'm wondering if we're not missing something here 🤔

Shouldn't we need a thrid state, "Not pause nor resumed", because it feels to me that we're lying here?
We're "saving" a resumed/paused information which is not true.

What if we were introducing this third state? 
Let's denote it `x` for the example:
```
1xxx0xx1xx
```

🤔

We'd have a more accurate information yet keeping the current state of the partition.
`1xx` means that the partition is in a "resume" state and `1xx0x` would me that we're in a paused state

erikvanoosten: @guizmaii I looked what `KafkaConsumer.resume()` and `KafkaConsumer.pause()` are actually doing and I now believe that the little additional work they need to do when we resume or pause all partitions (even when they were already resumed or paused), does not weigh against the additional complexity of our `resumeAndPausePartitions` method. I have therefore removed this optimization.

In addition, I have strengthened and fixed the documentation of this feature in many place.

See commit https://github.com/zio/zio-kafka/pull/809/commits/3edcb5a3f394297cc4cbddf479d1c4398f956314

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
erikvanoosten: I have changed it to the default value in https://github.com/zio/zio-kafka/pull/809/commits/9676780e51e915c329670aa06c66d23c10b97b3b

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: @guizmaii Can this conversation be resolved?

guizmaii:
Performances with the new changes:
```scala
[info] Result "zio.kafka.bench.ConsumersComparisonBenchmark.kafkaClients":
[info]   640.530 ±(99.9%) 1.855 ms/op [Average]
[info]   (min, avg, max) = (630.543, 640.530, 647.617), stdev = 3.747
[info]   CI (99.9%): [638.676, 642.385] (assumes normal distribution)
...
[info]   737.614 ±(99.9%) 19.178 ms/op [Average]
[info]   (min, avg, max) = (677.381, 737.614, 820.314), stdev = 38.740
[info]   CI (99.9%): [718.437, 756.792] (assumes normal distribution)
...
[info] Benchmark                                  Mode  Cnt    Score    Error  Units
[info] ConsumersComparisonBenchmark.kafkaClients  avgt   50  640.530 ±  1.855  ms/op
[info] ConsumersComparisonBenchmark.zioKafka      avgt   50  737.614 ± 19.178  ms/op
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
guizmaii: > I finally understand your confusion! I will update the code a bit to clear this up.
>
> A partition is always either paused or resumed.
>
> Some partitions are not added to resumeTps or pauseTps because they are already resumed/paused. This is a (potentially premature) performance optimization.

I don't think I was confused by this.

> A partition is always either paused or resumed.

Yes, I understand this but we're not changing its state in each "turn". 
Some "turns" are not resuming nor pausing (or at least were not before your last changes). 
That's why I was suggesting this third state, which I denoted `x` for the example.

As I was explaining, with this third state, we'd end up with an history looking like this `1xx0x1xx1` (random made-up example) and that we could always know in which state the partition was based on its last number before the `xx`:
- `1`, `1xx`, `1xx0x1`, `1xx0x1xx`, etc. == resumed
- `0`, `0xx`, `1xx0x`, `1xx0x1xx0`, etc. == paused

It was a more accurate representation of the state of the partition compared to your previous code, which added `1` or `0` to the history even if `.resume` or `.pause` was not called.

Based on this more accurate information, we could take more accurate decisions/make more accurate predictions on when to resume.

erikvanoosten: I do not understand your proposal, `1x` contains the exact same information as `11`.

erikvanoosten: > I don't think I was confused by this.

Sorry for thinking you were! 😄

erikvanoosten: BTW, there is one aspect that we are not storing in the history: whether we are resuming because of a request, or because of a predicted request. With this information we could possibly make more accurate predictions. However, I suspect that the whole prediction logic will become very complex fast. It may even need more information like the size of the record queue.

I didn't dive into this approach because this simple version is very performant (simple bit operations) and already brings a lot of benefit.

We can definitely improve further, but I'd like to focus on other tasks first.

guizmaii: >  then `1x` contains the exact same information as `11`.

I don't really agree with that.

`1x` contains 3 pieces of information:
- the partition is in a resume state
- the last "turn/round" didn't change the state of the partition
- the partition state was changed 2 "turns/rounds" before

`11` only contains 1 piece of information:
- the partition is in a resume state

guizmaii: @erikvanoosten Can you please add more documentation on each pattern about why each pattern means that we need to resume optimistically, please?

That's probably what I struggle with the most within your proposal.
For example, why do we need to optimistically resume after 1 pause followed by 5 resumes, ie. the `011111`?

Also, can you please document how you found these patterns (how you found we needed these ones and not different ones) and why you think they cover all the possibilities?

Because that's my main issue with your proposal. 
I can be wrong and/or have missed something, but It feels to me like dark magic that will be incredibly hard to understand the behaviour it'll generate and so will be incredibly hard to debug and fix if necessary.

erikvanoosten: > `11` only contains 1 piece of information:

If we look at the second bit in `11`, and then compare it the previous bit we see that they are equal. And therefor we know that the state of the partition did not change (so we know bullet 2 as well).

Because `11` is implicitly the same as `00000000000000000000000000000011` because we store the last 32 states. So we also know that 2 polls ago the partition state changed from paused to resume (so we know bullet 3 as well). Only if the state didn't change for 32 polls, we can not derive when the partition state changed.

BTW, calling `resume` on an already resumed partition is a no-op. So it is not something we need to track.

erikvanoosten: > Can you please add more documentation...

@guizmaii Sure. More documentation available in https://github.com/zio/zio-kafka/pull/809/commits/4ac4e1a25897983da9da7d800dd919088c8cc964

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PollHistory.scala
```diff
@@ -0,0 +1,131 @@
...
+   */
+  def addPollHistory(resumed: Boolean): PollHistory
+
+  /**
+   * @return
+   *   the recorded history as a bit string where each character is either a '1': the partition was resumed or '0': the
+   *   partition was paused. The oldest poll comes first and the newest last. Leading zeros are removed. Since the
+   *   history length is limited, the return string will always have a bounded length.
+   */
+  def asBitString: String
```
guizmaii, line 33: This is only used in tests. 

@erikvanoosten Can you please remove it from the interface and put the code in the test directory, please?

erikvanoosten: How should we do that? Everything in PollHistoryImpl is private so we cannot give the test code access.

guizmaii: Here's a way to do it https://github.com/zio/zio-kafka/pull/838 :)

erikvanoosten: Thanks, yes, we need to expose something to do this. I'll use your solution from 838.

guizmaii (approved):

PR #812, opened 2023-04-15
New code owners part 2

erikvanoosten:
In build.sbt there is also this text:

```
Copyright 2021 Itamar Ravid and the zio-kafka contributors.
```

Should we keep that? Should we update the year to `2021 - <current year>`?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -25,6 +25,12 @@ inThisBuild(
     ),
     developers := List(
       Developer(
+        "iravid",
+        "Itamar Ravid",
+        "iravid@iravid.com",
+        url("https://github.com/iravid")
+      ),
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -25,10 +25,22 @@ inThisBuild(
     ),
     developers := List(
       Developer(
-        "iravid",
-        "Itamar Ravid",
-        "iravid@iravid.com",
-        url("https://github.com/iravid")
```
guizmaii, line 31: I think we should leave Itamar, as he initiated the project

erikvanoosten: You are right. I have added him back.

svroonland (approved):

Issue #590, opened 2023-01-16
Many records duplicately processed after rebalancing


svroonland:
Originally reported on the zio-kafka discord by user `amszio`

> We are working on a new data processing application based on zio-kafka v2.0.1. Performance of the zio-kafka is amazing. However, we found a major issue during our testing phase. During rebalancing, zio-kafka is duplicating messages. Issue exists in v2.0.1, v2.0.2 and v2.0.3 (scala 2.13 and 3.2.1 have same issue). I saw https://github.com/zio/zio-kafka/issues/280 is fixed but looks like issue still exists. Anyone faces same or similar issue? Any help is appreciated.
> 
> To reproduce locally;
> * I've copied following code piece from zio-kafka homepage to be sure it's not due my business logic
> * Have a running kafka (docker-compose).
> * Input and output topics with min 2 partitions.
> * Send some data to input topic.
> * Start first zio-kafka app.
> * Start second zio-kafka app after first zio-kafka start consuming messages (this triggers rebalancing).
> * Wait until all messages are consumed by two apps.
> * Count messages in output_topic.
> 
> Note: When I start two applications at the same time and don't stop any of them till the end, this issue doesn't exist, both input and output topics have same amount of data.
> 
>     Consumer
>       .subscribeAnd(Subscription.topics("input_topic"))
>       .plainStream(Serde.byteArray, Serde.string)
>       .map { record =>
>         val producerRecord: ProducerRecord[Array[Byte], String] = new ProducerRecord("output_topic", record.record.value)
>         (producerRecord, record.offset)
>       }
>       .mapChunksZIO { chunk =>
>         val records = chunk.map(_._1)
>         val offsetBatch = OffsetBatch(chunk.map(_._2))
>         Console.printLine(s"Producing ${records.size} records") *>
>           ZIO.attemptBlocking(Thread.sleep(1000)) *> //Not to send tons of messages to reproduce the issue.
>           Producer.produceChunk[Any, Array[Byte], String](records, Serde.byteArray, Serde.string) *>
>           offsetBatch.commit.as(Chunk(()))
>       }
>       .runDrain

svroonland:
See #591

PR #830, opened 2023-04-22
Await commits during revoke

erikvanoosten:
To solve #590, in this PR we introduce a new mode that holds up a rebalance until the offsets of messages provided to streams of revoked partitions have been committed.

### Motivation

Here is a common (single partition) scenario around rebalances:

1. a consumer polls some messages and puts them in the streams (let's say messages with offsets 0 to 100)
1. asynchronously, the user processes these messages. Some of them are committed (let's say up to offset 50), the rest is still being processed when...
1. a rebalance happens, the partition is revoked and assigned to another consumer on another instance
1. the consumer continues to process the remaining messages with offsets 50 to 100, and tries to commit those offsets
1. _at the same time,_ another consumer on another instance, starts consuming from the last committed offset (which is 50) and will process the same messages with offsets 50 to 100

Messages with offsets 50 to 100 are being processed by both consumers simultaneously. Note that both consumers will try to commit these offsets. Until the first consumer is ready, the stored offsets can go up and down and are therefore unreliable.

After merging this PR, the scenario will unfold as follows:

1. a consumer polls some messages and puts them in the streams (let's say messages with offsets 0 to 100). Zio-kafka keeps track of the highest provided offset
1. asynchronously, the user processes these messages. Some of them are committed (let's say up to offset 50), the rest is still being processed when...
1. a rebalance happens, the partition is revoked and assigned to another consumer on another instance
   * the consumer continues to process the remaining messages with offsets 50 to 100, and tries to commit those offsets
   * inside the onRevoked callback, zio-kafka continues to process commit commands from the user
   * zio-kafka continues to do so until the commit with the highest provided offset (offset 100) completes
   * the onRevoked callback completes, signalling to Kafka that the next consumer may start consuming from the partition
1. another consumer on another instance, starts consuming from the last committed offset (which is now 100, problem solved!)

### Complications

1. The chosen solution is not suitable for all consumers.
   - There are uses cases where not all messages are read from the stream. For example, some want to read exactly 100 messages from a topic and then stop consuming. In that case the user has no intention to commit all messages and therefore we should not wait for that to happen. Since stream consumers can basically do whatever they want, the only way we can support such use cases is by letting the consumer tell zio-kafka that they are done with committing. This requires an API change. For example, we can let the user tell zio-kafka that a given commit is the last one.
   - Not all consumers commit offsets (to Kafka) in the first place. In a future PR we could make it work for commits to other stores though. As a workaround, when this PR is merged, these users can commit to both places.
4. Prior to Kafka client 3.6.0 there is no way to wait for commits to complete except for calling poll. However, calling poll from the rebalance listener callback does not work.
5. The Kafka client requires that any nested invocations (that is, from the rebalance listener callback) to the client happens from the same thread. This is at very much at odds with how ZIO works. Attempts to convince the Kafka committers to relax this requirement failed; they could not be convinced that this is a problem. We can somewhat workaround this problem by using a special run-on-caller-thread zio-runtime (already part of zio-kafka). However, some operations, in particular `ZIO.timeout`, still make processing jump to another thread. Therefore, we cannot use `ZIO.timeout` to limit the time we spend in the onRevoke callback.
6. The new functionality is somewhat complicated to test. We need to make sure that there are unprocessed messages in the stream when a rebalance starts.

### Related

The same issue is present in:
- f2s-kafka: https://github.com/fd4s/fs2-kafka/issues/1200
- alpakka-kafka: https://github.com/akka/alpakka-kafka/issues/1038

In fact, every program that does polls and commits asynchronously is likely affected.

### Things to do

- [x] Wait for Kafka 3.6.0 (it contains https://github.com/apache/kafka/pull/13678).
- [ ] Make it work.
- [x] Come up with a better name for the new mode.
- [ ] Rewrite or duplicate existing tests so that they commit everything and can work with this new mode.
- [ ] Add more test specifically for the new mode.
- [ ] Reduce logging.
- [ ] Documentation.

### Non-goals

This PR does not try to solve the following goals. However, these can be addressed in future PRs.

- Support transactional consumer/producer.
- Support external commits.
- Support consumers that want to commit only a portion of the given messages.

This branch is based on the work of abandoned #788.

### DIFF
--- a/zio-kafka-test/src/test/scala/zio/kafka/ZIOSpecDefaultSlf4j.scala
+++ b/zio-kafka-test/src/test/scala/zio/kafka/ZIOSpecDefaultSlf4j.scala
@@ -2,16 +2,21 @@ package zio.kafka
 
 import zio.Chunk
 import zio.logging.backend.SLF4J
-import zio.test.{ TestAspect, TestAspectAtLeastR, TestEnvironment, ZIOSpecDefault }
+import zio.test.{TestAspect, TestAspectAtLeastR, TestEnvironment, ZIOSpecAbstract, ZIOSpecDefault}
 
 /**
  * Use this class instead of `ZIOSpecDefault` if you want your tests to use SLF4J to log.
  *
  * Useful when you want to use logback to configure your logger, for example.
  */
-abstract class ZIOSpecDefaultSlf4j extends ZIOSpecDefault {
+trait ZIOSpecDefaultSlf4j extends ZIOSpecDefault with ZIOSpecAbstractSlf4j
 
-  override def aspects: Chunk[TestAspectAtLeastR[TestEnvironment]] =
+/**
+ * Mix this into your ZIO Spec if you want your tests to use SLF4J to log.
+ */
+trait ZIOSpecAbstractSlf4j extends ZIOSpecAbstract {
+
+  abstract override def aspects: Chunk[TestAspectAtLeastR[Environment with TestEnvironment]] =
     super.aspects :+ TestAspect.fromLayer(zio.Runtime.removeDefaultLoggers >>> SLF4J.slf4j)
 
 }
--- a/zio-kafka-test/src/test/scala/zio/kafka/consumer/RebalanceSafeCommitConsumerSpec.scala
+++ b/zio-kafka-test/src/test/scala/zio/kafka/consumer/RebalanceSafeCommitConsumerSpec.scala
@@ -1,42 +1,29 @@
 package zio.kafka.consumer
 
 import io.github.embeddedkafka.EmbeddedKafka
-import org.apache.kafka.clients.consumer.{
-  ConsumerConfig,
-  ConsumerPartitionAssignor,
-  CooperativeStickyAssignor,
-  RangeAssignor
-}
-import org.apache.kafka.clients.producer.ProducerRecord
+import org.apache.kafka.clients.consumer.{ConsumerConfig, ConsumerPartitionAssignor, CooperativeStickyAssignor, RangeAssignor}
+import org.apache.kafka.clients.producer.{ProducerRecord, RecordMetadata}
 import org.apache.kafka.common.TopicPartition
 import zio._
-import zio.kafka.ZIOSpecDefaultSlf4j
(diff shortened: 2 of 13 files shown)


KamilPostrozny:
Question - are you trying to make it work while using external offsets (not commiting to kafka)? Or is it only working in case when we commit to kafka using `.commit`?

erikvanoosten:
> Question - are you trying to make it work while using external offsets (not commiting to kafka)? Or is it intended to work only in case when we commit to kafka using `.commit`?

First of all, the new mode will be optional and not enabled by default. It should be kept disabled for external commits.

However, supporting external commits would be very nice. We should try that in a later PR. Would you like to create an issue for that? Bonus if you have suggestions for the API or on how to implement it.

KamilPostrozny:
> > Question - are you trying to make it work while using external offsets (not commiting to kafka)? Or is it intended to work only in case when we commit to kafka using `.commit`?
> 
> First of all, the new mode will be optional and not enabled by default. It should be kept disabled for external commits.
> 
> However, supporting external commits would be very nice (thanks for asking!). We should try that in a later PR. Would you like to create an issue for that? Bonus if you have suggestions for the API or on how to implement it.

But from what I understand it would be possible to use this mode when I would commit both externally and internally, right? If only commits to kafka happen after external commits then it should work as intended. 

I will create the issue for the case with external-only commits but I don't have any idea yet on how to implement it.

erikvanoosten:
> But from what I understand it would be possible to use this mode when I would commit both externally and internally, right? If only commits to kafka happen after external commits then it should work as intended.

Yes, that would work.

erikvanoosten:
I now have a working draft. It requires a patched kafka library (it must contain PR https://github.com/apache/kafka/pull/13678) and a patched embedded-kafka.

erikvanoosten:
Replaced by #1098.

erikvanoosten:
> Question - are you trying to make it work while using external offsets (not commiting to kafka)? Or is it intended to work only in case when we commit to kafka using `.commit`?

@KamilPostrozny zio-kafka 2.10.0 will support this with external commits. Zio-kafka 3.0.0 will also support this with transactional commits.

PR #836, opened 2023-05-02
Jules' review

guizmaii:


zio-kafka/src/main/scala/zio/kafka/consumer/internal/PollHistory.scala
```diff
@@ -31,28 +32,41 @@ private[internal] trait PollHistory {
...
+   */
+  private object EmptyPollHistory extends PollHistory {
+    override def optimisticResume: Boolean                            = false
+    override def addPollHistory(resumed: PoolHistoryBit): PollHistory = new PollHistoryImpl(resumed)
+    override def latestWasResumed: Boolean                            = false
+  }
+
+  val Empty: PollHistory = EmptyPollHistory
+
+  type PoolHistoryBit = Int
```
erikvanoosten, line 49:
```suggestion
  type PollHistoryBit = Int
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PollHistory.scala
```diff
@@ -31,28 +32,41 @@ private[internal] trait PollHistory {
 }
 
 private[internal] object PollHistory {
-  val Empty: PollHistory = PollHistoryImpl.Empty
 
-  private class PollHistoryImpl private (resumeBits: Int) extends PollHistory {
+  /**
+   * An empty poll history.
+   *
+   * The implementation is optimized but is equivalent to `new PollHistoryImpl(0)`.
```
erikvanoosten, line 39: This will hurt performance quite badly.

The empty case will only be used for the first poll of a partition. So it will never get a chance to be properly optimized by the JVM.

The biggest effect though will be because there are now multiple implementations for `PollHistory`. This causes the JVM to not perform all the inlining optimizations it can do when there is a single implementation.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PollHistory.scala
```diff
@@ -21,7 +22,7 @@ private[internal] trait PollHistory {
    * @param resumed
    *   true when this partition was 'resumed' before the poll, false when it was 'paused'
    */
-  def addPollHistory(resumed: Boolean): PollHistory
+  def addPollHistory(resumed: PoolHistoryBit): PollHistory
```
erikvanoosten, line 25: This is tricky. Only values 0 and 1 make sense however, PolIHistoryBit is an Int! This makes it possible to use the method in a totally wrong way.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -48,7 +49,8 @@ private[internal] final class PartitionStreamControl private (
   val tpStream: (TopicPartition, ZStream[Any, Throwable, ByteArrayCommittableRecord]) =
     (tp, stream)
 
-  def pollHistory: PollHistory = pollResumedHistory
+  def optimisticResume: Boolean = pollResumedHistory.optimisticResume
+  def latestWasResumed: Boolean = pollResumedHistory.latestWasResumed
```
erikvanoosten, line 53: 👍

erikvanoosten (commented):
Hiding poll-history further is a good idea. However, specializing the empty history and using Int as parameter type (see comments below) are IMHO not things we should do.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PollHistory.scala
guizmaii: Removed! Thanks 🙂

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PollHistory.scala
guizmaii: If we were introducing newtypes, we could go further with type-safety and only allow `0` and `1` but that's require us to depend on zio-prelude and, even if I love zio-prelude's newtypes, I prefer not to introduce a new dependency

guizmaii: Anyway, I removed it and went back to the `boolean`. It was not the best idea

guizmaii:
@erikvanoosten 
> However, specializing the empty history and using Int as parameter type (see comments below) are IMHO not things we should do.

You're right. I removed them

PR #843, opened 2023-05-11
Remove all rights reserved

erikvanoosten:
The phrase 'all rights reserved' is not needed. Also, since this library is provided under the Apache 2 license, it doesn't even make sense.

Also: add current year to copyright statement.


erikvanoosten:
This is a retry from #840.

guizmaii (approved):

PR #844, opened 2023-05-12
Pre-emptively pause long resumed partitions to prevent read starvation

erikvanoosten:
The Kafka client has a tendency to consume everything from 1 partition before going to the next one. This can cause a partition from not being read for a long time, or even forever. To prevent this, when multiple partitions are resumed, we force pause the longest resumed partition (though we allow it to be resumed for 4 polls).

Zio-kafka version 2.3 introduced optimistic pre-fetching. This causes partitions to be unpaused much more often. Because of this, mentioned Kafka behavior became more apparent for high load applications.

Todo before this PR goes out of draft:
 - [x] make algorithm less eager
 - [x] write tests
 - [ ] Fix performance regression (see https://github.com/zio/zio-kafka/pull/844#issuecomment-1573545507)
 - [ ] do not use package object (will be dropped in scala 3)

Also, since this might be related, we need to see if we can fix the increase of CPU usage and GC pressure of zio-kafka 2.3.


guizmaii:
I'm not sure we want to manipulate this behaviour. If Kafka is doing that, it's probably because it optimizes the consumption, the interaction between the client and the server and the loading of data from the disk on the server side

erikvanoosten:
The Kafka client optimizes for throughput. Some people need low latency, or a balance between latency and throughput.

svroonland:
Could we write a test case demonstrating this behavior?

erikvanoosten:
> Could we write a test case demonstrating this behavior?

By 'this behavior' do you mean the tendency of the client to consume one partition as long as possible, or do you mean a test that validates that the proposed implementation works as expected (or something else)?

svroonland:
The former

erikvanoosten:
Earlier I wrote:
> the akka streaming client does something like this.

This statement was based on anecdotal evidence 😉 , however, now I look in the Alpakka code, I cannot find such a feature. So now I am not sure where I heard this.

erikvanoosten:
> The former

We have observed this at work. You can see this by tracking lag of the different partitions.

We're not the only the see this. A user reported a few days ago that this is happening to them on zio-kafka: https://discord.com/channels/629491597070827530/629497941719121960/1105883502710366228

Also this is interesting: https://stackoverflow.com/questions/76175977/how-to-make-consumer-to-switch-between-partitions-in-round-robbin-fashion

erikvanoosten:
> The implementation in this PR is very eager. We can easily tune it down a bit. For example, we could delay force-pausing a partition after at least n resume polls. I doubt anyone will suffer from this.

This has now been implemented.

zio-kafka-test/src/test/scala/zio/kafka/consumer/internal/package.scala
```diff
@@ -0,0 +1,18 @@
+package zio.kafka.consumer
+
+package object internal {
```
guizmaii, line 3: `package object` are remove from Scala3 IIRC

See https://docs.scala-lang.org/scala3/reference/dropped-features/package-objects.html

You can just use an `object` I think

guizmaii:
Benchmarks results:

Before
```scala
zioKafkaBench/Jmh/run -wi 10 -i 10 -r 1 -w 1 -t 1 -f 3 -foe true
...
[info]   739.566 ±(99.9%) 21.019 ms/op [Average]
[info]   (min, avg, max) = (682.067, 739.566, 813.396), stdev = 31.461
[info]   CI (99.9%): [718.547, 760.586] (assumes normal distribution)
[info] # Run complete. Total time: 00:18:24
...
[info] Benchmark                                        Mode  Cnt     Score     Error  Units
[info] ConsumerBenchmark.throughput                     avgt   30   205.133 ±   6.544  ms/op
[info] ConsumerBenchmark.throughputWithCommits          avgt   30   284.912 ±   5.204  ms/op
[info] ConsumersComparisonBenchmark.kafkaClients        avgt   30   645.361 ±   3.004  ms/op
[info] ConsumersComparisonBenchmark.manualKafkaClients  avgt   30   641.042 ±   2.877  ms/op
[info] ConsumersComparisonBenchmark.manualZioKafka      avgt   30  2651.134 ± 100.297  ms/op
[info] ConsumersComparisonBenchmark.zioKafka            avgt   30   739.566 ±  21.019  ms/op
```

After
```scala
zioKafkaBench/Jmh/run -wi 10 -i 10 -r 1 -w 1 -t 1 -f 3 -foe true
...
[info]   1499.816 ±(99.9%) 216.967 ms/op [Average]
[info]   (min, avg, max) = (1002.257, 1499.816, 2254.385), stdev = 324.746
[info]   CI (99.9%): [1282.849, 1716.783] (assumes normal distribution)
[info] # Run complete. Total time: 00:18:15
...
[info] Benchmark                                        Mode  Cnt     Score     Error  Units
[info] ConsumerBenchmark.throughput                     avgt   30   222.238 ±  13.788  ms/op
[info] ConsumerBenchmark.throughputWithCommits          avgt   30   283.729 ±   6.562  ms/op
[info] ConsumersComparisonBenchmark.kafkaClients        avgt   30   645.774 ±   3.357  ms/op
[info] ConsumersComparisonBenchmark.manualKafkaClients  avgt   30   648.834 ±   3.384  ms/op
[info] ConsumersComparisonBenchmark.manualZioKafka      avgt   30  2412.388 ± 350.380  ms/op
[info] ConsumersComparisonBenchmark.zioKafka            avgt   30  1499.816 ± 216.967  ms/op
```

This might be expected, but there's a clear performance regression.
I think these results are another argument for me to ask for this behaviour to be put behind an option, which would be not activated by default.

if this is an optimisation for latency, I'd see something like:
```scala
sealed trait Mode
object Mode {
  case object Throughput extends Mode
  case object Latency    extends Mode
}

object Consumer {
  def make(
    ...
    mode: Mode = Mode.Throughput // To keep the previous behaviour as the default one
  )
}
```

WDYT?

guizmaii:
@erikvanoosten 2 questions about your comment https://github.com/zio/zio-kafka/pull/844#issuecomment-1570223553

1. You say:
> CPU and GC pressure is slightly elevated but this seems to be a ZIO issue (see https://github.com/zio/zio/issues/8158)

Now that ZIO 2.0.14 containing the mentioned fix is released. Can you confirm that this was indeed because of the issue in ZIO, please?

2. When you say:
> All other metrics like heap usage, latency, consumer lag, etc. are the same as before.

What do you mean by "before"? Which version of zio-kafka are you talking about 2.3.x or 2.2? People are reporting high CPU and memory usage since 2.3.0.

guizmaii:
I think before going forward with these changes, I'd like to introduce the option to disable the optimistic resume so that users like @myazinn who are experiencing issues (high CPU and memory usage, high lag) with 2.3.x can disable it and tell us if it fixes their issue.

WDYT?
Cc @svroonland

erikvanoosten:
> What do you mean by "before"? Which version of zio-kafka are you talking about 2.3.x or 2.2? People are reporting high CPU and memory usage since 2.3.0.

The version I am running is a fork of 2.2. So this is comparing 2.2 with and without the feature this PR introduces.

erikvanoosten:
> > CPU and GC pressure is slightly elevated but this seems to be a ZIO issue (see https://github.com/zio/zio/issues/8158)

> Now that ZIO 2.0.14 containing the mentioned fix is released. Can you confirm that this was indeed because of the issue in ZIO, please?

Sorry, no I cannot confirm that. Worse, I have to (mostly) withdraw my previous observation as well.

I took a second look at the changes from introducing this feature and also at the changes from updating to zio 2.0.14. Now that I am properly averaging the CPU and GC metrics over all the pods, and also, look at them over a longer period (days instead of a few hours) I see that these metrics are too volatile to attribute differences to these features.

The only metric that did change is latency (time from produce in source to end of processing). That metric was not affected by this PR. Upgrading to zio 2.0.14 however had a very strong positive effect.

Please note that these observations are all based on zio-kafka 2.2.

zio-kafka-test/src/test/scala/zio/kafka/consumer/internal/package.scala
erikvanoosten: I'd better move it to a proper class then. It still needs to compile with earlier scala versions.

svroonland:
I can attest that partition skew is not nice to have in streaming applications, where there's usually some latency constraints. Was not aware that kafka itself can cause this, not just a skewed number of messages on partitions. I believe at the core this is a kafka issue that should be resolved in the kafka client or broker. So about the KIP-41 that mentions this problem, was that never implemented? Can we file a new issue somewhere?

As an alternative for this PR:

If I remember correctly, in a project I did a few years ago we had a large variation in message processing times due to the varying nature of the messages. We kept track of the latest message timestamp on the partitions as well as the timestamp of the last processed messages. If the diff for one of the partitions got too large, we would pause the other ones. 

We could implement something similar as a separate component that you can pass your partition streams through, eg a function `Stream[Throwable, (TopicPartition, ZStream[R, Throwable, CommittableRecord[K, V]])] = Stream[Throwable, (TopicPartition, ZStream[R, Throwable, CommittableRecord[K, V]])]`.

erikvanoosten:
> I believe at the core this is a kafka issue that should be resolved in the kafka client or broker.

Yes, definitely that. I am not sure this will happen any time soon though. The best I think we can do is influence the requirements for the kafka client rewrite.

> If the diff for one of the partitions got too large, we would pause the other ones.

I like that idea. It is very fundamental because you directly attack the problem: exceeding latency requirements.

erikvanoosten:
Since:
* this PR causes a major performance degradation
* there is a better idea (https://github.com/zio/zio-kafka/pull/844#issuecomment-1574761690)

I am closing this PR.

PR #845, opened 2023-05-12
Rename the `zio-kafka-test-utils` module to `zio-kafka-testkit` && change its base package from `zio.kafka` to `zio.kafka.testkit`

guizmaii:
With the new name, I want to make this package more easily discoverable by our users.
The package change is to better separate things between zio-kafka and its testkit


svroonland (commented):
Agreed with Erik on the library name change being inconvenient for users. Test kit / utils is quite similar. The namespace change is good though

erikvanoosten (dismissed):
@guizmaii convinced me that we can make this change; only power-users have found their way to test-utils yet. Together with proper documentation, we can make the switch.

guizmaii:
@svroonland See my answer to Erik on Discord: https://discord.com/channels/629491597070827530/629497941719121960/1107265058695815168

Let me know what you think :)

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/Kafka.scala
erikvanoosten, line 1: @guizmaii Can you please check which line-end characters are used? Github shows such a large diff, I suspect that these were changed to `/r/n`.

erikvanoosten (approved):
Just one comment. I am fine with merging this PR with #846.

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/Kafka.scala
guizmaii: It was probably not using the correct line-end character

image

erikvanoosten: > It was probably not using the correct line-end character

Ah, it was using the wrong line-ending before! 👍

erikvanoosten: I checked, this was the only file left to use CRLF.

PR #846, opened 2023-05-12
Write `zio-kafka-testkit` documentation

guizmaii:
Related to #845 

The doc can be read here: https://github.com/zio/zio-kafka/blob/testing_doc/docs/writing-tests.md

TODO:
- [x] Add `ZIOSpecWithKafka` doc
- [x] use mdoc


docs/writing-tests.md
```diff
@@ -0,0 +1,121 @@
...
+We could have decided to share one instance of `Producer` between all tests by moving the `producer` layer to the `provideSomeShared`, which would have
+looked like this:
+```scala
+suite("producer test suite")(
+  // ... tests ...
+).provideSomeShared[Scope](Kafka.embedded, producer)
+```
+
+In this example, we decided to share an instance of Kafka for the entire suite, with the `.provideSomeShared[Scope](Kafka.embedded)`.     
+Kafka is slow to start, so it is better to share it between all tests, so we only start it once.
```
guizmaii, line 58:
```suggestion
Kafka is slow to start, so it is better to share it between all tests of the suite so we only start it once for this test suite.     
```

docs/writing-tests.md
```diff
@@ -0,0 +1,121 @@
...
+import zio.test._
+
+object ProducerSpec extends ZIOSpecDefault {
+  override def spec: Spec[TestEnvironment & Scope, Any] =
+    (
+      suite("producer test suite")(
+        test("one record") {
+          for {
+            _ <- Producer.produce(new ProducerRecord("topic", "boo", "baa"), Serde.string, Serde.string)
+          } yield assertCompletes
```
erikvanoosten, line 34: Could we extend this to also consume the produced data? I would expect that most people want to do some assertions on the produced data. It doesn't need to be in this example; since there are multiple ways to wait until consumed data is available, we could do this in follow-up examples.

zio-kafka-example/src/test/scala/zio/kafka/example/ProducerSpec.scala
erikvanoosten, line 1: We can explore things like mdoc. That would make sure that code examples are up to date.

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/embedded/Kafka.scala
```diff
@@ -32,6 +32,9 @@ object Kafka {
     override def stop(): UIO[Unit]              = ZIO.unit
   }
 
+  /**
+   * Creates an in-memory Kafka instance with a random port
```
erikvanoosten, line 36:
```suggestion
   * Creates an in-memory Kafka instance with a random port.
```

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/embedded/Kafka.scala
```diff
@@ -50,6 +53,9 @@ object Kafka {
     )(_.stop())
   }
 
+  /**
+   * Creates an in-memory Kafka instance with a random port and SASL authentication
```
erikvanoosten, line 57:
```suggestion
   * Creates an in-memory Kafka instance with a random port and SASL authentication.
```

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/embedded/Kafka.scala
```diff
@@ -74,6 +80,9 @@ object Kafka {
     )(_.value.stop())
   }
 
+  /**
+   * Creates an in-memory Kafka instance with a random port and SSL authentication
```
erikvanoosten, line 84:
```suggestion
   * Creates an in-memory Kafka instance with a random port and SSL authentication.
```

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/embedded/Kafka.scala
```diff
@@ -105,6 +114,11 @@ object Kafka {
     )(_.stop())
   }
 
+  /**
+   * Will connect to a Kafka instance running on localhost:9092 (with Docker, for example)
```
erikvanoosten, line 118:
```suggestion
   * Will connect to a Kafka instance running on localhost:9092 (with Docker, for example).
```

docs/writing-tests.md
```diff
@@ -0,0 +1,121 @@
...
+        }
+        // ... more tests ...
+      )
+        .provideSome[Kafka](producer)             // Here, we provide a new instance of Producer per test
+        .provideSomeShared[Scope](Kafka.embedded) // Here, we provide an instance of Kafka for the entire suite
+    ) @@ timeout(5.minutes) @@ sequential
+}
+```
+
+This test is extremely simple.
```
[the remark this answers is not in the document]
guizmaii: What about "This test is very basic"?

docs/writing-tests.md
```diff
@@ -0,0 +1,121 @@
...
+      suite("producer test suite")(
+        test("one record") {
+          for {
+            _ <- Producer.produce(new ProducerRecord("topic", "boo", "baa"), Serde.string, Serde.string)
+          } yield assertCompletes
+        }
+        // ... more tests ...
+      )
+        .provideSome[Kafka](producer)             // Here, we provide a new instance of Producer per test
+        .provideSomeShared[Scope](Kafka.embedded) // Here, we provide an instance of Kafka for the entire suite
```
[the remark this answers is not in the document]
guizmaii: > Is that correct? Isn't this shared between all suites?

AFAIU `provideSomeShared `, I don't think so.
I don't think there's this kind of magic in ZIO.

In ZIO everything is explicit, so sharing something on a suite, i.e.:
```scala
suite(...)(
  // ....
). provideSomeShared(something)
```
has no side-effect on the other test suites. It's only localized to this suite.

Or I don't understand how `provideSomeShared` is working

zio-kafka-example/src/test/scala/zio/kafka/example/ProducerSpec.scala
guizmaii: @khajavi Can we use mdoc?

docs/writing-tests.md
guizmaii: I'll add this in the Consumer examples I'll put in the documentation I think

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
```diff
@@ -8,55 +8,61 @@ import zio.kafka.consumer.Consumer.{ AutoOffsetStrategy, OffsetRetrieval }
...
-    ZIO.serviceWith[Kafka](_.bootstrapServers).map(ProducerSettings(_))
+    ZIO
+      .serviceWith[Kafka](_.bootstrapServers)
+      .map(ProducerSettings(_))
 
+  /**
+   * Producer instance you can use in your tests. It uses the default Producer settings.
+   */
   val producer: ZLayer[Kafka, Throwable, Producer] =
-    (ZLayer.fromZIO(producerSettings) ++ ZLayer.succeed(Serde.string: Serializer[Any, String])) >>>
```
guizmaii, line 41: This `ZLayer.succeed(Serde.string: Serializer[Any, String])` was useless

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
```diff
@@ -8,55 +8,61 @@ import zio.kafka.consumer.Consumer.{ AutoOffsetStrategy, OffsetRetrieval }
...
+    ZLayer.makeSome[Kafka, Producer](
+      ZLayer(producerSettings),
       Producer.live
+    )
 
+  /**
+   * Default transactional Producer settings you can use in your tests.
+   */
   val transactionalProducerSettings: ZIO[Kafka, Nothing, TransactionalProducerSettings] =
-    ZIO.serviceWith[Kafka](_.bootstrapServers).map(TransactionalProducerSettings(_, "test-transaction"))
```
guizmaii, line 45: This `ZLayer.succeed(Serde.string: Serializer[Any, String])` was useless

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
```diff
@@ -8,55 +8,61 @@ import zio.kafka.consumer.Consumer.{ AutoOffsetStrategy, OffsetRetrieval }
...
+      .serviceWith[Kafka](_.bootstrapServers)
+      .map(TransactionalProducerSettings(_, "test-transaction"))
 
+  /**
+   * Transactional Producer instance you can use in your tests. It uses the default transactional Producer settings.
+   */
   val transactionalProducer: ZLayer[Kafka, Throwable, TransactionalProducer] =
-    (ZLayer.fromZIO(transactionalProducerSettings) ++ ZLayer.succeed(
-      Serde.string: Serializer[Any, String]
-    )) >>>
```
guizmaii, line 50: This `ZLayer.succeed(Serde.string: Serializer[Any, String])` was useless

docs/writing-tests.md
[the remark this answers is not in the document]
erikvanoosten: Yes, there is such a mechanism. It is used by the now deprecated `ZIOSpecWithKafka`. I thought `provideShared` gave the same guarantee. But now I understand that is not the case I think we should undepricate `ZIOSpecWithKafka`.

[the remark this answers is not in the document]
guizmaii: I'll ask the question to be sure

docs/writing-tests.md
```diff
@@ -0,0 +1,154 @@
...
+```
+
+Finally, we annotate the suite with the `timeout` and `sequential` aspects.   
+The `timeout` aspect is used to specify a timeout for the entire suite. If the suite takes more than 5 minutes to run, it will fail.    
+The `sequential` aspect is used to specify that the tests in the suite must be run sequentially. This is necessary because Kafka is a shared resource.
+We don't want tests to interfere with each other.
+
+## Testing a consumer
+
+TODO JULES
```
guizmaii, line 75: TODO @guizmaii

zio-kafka-example/src/test/scala/zio/kafka/example/ProducerSpec.scala
khajavi: Yes, this way:

```scala mdoc:compile-only
scala code
```

guizmaii: Thanks!

docs/writing-tests.md
```diff
@@ -0,0 +1,257 @@
...
+        .provideSome[Kafka](producer)             // No need here to provide a Kafka instance, it is already provided by the `ZIOSpecWithKafka` trait
+    ) @@ timeout(5.minutes) @@ sequential
+}
+```
+More details about this `ZIOSpecWithKafka` trait [below](#ziospecwithkafka-trait).
+
+Finally, we annotate the suite with the `timeout` and `sequential` aspects.   
+The `timeout` aspect from zio-test is used to specify a timeout for the entire suite. If the suite takes more than 5 minutes to run, it will fail.    
+The `sequential` aspect from zio-test is used to specify that the tests in the suite must be run sequentially. This is necessary because Kafka is a shared resource.
+We don't want tests to interfere with each other.
```
erikvanoosten, line 87: It is possible to write your tests such that they do not interfere with each other. Therefore I propose to weaken the text a bit to:

```suggestion
The `sequential` aspect from zio-test is used to specify that the tests in the suite must be run sequentially instead of concurrently. Kafka is shared between the tests so running them sequentially prevents tests to influence each other.
```

docs/writing-tests.md
```diff
@@ -0,0 +1,257 @@
...
+        // ... tests ...
+      )
+        .provideSome[Kafka](producer)             // No need here to provide a Kafka instance, it is already provided by the `ZIOSpecWithKafka` trait
+    ) @@ timeout(5.minutes) @@ sequential
+}
+```
+More details about this `ZIOSpecWithKafka` trait [below](#ziospecwithkafka-trait).
+
+Finally, we annotate the suite with the `timeout` and `sequential` aspects.   
+The `timeout` aspect from zio-test is used to specify a timeout for the entire suite. If the suite takes more than 5 minutes to run, it will fail.
```
erikvanoosten, line 85: I propose that we split the paragraph in 2 (add a newline after `it will fail`):
```suggestion
Finally, we annotate the suite with the `timeout` and `sequential` aspects.   
The `timeout` aspect from zio-test is used to specify a timeout for the entire suite. If the suite takes more than 5 minutes to run, it will fail.

```

docs/writing-tests.md
```diff
@@ -0,0 +1,257 @@
...
+            group  <- randomGroup
+
+            _ <- produceMany(topic, kvs) // Comes from `KafkaTestUtils`. Produces messages to the topic.
+
+            records <- Consumer
+                         .plainStream(Subscription.Topics(Set(topic)), Serde.string, Serde.string)
+                         .take(5)
+                         .runCollect
+                         .provideSome[Kafka](
+                           // Comes from `KafkaTestUtils`
```
erikvanoosten, line 121: To make clear that this statement is about the `consumer` on line 122 I propose we add a colon:
```suggestion
                           // Comes from `KafkaTestUtils`:
```

docs/writing-tests.md
```diff
@@ -0,0 +1,257 @@
...
+This test is also a quite minimal example.    
+We produce 5 messages thanks to the `KafkaTestUtils.produceMany` method from zio-kafka-testkit, then we consume them with the `Consumer.plainStream` method from zio-kafka.     
+Finally, we use the `hasSameElements` assertion from zio-test to check that the consumed records are the same as the ones we produced.
+
+In this example, we're reusing the `producer` and the `Kafka.embedded` layers we've seen in the [Producer test example](#testing-a-producer).    
+We're also using the `KafkaTestUtils.consumer` layer from zio-kafka-testkit to instantiate a new `Consumer`.    
+
+Finally, we use the `KafkaRandom` trait from zio-kafka-testkit and its methods to generate random values for the Consumer client ID, the Consumer group ID and the topic name.     
+More details about this `KafkaRandom` trait [later in this page](#kafkarandom-trait).    
+Using random values for these parameters is important to avoid conflicts between tests as we share one Kafka instance between all the tests of the suite.
```
erikvanoosten, line 144: Saying we share a Kafka instance is enough, how is less important for this paragraph.

Also, as I have said before, I am not in favor of using random values. I prefer my test parameters to be the same for each run. (With a big exception for property based testing.)

Also, since this text is important, IMHO it should be a new paragraph.

```suggestion

Using different values for these parameters is important to avoid conflicts between tests as we share one Kafka instance between the tests. Using random values is a convenient way to make sure they are different.
```

docs/writing-tests.md
```diff
@@ -0,0 +1,257 @@
...
+   */
+  val saslEmbedded: ZLayer[Any, Throwable, Kafka.Sasl]
+
+  /**
+   * Creates an in-memory Kafka instance with a random port and SSL authentication configured.
+   */
+  val sslEmbedded: ZLayer[Any, Throwable, Kafka]
+}
+```
+
```
erikvanoosten, line 184:
```suggestion
The in-memory Kafka instances are created using [embedded-kafka](https://github.com/embeddedkafka/embedded-kafka).
```

zio-kafka-bench/src/main/scala/zio/kafka/bench/ConsumersComparisonBenchmark.scala
```diff
@@ -68,7 +68,7 @@ class ConsumersComparisonBenchmark extends ZioBenchmark[Env] {
         producer,
         settings,
         kafkaConsumer,
-        simpleConsumer()
+        minimalConsumer()
```
erikvanoosten, line 71: 👍

zio-kafka-example/src/main/scala/zio/kafka/example/Main.scala
erikvanoosten, line 1: What is the purpose of this file?

erikvanoosten (approved):
Nice!
Some small remarks but nothing exciting.

zio-kafka-example/src/main/scala/zio/kafka/example/Main.scala
guizmaii: It's a minimal example app. See https://github.com/zio/zio-kafka/pull/807

docs/writing-tests.md
guizmaii: That's already pretty explicit IMO

docs/writing-tests.md
guizmaii: I'd prefer to keep as is

docs/writing-tests.md
guizmaii: I prefer to keep these things grouped.

The paragraph is structured in 3 parts:
 1. Introduction to the concept we'll study in this paragraph: "Finally, we annotate the suite with the `timeout` and `sequential` aspects."   
 2. The first concept: "The `timeout` aspect from zio-test ..."
 3. The second concept: "The `sequential` aspect from zio-test ..."

docs/writing-tests.md
guizmaii: I think we're saying the same thing but differently

docs/writing-tests.md
erikvanoosten: That is true. And still, it took me 5 seconds to find out. Small things like this can take away friction.

docs/writing-tests.md
erikvanoosten: The original text uses the word 'necessary'. This is strictly not correct because there are other ways to achieve the same goal.

docs/writing-tests.md
erikvanoosten: That answer disappoints me a bit. I'd have preferred you tell me where I am wrong.

PR #858, opened 2023-05-23
[Internal] Use SLF4J/Logback to configure logging in tests

guizmaii:
This allows to more finely configure logging in tests


zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
```diff
@@ -551,7 +551,7 @@ object ConsumerSpec extends ZIOSpecDefault with KafkaRandom {
             Consumer
               .partitionedStream(subscription, Serde.string, Serde.string)
               .flatMapPar(Int.MaxValue) { case (tp, partStream) =>
-                val registerAssignment = ZStream.logInfo(s"Registering partition ${tp.partition()}") *>
+                val registerAssignment = ZStream.logDebug(s"Registering partition ${tp.partition()}") *>
```
guizmaii, line 554: I changed these to have less noisy CI results by default.

To get these logs back, you only need to change the `INFO` to `DEBUG` in the `zio-kafka-test/src/test/resources/logback.xml` file

erikvanoosten (approved):

PR #867, opened 2023-05-25
[Internal] Extract `Command` from the `Runloop` file and rename it to `RunloopCommand`

guizmaii:


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -469,24 +466,28 @@ private[consumer] final class Runloop private (
...
    */
-  def run: ZIO[Scope, Throwable, Any] =
+  def run: ZIO[Scope, Throwable, Any] = {
+    import Runloop.StreamOps
+
     ZStream
       .fromQueue(commandQueue)
       .timeoutFail[Throwable](RunloopTimeout)(runloopTimeout)
-      .takeWhile(_ != StopRunloop)
+      .takeWhile(_ ne RunloopCommand.StopRunloop)
```
erikvanoosten, line 475: In general `eq` and `ne` should only be used when reference equality is essential. That is not the case here. To validate its correctness you have to lookup what `StopRunLoop` is (it is an object, ah, okay, no problem here...).

```suggestion
      .takeWhile(_ != RunloopCommand.StopRunloop)
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -469,24 +466,28 @@ private[consumer] final class Runloop private (
...
       .runFoldChunksDiscardZIO(State.initial) { (state, commands) =>
         for {
-          _                  <- ZIO.logTrace(s"Processing ${commands.size} commands: ${commands.mkString(",")}")
-          stateAfterCommands <- ZIO.foldLeft(commands)(state)(handleCommand)
+          _ <- ZIO.logTrace(s"Processing ${commands.size} commands: ${commands.mkString(",")}")
+          streamCommands = commands.collect { case cmd: RunloopCommand.StreamControl => cmd }
```
erikvanoosten, line 479: The current implementation simply handles all commands, sometimes by doing nothing. This change will cause additional memory pressure. Can you elaborate a bit why this is desirable?

erikvanoosten (commented):
Runloop.scala is pretty big so moving Command out is a good idea. I am a bit worried about the additional memory allocation compared to the current implementation.

Have you considered importing `RunloopCommand._` in Runloop?

guizmaii:
> Have you considered importing `RunloopCommand._` in Runloop?

Yes, but I explicitly made the choice to always write `RunloopCommand.Xxx` to make it explicit that what we're dealing with is a command. It was annoying to me to have to remember that the `Request` type is a command, for example.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
guizmaii: I'm sorry but I'm not sure to understand your comment 🤔

I'd prefer to keep `ne`. It checks less things than `!=` so is faster.
Scalac complains if you misuse `ne` or `eq`

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
guizmaii: Here's how I see things:
1. Calling `handleCommand` many times for things that are not handled can create slowness (useless function call and pattern matching)
2. there's also some GC pressure as we have to instantiate a `ZIO.succeed(state)` for each non-handled command, and this `ZIO.succeed` will need to be interpreted by the ZIO runloop.

IMO, it's better to filter the list to only call it on things that need to be handled.

Also, it improves the typesafety of the `handleCommand` method, which only accepts a certain type of command now (`StreamControl`), so we don't have to deal with things we don't need to have to deal with.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: If StopRunLoop becomes a variable, reference equality may no longer be correct. One really should not use eq/ne frivolously. It will prevent problems later.

The performance gain is picoseconds at best. Compared to all the other things we do it is not worth talking about.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: CPUs are must faster then memory. So an additional function call or pattern matching has little influence on the total runtime. Adding more memory accesses though will make your program slower. Anyways, so I do not believe no. 1 to be true.

However, no, 2 is indeed a good point.

erikvanoosten (approved):

PR #879, opened 2023-05-29
Producer.produceChunkAsyncWithFailures

lukestephenson:
Similar to produceChunkAsync, but preserves all the failures in the chunk.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
@@ -10,6 +10,7 @@ import zio.stream.{ ZPipeline, ZStream }
 
 import java.util.concurrent.atomic.AtomicLong
 import scala.jdk.CollectionConverters._
+import scala.util.control.NonFatal
 
 trait Producer {
 
@@ -149,6 +150,26 @@ trait Producer {
   ): RIO[R, Task[Chunk[RecordMetadata]]]
 
   /**
+   * Produces a chunk of records. The effect returned from this method has two layers and describes the completion of
+   * two actions:
+   *   1. The outer layer describes the enqueueing of all the records to the Producer's internal buffer.
+   *   1. The inner layer describes receiving an acknowledgement from the broker for the transmission of the records.
+   *
+   * It is possible that for chunks that exceed the producer's internal buffer size, the outer layer will also signal
+   * the transmission of part of the chunk. Regardless, awaiting the inner layer guarantees the transmission of the
+   * entire chunk.
+   *
+   * This variant of `produceChunkAsync` more accurately reflects that individual records within the Chunk can fail to
+   * publish, rather than the failure being at the level of the Chunk.
+   *
+   * This variant does not accept serializers as they may also fail independently of each record and this is not
+   * reflected in the return type.
+   */
+  def produceChunkAsyncWithFailures(
+    records: Chunk[ProducerRecord[Array[Byte], Array[Byte]]]
+  ): UIO[UIO[Chunk[Either[Throwable, RecordMetadata]]]]
+
+  /**
    * Flushes the producer's internal buffer. This will guarantee that all records currently buffered will be transmitted
    * to the broker.
    */
@@ -161,7 +182,6 @@ trait Producer {
 }
 
 object Producer {
-
   val live: RLayer[ProducerSettings, Producer] =
     ZLayer.scoped {
       for {
@@ -198,7 +218,9 @@ object Producer {
     for {
       runtime <- ZIO.runtime[Any]
       sendQueue <-
-        Queue.bounded[(Chunk[ByteRecord], Promise[Throwable, Chunk[RecordMetadata]])](settings.sendBufferSize)
+        Queue.bounded[(Chunk[ByteRecord], Promise[Nothing, Chunk[Either[Throwable, RecordMetadata]]])](
+          settings.sendBufferSize
+        )
       producer = new ProducerLive(javaProducer, runtime, sendQueue)
       _ <- ZIO.blocking(producer.sendFromQueue).forkScoped
     } yield producer
@@ -291,6 +313,14 @@ object Producer {
     ZIO.serviceWithZIO[Producer](_.produceChunkAsync(records, keySerializer, valueSerializer))
 
   /**
+   * Accessor method for [[Producer.produceChunkAsyncWithFailures]]]
+   */
+  def produceChunkAsyncWithFailures(
+    records: Chunk[ProducerRecord[Array[Byte], Array[Byte]]]
+  ): RIO[Producer, UIO[Chunk[Either[Throwable, RecordMetadata]]]] =
+    ZIO.serviceWithZIO[Producer](_.produceChunkAsyncWithFailures(records))
+
+  /**
    * Accessor method
    */
   def produceChunk(
@@ -323,7 +353,7 @@ object Producer {
 private[producer] final class ProducerLive(
   private[producer] val p: JProducer[Array[Byte], Array[Byte]],
   runtime: Runtime[Any],
-  sendQueue: Queue[(Chunk[ByteRecord], Promise[Throwable, Chunk[RecordMetadata]])]
+  sendQueue: Queue[(Chunk[ByteRecord], Promise[Nothing, Chunk[Either[Throwable, RecordMetadata]]])]
 ) extends Producer {
 
   override def produce(record: ProducerRecord[Array[Byte], Array[Byte]]): Task[RecordMetadata] =
@@ -348,9 +378,9 @@ private[producer] final class ProducerLive(
   // noinspection YieldingZIOEffectInspection
   override def produceAsync(record: ProducerRecord[Array[Byte], Array[Byte]]): Task[Task[RecordMetadata]] =
     for {
-      done <- Promise.make[Throwable, Chunk[RecordMetadata]]
+      done <- Promise.make[Nothing, Chunk[Either[Throwable, RecordMetadata]]]
       _    <- sendQueue.offer((Chunk.single(record), done))
-    } yield done.await.map(_.head)
+    } yield done.await.flatMap(result => ZIO.fromEither(result.head))
 
   override def produceAsync[R, K, V](
     record: ProducerRecord[K, V],
@@ -382,13 +412,13 @@ private[producer] final class ProducerLive(
   override def produceChunkAsync(
     records: Chunk[ProducerRecord[Array[Byte], Array[Byte]]]
   ): Task[Task[Chunk[RecordMetadata]]] =
-    if (records.isEmpty) ZIO.succeed(ZIO.succeed(Chunk.empty))
-    else {
-      for {
-        done <- Promise.make[Throwable, Chunk[RecordMetadata]]
-        _    <- sendQueue.offer((records, done))
-      } yield done.await
-    }
+    produceChunkAsyncWithFailures(records).map(_.flatMap { chunkResults =>
+      val (errors, success) = chunkResults.partitionMap(identity)
+      errors.headOption match {
+        case Some(error) => ZIO.fail(error) // Only the first failure is returned.
+        case None        => ZIO.succeed(success)
+      }
+    })
 
   override def produceChunkAsync[R, K, V](
     records: Chunk[ProducerRecord[K, V]],
@@ -398,6 +428,18 @@ private[producer] final class ProducerLive(
     ZIO
       .foreach(records)(serialize(_, keySerializer, valueSerializer))
       .flatMap(produceChunkAsync)
+
+  // noinspection YieldingZIOEffectInspection
+  override def produceChunkAsyncWithFailures(
+    records: Chunk[ByteRecord]
+  ): UIO[UIO[Chunk[Either[Throwable, RecordMetadata]]]] =
+    if (records.isEmpty) ZIO.succeed(ZIO.succeed(Chunk.empty))
+    else {
+      for {
+        done <- Promise.make[Nothing, Chunk[Either[Throwable, RecordMetadata]]]
+        _    <- sendQueue.offer((records, done))
+      } yield done.await
+    }
 
   override def flush: Task[Unit] = ZIO.attemptBlocking(p.flush())
 
@@ -411,26 +453,25 @@ private[producer] final class ProducerLive(
     ZStream
       .fromQueueWithShutdown(sendQueue)
       .mapZIO { case (serializedRecords, done) =>
-        ZIO.attempt {
-          val it: Iterator[(ByteRecord, Int)] = serializedRecords.iterator.zipWithIndex
-          val res: Array[RecordMetadata]      = new Array[RecordMetadata](serializedRecords.length)
-          val count: AtomicLong               = new AtomicLong
-          val length                          = serializedRecords.length
+        ZIO.succeed {
+          try {
+            val it: Iterator[(ByteRecord, Int)] = serializedRecords.iterator.zipWithIndex
+            val res: Array[Either[Throwable, RecordMetadata]] =
+              new Array[Either[Throwable, RecordMetadata]](serializedRecords.length)
+            val count: AtomicLong = new AtomicLong
+            val length            = serializedRecords.length
 
-          while (it.hasNext) {
-            val (rec, idx): (ByteRecord, Int) = it.next()
+            while (it.hasNext) {
+              val (rec, idx): (ByteRecord, Int) = it.next()
 
-            val _ = p.send(
-              rec,
-              (metadata: RecordMetadata, err: Exception) =>
-                Unsafe.unsafe { implicit u =>
-                  exec {
-                    if (err != null) {
-                      exec {
-                        runtime.unsafe.run(done.fail(err)).getOrThrowFiberFailure()
-                      }
-                    } else {
-                      res(idx) = metadata
+              val _ = p.send(
+                rec,
+                (metadata: RecordMetadata, err: Exception) =>
+                  Unsafe.unsafe { implicit u =>
+                    exec {
+                      if (err != null) res(idx) = Left(err)
+                      else res(idx) = Right(metadata)
+
                       if (count.incrementAndGet == length) {
                         exec {
                           runtime.unsafe.run(done.succeed(Chunk.fromArray(res))).getOrThrowFiberFailure()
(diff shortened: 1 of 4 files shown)


zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -172,24 +209,23 @@ object Producer {
...
+                          runtime.unsafe.run(done.succeed(Chunk.fromArray(res))).getOrThrowFiberFailure(): Unit
                         }
                       }
                     }
                   }
               )
             }
           }
-            .foldCauseZIO(done.failCause, _ => ZIO.unit)
+            .foldCauseZIO(
```
lukestephenson, line 225: I didn't get to the bottom of what this is for, but appears to be an internal implementation failure, not something to expose. Anyway, if folks like the general idea of this PR, this will need further investigation.

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -94,6 +94,26 @@ trait Producer {
...
+   *
+   * This variant of `produceChunkAsync` more accurately reflects that individual records within the Chunk can fail to
+   * publish, rather than the failure being at the level of the Chunk.
+   *
+   * This variant does not accept serializers as they may also fail indepenently of each record and this is not
+   * reflected in the return type.
+   */
+  def produceChunkAsyncWithFailures(
+    records: Chunk[ProducerRecord[Array[Byte], Array[Byte]]]
+  ): Task[UIO[Chunk[Either[Throwable, RecordMetadata]]]]
```
lukestephenson, line 115: An alternative to the return type would be:
```
Task[UIO[Chunk[(ProducerRecord, Either[Throwable, RecordMetadata])]]]
```
So that each failure is immediately tied to the ProducerRecord. But currently I'm zipping the two together to get that context.

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -94,6 +94,26 @@ trait Producer {
     valueSerializer: Serializer[R, V]
   ): RIO[R, Task[Chunk[RecordMetadata]]]
 
+  /**
+   * Produces a chunk of records. The effect returned from this method has two layers and describes the completion of
+   * two actions:
+   *   1. The outer layer describes the enqueueing of all the records to the Producer's internal buffer. 2. The inner
+   *      layer describes receiving an acknowledgement from the broker for the transmission of the records.
```
guizmaii, line 101: FYI, I learnt recently that if you want scaladoc (and scalafmt) to keep you bullet points well formatted you need to write:
```scala
/**
  * 1. My first point
  * 1. My second point
  */
```

See https://contributors.scala-lang.org/t/do-numbered-lists-actually-work-in-scaladoc-markup/2748/2?u=guizmaii

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -94,6 +94,26 @@ trait Producer {
...
+   *
+   * It is possible that for chunks that exceed the producer's internal buffer size, the outer layer will also signal
+   * the transmission of part of the chunk. Regardless, awaiting the inner layer guarantees the transmission of the
+   * entire chunk.
+   *
+   * This variant of `produceChunkAsync` more accurately reflects that individual records within the Chunk can fail to
+   * publish, rather than the failure being at the level of the Chunk.
+   *
+   * This variant does not accept serializers as they may also fail indepenently of each record and this is not
+   * reflected in the return type.
```
guizmaii, line 111: Isn't this possible failure part of the `Either[Throwable, ...]`? 🤔

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
guizmaii: Looks like a good idea if people want to retry failed records, no? 🤔

svroonland:
What is a situation in which some of the records can fail to publish?

lukestephenson:
> What is a situation in which some of the records can fail to publish?

My understanding is that the level of failure in Kafka is a batch. Where as prior to this PR, zio-kafka represents the level of failure as a Chunk (where multiple Chunks could be part of one Kafka batch, or one Chunk could be split into multiple Kafka batches).

In the unit test that accompanies the PR, we have 3 records being published, 2 of these go a one topic and are valid records. 1 of them is for a compacted topic and does not have a key, which means it can't be published. So 2 succeed and 1 fail. Note that the 2 messages on the uncompacted topic are in a different batch than the one for the compacted topic (different topic partitions). If multiple records were published to the compacted topic, and only one of them was invalid (eg missing the key), then the others which are in the same batch would still fail (the brokers can't split the good messages from the bad ones in a batch).

Also note if we had a `Chunk(smallRecord, hugeRecord, smallRecord)` where the `hugeRecord` was too large, the `hugeRecord` would fail to publish, but the other records would still succeed (the others will be in separate batches because of the size limit for a single batch).

These examples we wouldn't want to retry, they would never succeed. But knowing which record couldn't be published might be useful to debug / report the failure for investigation.

Also imagine a scenario where a chunk has records for multiple topics. Only one of those topics may be in a bad state (eg under min ISR), in which case that might be a condition the user application may want to retry.

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
lukestephenson: The return type for a lot of these methods is something like `Task[Task[RecordMetadata]]`. If serialisation fails, that is typically represented on the failure of **Task[** Task[RecordMetadata] **]**.

So while it would be possible to shift that failure into the `Either` within the inner `Task`, I just find it pretty inconsistent that where that failure is represented depends on which method you use on this API. As per the docs on `produceAsync`:
> The outer layer describes the enqueueing of the record to the Producer's internal buffer.
That includes Serialization. IMO it's confusing if that affect then has it's failures represented somewhere else.

So agree it is possible, not sure how clearly it is conveyed when the failure types are `Throwable`

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
lukestephenson: No necessarily for retrying, might just be for recording more specific details about what failed.

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
lukestephenson: Fixed. Thanks

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
lukestephenson: Also, serializers allow an environment to be requested via ZIO. This would be on a different effect layer to where the failures are represented. I'd rather not expose an API with what I feel is a big inconsistency. What do you think?

lukestephenson:
PR is failing with:
```
HEAD is now at 8a3edc26 Merge e147b3908a3f64b3a7537aa6ce6a0002e27912a0 into 7188b660cf305c5d21f9e568751802146977a461
Error: Command 'git' failed with args '-c user.name=github-action-benchmark -c user.email=github@users.noreply.github.com -c http.https://github.com/.extraheader= push ***github.com/zio/zio-kafka.git gh-pages:gh-pages --no-verify': remote: Permission to zio/zio-kafka.git denied to github-actions[bot].
fatal: unable to access 'https://github.com/zio/zio-kafka.git/': The requested URL returned error: 403
```

guizmaii:
@lukestephenson What's the status of your PR?

lukestephenson:
> @lukestephenson What's the status of your PR?

I've just merged master in, looks like some linting checks are now failing which I should be able to fix shortly. 

The main outstanding task for the PR is confirmation that other contributors are happy with an API being exposed that works on `ProducerRecord[Array[Byte], Array[Byte]]` rather than using the Serializer approach that is on all the other APIs. Just to clarify again because it's been a while, this is because the Serializers can return a failure and I didn't want to have to reflect this on the new API.

lukestephenson:
Looking into the test failures

lukestephenson:
Looks like I'm hitting a scala 3.3.0 bug. Still investigating, but the tests are fine on other Scala versions, and failing with the following on 3.3.0:
```
Exception in thread "zio-fiber-34179" scala.MatchError: (producerspec-client-c91742ad-3d36-4f4f-af19-429da1aa8ac4,boo,baa,null,boo,hello,world,Chunk(ProducerRecord(topic=producerspec-topic-8377fcd5-1a4c-4630-8414-6da68f327ee5, partition=null, headers=RecordHeaders(headers = [], isReadOnly = false), key=[B@7d3d62ac, value=[B@1bf67c5d, timestamp=null),ProducerRecord(topic=producerspec-topic-4a923490-2b1e-4f12-923e-10aebb958485, partition=null, headers=RecordHeaders(headers = [], isReadOnly = false), key=null, value=[B@4133c4e9, timestamp=null),ProducerRecord(topic=producerspec-topic-8377fcd5-1a4c-4630-8414-6da68f327ee5, partition=null, headers=RecordHeaders(headers = [], isReadOnly = false), key=[B@7dbde679, value=[B@6b83d360, timestamp=null))) (of class scala.Tuple8)
    	at zio.kafka.ProducerSpec$.$anonfun$5$$anonfun$1$$anonfun$1$$anonfun$2$$anonfun$1$$anonfun$2(ProducerSpec.scala:118)
    	at zio.kafka.ProducerSpec.spec(ProducerSpec.scala:118)
    	at zio.kafka.ProducerSpec.spec(ProducerSpec.scala:126)
```

lukestephenson:
ok. Tests are now passing. Appeared to hit a bug in scala 3.3.0 or some other compiler plugin that is being used.

Benchmarks / profiling actions are failing. It doesn't look like they have the required permissions to work on a github fork.

guizmaii:
> Benchmarks / profiling actions are failing. It doesn't look like they have the required permissions to work on a github fork.

Yes, they don't pass on forks' PRs. We might be able to fix this but no idea how 🤔

lukestephenson:
Hello. I've just merged the latest changes into my PR. I'm still keen for this functionality. Thanks

guizmaii:
@lukestephenson I'll have a look this weekend 🙂

guizmaii:
@lukestephenson 
> The main outstanding task for the PR is confirmation that other contributors are happy with an API being exposed that works on `ProducerRecord[Array[Byte], Array[Byte]]` rather than using the Serializer approach that is on all the other APIs.

We've recently added new method to the Producer that operate on `ProducerRecord[Array[Byte], Array[Byte]]` so no issue with that.

> If multiple records were published to the compacted topic, and only one of them was invalid (eg missing the key), then the others which are in the same batch would still fail (the brokers can't split the good messages from the bad ones in a batch).

(Do we need to add|Do you have) a unit test for this case?

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -411,24 +453,23 @@ private[producer] final class ProducerLive(
...
                   }
                 }
             )
           }
         }
-          .foldCauseZIO(done.failCause, _ => ZIO.unit)
+          .foldCauseZIO(
+            cause => ZIO.dieMessage(cause.toString),
+            _ => ZIO.unit
+          ) // TODO how should these failures be represented
```
guizmaii, line 472: @lukestephenson What about this TODO?

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
guizmaii: Why didn't you go this way in the end?

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
guizmaii: > This would be on a different effect layer to where the failures are represented. I'd rather not expose an API with what I feel is a big inconsistency. What do you think?

Agreed. I'm fine with the current version taking `ProducerRecord[Array[Byte], Array[Byte]]`

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -411,24 +453,23 @@ private[producer] final class ProducerLive(
...
                       }
                     }
                   }
                 }
             )
           }
         }
-          .foldCauseZIO(done.failCause, _ => ZIO.unit)
+          .foldCauseZIO(
+            cause => ZIO.dieMessage(cause.toString),
```
guizmaii, line 470: Well, if we `die` here, we'd better use `.map` over `.mapZIO`, DYT?
```scala
ZStream
      .fromQueueWithShutdown(sendQueue)
      .map {
        ...
      }
```

guizmaii (commented):
@lukestephenson See comments and questions, especially this one please: https://github.com/zio/zio-kafka/pull/879/files#r1284996440

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
svroonland: How about instead of the `foldCause`:

```scala
.catchAll { e =>
  done.succeed(serializedRecords.map(_ => Left(e)))
}
```

If something unrecoverable happens, it's a programming error and we just let  the Producer die.

svroonland (commented):
Looks good, have a look at my suggestion for the todo

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
guizmaii: @svroonland @lukestephenson Here's what I now understand about the previous version of the code:

`p.send` returns a `Future` but can also throw. 
So there are 2 channels of error in this function call.
If we have a quick look at the `KafkaProducer` code:
```java
    /**
     * Implementation of asynchronously send a record to a topic.
     */
    private Future<RecordMetadata> doSend(ProducerRecord<K, V> record, Callback callback) {
        ...
        
        try {
           ...
           
            // handling exceptions and record the errors;
            // for API exceptions return them in the future,
            // for other exceptions throw directly
        } catch (ApiException e) {
            log.debug("Exception occurred during message send:", e);
            if (callback != null) {
                TopicPartition tp = appendCallbacks.topicPartition();
                RecordMetadata nullMetadata = new RecordMetadata(tp, -1, -1, RecordBatch.NO_TIMESTAMP, -1, -1);
                callback.onCompletion(nullMetadata, e);
            }
            this.errors.record();
            this.interceptors.onSendError(record, appendCallbacks.topicPartition(), e);
            if (transactionManager != null) {
                transactionManager.maybeTransitionToErrorState(e);
            }
            return new FutureFailure(e);
        } catch (InterruptedException e) {
            this.errors.record();
            this.interceptors.onSendError(record, appendCallbacks.topicPartition(), e);
            throw new InterruptException(e);
        } catch (KafkaException e) {
            this.errors.record();
            this.interceptors.onSendError(record, appendCallbacks.topicPartition(), e);
            throw e;
        } catch (Exception e) {
            // we notify interceptor about all exceptions, since onSend is called before anything else in this method
            this.interceptors.onSendError(record, appendCallbacks.topicPartition(), e);
            throw e;
        }
    }
```

The first "error channel" (when `Future` contains a failure) was handled by our callback (`if err != null`) and is still handled in the same way by our callback in the new code.

The diff is in the handling of the second "error channel" (when `p.send` throws).
In the previous code, we were recovering this error, preventing the `sendFromQueue` IO to fail (and so probably preventing the `Producer` to fail):
```scala
ZIO
  .attempt {
    ...
  }
  .foldCauseZIO(done.failCause, _ => ZIO.unit)
```
`done.failCause` returning `UIO[Boolean]`, this `ZIO.attempt.foldCauseZIO` is a `UIO[Unit]`
We were transmitting the thrown exception to the code calling our Producer thanks to this `done` promise. 

In the new code, in its current version, we `.die` the `sendFromQueue` IO with the exception thrown.
(The currently implemented dying mechanism is IMO incorrect but that's out of context of my reflection here)
So we're not transmitting anymore the thrown exception to the caller, or at least not in the same way.
Before we were failing the `Producer.produce` call while now, we're probably failing the `Producer` layer by failing the `sendFromQueue` IO.

If what I said until now isn't incorrect, if we want to preserve the previous behaviour, we should do something like this:
```scala
    ZStream
      .fromQueueWithShutdown(sendQueue)
      .mapZIO {
        // I could just have used `ZIO.attempt.foldCauseZIO` but it's probably a bit less performant.
        // The goal here is to use only 1 IO
        ZIO.succeed {
          try {
            ...  // previous code
          } catch {
              case NonFatal(e) =>
                Unsafe.unsafe { implicit u =>
                  runtime.unsafe.run(done.succeed(Chunk.single(Left(e)))).getOrThrowFiberFailure()
                }
            }
        }
      } // we don't need the `.foldCauseZIO` anymore
```

WDYT?

svroonland: Quite similar to what I proposed I believe, but perhaps more semantically correct to succeed the promise with a chunk of N times `Left(e)`, where `N` is the number of records.

guizmaii: > perhaps more semantically correct to succeed the promise with a chunk of N times Left(e), where N is the number of records.

Not wrong 🙂

lukestephenson: Thanks for this suggestion. I'll have a follow up commit shortly.

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
lukestephenson: The end contract is:
```
  override def produceChunkAsyncWithFailures(
    records: Chunk[ByteRecord]
  ): UIO[UIO[Chunk[Either[Throwable, RecordMetadata]]]]
```

I didn't zip together with the ProducerRecord as doing so is extra overhead that not everyone may need. But happy to change if it is valued more widely.

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
lukestephenson: I went with the suggestion in https://github.com/zio/zio-kafka/pull/879#discussion_r1285061537

lukestephenson:
Thanks all for the feedback. I believe I've covered it.

> > If multiple records were published to the compacted topic, and only one of them was invalid (eg missing the key), then the others which are in the same batch would still fail (the brokers can't split the good messages from the bad ones in a batch).
> 
> [Do we need to add|Do you have] a unit test for this case?

I do have a test for good and bad messages in a Chunk provided to our API. I don't have a test specifically for a Kafka batch with good and bad messages, as that doesn't affect our implementation. IMO we don't need a test for this.

guizmaii (approved):
LGTM

@lukestephenson Can you rebase your PR and fix the linting, please? 🙂

lukestephenson:
Rebased and linting applied

guizmaii:
Thanks for you work and patience @lukestephenson 🙂

PR #880, opened 2023-05-29
stopConsumption if called before the start of the stream start does nothing

guizmaii:


zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
```diff
@@ -1129,9 +1129,61 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
...
+                  SubscriptionFinalized,
+                  SubscriptionFinalized,
+                  RunloopFinalized,
+                  ConsumerFinalized
+                )
+              )
             )
-          }
+          },
+          test("toto - 1") {
```
guizmaii, line 1142: TODO @guizmaii Fix naming

PR #889, opened 2023-06-02
Add new `ConsumerSettings::enableOptimisticResume` setting (`true` by default) to allow users to disable the "optimistic resume" optimisation && Write benchmarks where this optimisation is disabled

guizmaii:
### Before
```scala
zioKafkaBench/Jmh/run -wi 10 -i 10 -r 1 -w 1 -t 1 -f 3 -foe true .*ConsumersComparisonBenchmark*
...
[info] Benchmark                                        Mode  Cnt     Score     Error  Units
[info] ConsumersComparisonBenchmark.kafkaClients        avgt   30   645.361 ±   3.004  ms/op
[info] ConsumersComparisonBenchmark.manualKafkaClients  avgt   30   641.042 ±   2.877  ms/op
[info] ConsumersComparisonBenchmark.manualZioKafka      avgt   30  2651.134 ± 100.297  ms/op
[info] ConsumersComparisonBenchmark.zioKafka            avgt   30   739.566 ±  21.019  ms/op
```

### After
```scala
zioKafkaBench/Jmh/run -wi 10 -i 10 -r 1 -w 1 -t 1 -f 3 -foe true .*comparison.*
...
[info] Result "zio.kafka.bench.comparison.KafkaClientBenchmarks.kafkaClients":
[info]   641.438 ±(99.9%) 3.349 ms/op [Average]
[info]   (min, avg, max) = (624.941, 641.438, 650.032), stdev = 5.012
[info]   CI (99.9%): [638.090, 644.787] (assumes normal distribution)
...
[info] Result "zio.kafka.bench.comparison.KafkaClientBenchmarks.manualKafkaClients":
[info]   642.514 ±(99.9%) 3.090 ms/op [Average]
[info]   (min, avg, max) = (632.069, 642.514, 656.932), stdev = 4.625
[info]   CI (99.9%): [639.424, 645.605] (assumes normal distribution)
...
[info] Result "zio.kafka.bench.comparison.ZioKafkaBenchmarks.manualZioKafka":
[info]   2605.392 ±(99.9%) 229.199 ms/op [Average]
[info]   (min, avg, max) = (1584.064, 2605.392, 2818.016), stdev = 343.055
[info]   CI (99.9%): [2376.193, 2834.592] (assumes normal distribution)
...
[info] Result "zio.kafka.bench.comparison.ZioKafkaBenchmarks.zioKafka":
[info]   734.987 ±(99.9%) 22.830 ms/op [Average]
[info]   (min, avg, max) = (683.740, 734.987, 826.528), stdev = 34.170
[info]   CI (99.9%): [712.158, 757.817] (assumes normal distribution)
...
[info] Result "zio.kafka.bench.comparison.ZioKafkaNoOptimisticResumeBenchmarks.manualZioKafka":
[info]   819.937 ±(99.9%) 120.082 ms/op [Average]
[info]   (min, avg, max) = (683.373, 819.937, 1243.105), stdev = 179.733
[info]   CI (99.9%): [699.855, 940.019] (assumes normal distribution)
...
[info] Result "zio.kafka.bench.comparison.ZioKafkaNoOptimisticResumeBenchmarks.zioKafka":
[info]   1311.588 ±(99.9%) 212.646 ms/op [Average]
[info]   (min, avg, max) = (755.915, 1311.588, 2329.325), stdev = 318.279
[info]   CI (99.9%): [1098.942, 1524.234] (assumes normal distribution)
...
[info] Benchmark                                            Mode  Cnt     Score     Error  Units
[info] KafkaClientBenchmarks.kafkaClients                   avgt   30   641.438 ±   3.349  ms/op
[info] KafkaClientBenchmarks.manualKafkaClients             avgt   30   642.514 ±   3.090  ms/op
[info] ZioKafkaBenchmarks.manualZioKafka                    avgt   30  2605.392 ± 229.199  ms/op
[info] ZioKafkaBenchmarks.zioKafka                          avgt   30   734.987 ±  22.830  ms/op
[info] ZioKafkaNoOptimisticResumeBenchmarks.manualZioKafka  avgt   30   819.937 ± 120.082  ms/op
[info] ZioKafkaNoOptimisticResumeBenchmarks.zioKafka        avgt   30  1311.588 ± 212.646  ms/op
```

### Observations

Interestingly, the optimistic resume optimisation seems to be very detrimental to the manual consumption performances


zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -21,15 +21,16 @@ import zio.kafka.security.KafkaCredentialStore
...
   bootstrapServers: List[String],
   properties: Map[String, AnyRef],
   closeTimeout: Duration,
   pollTimeout: Duration,
   offsetRetrieval: OffsetRetrieval = OffsetRetrieval.Auto(),
   rebalanceListener: RebalanceListener = RebalanceListener.noop,
   restartStreamOnRebalancing: Boolean = false,
-  runloopTimeout: Duration = ConsumerSettings.defaultRunloopTimeout
+  runloopTimeout: Duration = ConsumerSettings.defaultRunloopTimeout,
+  disableOptimisticResume: Boolean = false // By default, the "optimistic resume" optimisation is enabled
```
erikvanoosten, line 33: I prefer positive wording to prevent double negations:

```suggestion
  enableOptimisticResume: Boolean = true
```

or better yet:

```suggestion
  optimisticResume: Boolean = true
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -241,11 +242,15 @@ private[consumer] final class Runloop private (
...
-    val resumeTps = new java.util.ArrayList[TopicPartition](streams.size)
-    val pauseTps  = new java.util.ArrayList[TopicPartition](streams.size)
-    streams.foreach { stream =>
-      val tp       = stream.tp
-      val toResume = requestedPartitions.contains(tp) || stream.optimisticResume
+    val resumeTps               = new java.util.ArrayList[TopicPartition](streams.size)
+    val pauseTps                = new java.util.ArrayList[TopicPartition](streams.size)
+    val optimisticResumeEnabled = !consumerSettings.disableOptimisticResume
+    val iterator                = streams.iterator
+    while (iterator.hasNext) {
```
erikvanoosten, line 249: If you look at `foreach` it does exactly this: get the iterator and use it in a while loop. I do not think this change will make the program even a nanosecond faster (the JVM will inline this anyway) and for sure it is less nice to read.

erikvanoosten (commented):
There is no reason to inline `Chunk.foreach`, otherwise this PR LGTM.

BTW, I didn't check the benchmark related code in detail.

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
guizmaii: Changed. Thanks!

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
guizmaii: I rollbacked to `foreach` :)

guizmaii:
> BTW, I didn't check the benchmark related code in detail.

I just re-organised and copy/pasted the already existing benchmarks. 
Its the only way to be able to have different sets of settings if we want to keep the instantiation of the Consumer outside of the benchmarked code (which I absolutely want)

erikvanoosten (approved):
I am still wondering if we should call the setting `optimisticResume` instead of `enableOptimisticResume` 🤷

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -21,15 +21,16 @@ import zio.kafka.security.KafkaCredentialStore
  *   subscribed for long periods during its lifetime, this timeout should take that into account as well. When the
  *   timeout expires, the plainStream/partitionedStream/etc will fail with a [[Consumer.RunloopTimeout]].
  */
```
erikvanoosten, line 23: Suggestion:
```
@param enableOptimisticResume
  Zio-kafka pauses partitions for which its associated stream does not require data. When the
 stream needs more data, the partition is resumed. When optimistic resume is enabled (the default)
 zio-kafka predicts whether a stream needs more data in the _next_ poll. By resuming the partition
 slightly ahead of time, the stream does not need to wait. This lowers latency and can double
throughput. You can disable this feature when latency and throughput are not a concern and your
application is tight on memory and CPU.
```

guizmaii:
> I am still wondering if we should call the setting optimisticResume instead of enableOptimisticResume 🤷

`enableOptimisticResume` is more explicit

svroonland (approved):
Good for users to be able to toggle this feature to see the effect on performance

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
guizmaii: I don't want to be too specific for now. 
Let's work with the users who are reporting issues. 
Let's see what is the effect of disabling this optimisation on their workload, and then later, we will document this option properly if needed.

erikvanoosten: Less specific:
```
@param enableOptimisticResume
  Zio-kafka pauses partitions for which its associated stream does not require data. When the
 stream needs more data, the partition is resumed. When optimistic resume is enabled (the default)
 zio-kafka predicts whether a stream needs more data in the _next_ poll. By resuming the partition
 slightly ahead of time, the stream does not need to wait. This can lower latency and can increase
throughput.
```

erikvanoosten: See https://github.com/zio/zio-kafka/pull/894

PR #893, opened 2023-06-03
[Testkit] Allow to customise broker properties

V-Lamp:


zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
```diff
@@ -320,6 +322,19 @@ object KafkaTestUtils {
...
+   * To be used together with [[withAdmin]], [[withSaslAdmin]], [[withSslAdmin]] or [[withAdminClient()]]. Useful for
+   * when you set "auto.create.topics.enable" -> "false" in your Kafka brokers.
+   */
+  def createTopics(
+    adminClient: AdminClient,
+    topics: Iterable[String],
+    partitions: Int = 1,
+    replicationFactor: Short = 1,
+    configs: Map[String, String] = Map.empty
+  ): Task[Unit] =
```
guizmaii, line 335: Shouldn't the `AdminClient` requirement be passed through the environment:
```scala
  def createTopics(
    topics: Iterable[String],
    partitions: Int = 1,
    replicationFactor: Short = 1,
    configs: Map[String, String] = Map.empty
  ): RIO[AdminClient, Unit] = ZIO.serviceWithZIO[AdminClient](_.createTopics(topics.map(NewTopic(_, partitions, replicationFactor, configs))))
```

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/Kafka.scala
```diff
@@ -115,17 +111,10 @@ object Kafka {
...
-  private def startKafka(makeConfig: (Int, Int) => EmbeddedKafkaConfig): Task[EmbeddedK] =
-    ZIO.attemptBlocking {
-      try {
-        val (k, z) = nextPorts
-        EmbeddedKafka.start()(makeConfig(k, z))
-      } catch {
-        case NonFatal(e) => throw EmbeddedKafkaStartException("Failed to start embedded Kafka", e)
-      }
-    }
+  case class Ports(kafkaPort: Int, zookeeperPort: Int)
```
guizmaii, line 114:
```suggestion
  final case class Ports(kafkaPort: Int, zookeeperPort: Int)
```

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/Kafka.scala
```diff
@@ -105,6 +78,29 @@ object Kafka {
 
   final case class Sasl(value: Kafka) extends AnyVal
 
+  /**
+   * Creates an in-memory Kafka instance with a random port and configured with the provided kafka broker properties.
+   */
+  def withBrokerProps(brokerProps: Ports => Map[String, String]): ZLayer[Any, Throwable, Kafka] =
```
guizmaii, line 84: IMO, this should be private

```suggestion
  private def withBrokerProps(brokerProps: Ports => Map[String, String]): ZLayer[Any, Throwable, Kafka] =
```

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
```diff
@@ -320,6 +322,19 @@ object KafkaTestUtils {
       AdminClient.make(settings).flatMap(f)
     }
 
+  /**
+   * To be used together with [[withAdmin]], [[withSaslAdmin]], [[withSslAdmin]] or [[withAdminClient()]]. Useful for
+   * when you set "auto.create.topics.enable" -> "false" in your Kafka brokers.
```
guizmaii, line 327: IMO, this documentation does not document what the function is for, nor what it does.
You're describing your issue here, not the function behaviour and utility.

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/Kafka.scala
```diff
@@ -105,6 +78,29 @@ object Kafka {
...
+      for {
+        ports <- nextPorts
+        embeddedKafkaConfig = EmbeddedKafkaConfig(
+                                ports.kafkaPort,
+                                ports.zookeeperPort,
+                                brokerProps(ports)
+                              )
+        kafka <- ZIO.acquireRelease(
+                   ZIO
+                     .attempt(EmbeddedKafka.start()(embeddedKafkaConfig))
```
guizmaii, line 95:
```suggestion
                     .attemptBlocking(EmbeddedKafka.start()(embeddedKafkaConfig))
```

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/Kafka.scala
```diff
@@ -105,6 +78,29 @@ object Kafka {
...
+                     .map(EmbeddedKafkaService)
```
guizmaii, line 96:
```suggestion
                     .attemptBlocking(EmbeddedKafkaService(EmbeddedKafka.start()(embeddedKafkaConfig)))
```

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/Kafka.scala
```diff
@@ -105,6 +78,29 @@ object Kafka {
...
+                     .catchSome { case NonFatal(e) =>
+                       throw EmbeddedKafkaStartException("Failed to start embedded Kafka", e)
+                     }
```
guizmaii, line 99: You shouldn't throw here but return a `ZIO.fail`:
```suggestion
                     .catchSomeZIO { case NonFatal(e) =>
                       ZIO.fail(EmbeddedKafkaStartException("Failed to start embedded Kafka", e))
                     }
```

guizmaii (commented):
See comments :)

erikvanoosten (commented):
LGTM. Jules comments need to be addressed though.

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
guizmaii: Additional note:
Wouldn't it be more modular to only give a `createTopic` method:
```scala
def createTopic(
  topics: String,
  partitions: Int = 1,
  replicationFactor: Short = 1,
  configs: Map[String, String] = Map.empty
): RIO[AdminClient, Unit] = ZIO.serviceWithZIO[AdminClient](_.createTopics(NewTopic(topic, partitions, replicationFactor, configs)))
```

The usage would be:
```scala
test("myTest") {
  val topics: List[String] = ...
  for {
    _ <- topics.forEach(createTopic(_))
   ...
   yield ...
}
```

So that the user can still create as many topics as he/she needs but also can tune the each Topic `partitions`, `replicationFactor` and `configs` individually.

IMO, this `createTopic` (without the `s`) version would me generic and useful in more cases while still providing the same capabilities.

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/Kafka.scala
erikvanoosten: What Jules said, but like this (catchSomeZIO doesn't exist):

```suggestion
                     .catchNonFatalOrDie { e =>
                       ZIO.fail(EmbeddedKafkaStartException("Failed to start embedded Kafka", e))
                     }
```

guizmaii: Yeah I wasn't sure it exists 😄

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
V-Lamp: I agree, I just wanted to make it work better with the existing methods that are all shaped like `[T](f: AdminClient => RIO[Kafka, T]): ZIO[Kafka, Throwable, T]`

I removed this as I think a bigger change is needed:
- Accessor methods on AdminClient in the core project
- `KafkaUtils.sslAdminSettings`, `KafkaUtils.adminSettings` to be `Layer[Kafka, ..., AdminClient]`

What is the appetite for some breaking changes?

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/Kafka.scala
```diff
@@ -19,83 +19,56 @@ object Kafka {
...
-          "group.initial.rebalance.delay.ms" -> "0",
-          "authorizer.class.name"            -> "kafka.security.authorizer.AclAuthorizer",
-          "super.users"                      -> "User:ANONYMOUS"
-        )
-      )
-
-    ZIO.acquireRelease(
-      startKafka(embeddedKafkaConfig).map(EmbeddedKafkaService.apply)
-    )(_.stop())
+  val embedded: ZLayer[Any, Throwable, Kafka] = withBrokerProps { _ =>
```
V-Lamp, line 22: I suggest updating these 3 layers to `def embeddedX(withProperties: Ports => Map[String, String] = Map.emtpy)`

This is a (slightly) breaking change as it will need to add `()` at the end of existing calls.

But the benefit is that it will allow the user to only add/update the properties they want.

For example, `Kafka.embedded(_ => Map("auto.create.topics.enable" -> "false"))`

guizmaii:
@V-Lamp Can you rebase your branch on the `master` branch, please?
We've fix some CI job definitions there

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
guizmaii: > Accessor methods on AdminClient companion object in the core project

I'd prefer we avoid to add these functions. It's a burden for us to maintain (more code), and it just replaces `ZIO.serviceWithZIO[AdminClient](_.myFunctionCall)` which IMO not enough to justify this burden

> KafkaUtils.sslAdminSettings, KafkaUtils.adminSettings to be Layer[Kafka, ..., AdminClient]

I'm not sure to understand, sorry. 
Why would `KafkaUtils.sslAdminSettings` return an `AdminClient`?

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/Kafka.scala
guizmaii: I disagree with replacing it, but I'm not necessarily against adding a new variant which could be:
```scala
def embeddedWith(properties: Ports => Map[String, String])
```

V-Lamp: Added the 3 `def embeddedWith(properties: Ports => Map[String, String])` methods and made the helper method private

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/Kafka.scala
```diff
@@ -105,6 +106,29 @@ object Kafka {
 
   final case class Sasl(value: Kafka) extends AnyVal
 
+  private def embeddedWithBrokerProps(
+    customProps: Ports => Map[String, String],
+    presetProps: Ports => Map[String, String]
```
guizmaii, line 111: Sorry to be annoying but can we inverse these two parameters. IMO, it's makes it more explicit that `customProps` overrides the `presetProps` props

guizmaii (approved):
LGTM

Thanks for your contributions! ❤️

PR #899, opened 2023-06-04
Improve benchmarks

guizmaii:


zio-kafka-bench/src/main/scala/zio/kafka/bench/comparison/KafkaClientBenchmarks.scala
```diff
@@ -45,15 +59,28 @@ class KafkaClientBenchmarks extends ComparisonBenchmark {
...
-              var count = 0L
+              var count                                            = 0L
+              var first: ConsumerRecord[Array[Byte], Array[Byte]]  = null
+              var lasts: ConsumerRecords[Array[Byte], Array[Byte]] = null
               while (count < numberOfMessages) {
                 val records = consumer.poll(settings.pollTimeout)
                 count += records.count()
+                if (count == 0L) {
+                  first = records.iterator().next()
+                }
```
erikvanoosten, line 70: That won't work. When count == 0 records will be empty so next will fail with an exception.

.scalafmt.conf
erikvanoosten, line 1: Please do format changes in a separate PR.

.scalafmt.conf
```diff
@@ -1,21 +1,25 @@
-version=3.7.4
-project.git = true
-maxColumn = 120
-align {
-  preset = more
-  arrowEnumeratorGenerator = true
-}
+version = "3.7.3"
+runner.dialect = Scala213Source3 # https://scalameta.org/scalafmt/docs/configuration.html#scala-2-with--xsource3
+maxColumn = 140
```
erikvanoosten, line 3: I have a 14" screen, more than 120 is not practical.

.scalafmt.conf
guizmaii: It's still a draft PR

erikvanoosten: Understood, I'll stay away from drafts!

PR #909, opened 2023-06-07
Fix: Runloop thread name is always `zio-kafka-runloop-thread-0`

guizmaii:
Needs:
- https://github.com/zio/zio-kafka/pull/873


erikvanoosten (approved):
Perhaps we should have used a zio.Ref instead of an AtomicLong (btw, a Long?? how long do we want to run this program, an Int should be way more then enough...) 😄

PR #857, opened 2023-05-23
Using a Consumer without consuming shouldn't throw a `RunloopTimeout` exception

guizmaii:
Fixes #856 

We need this PR https://github.com/zio/zio-sbt/pull/220 to merged and a new release of zio-sbt to be made to fix the CI.

The solution implemented is to boot the Runloop lazily: we only start it when the user start a consuming session

### Before:
```scala
[info] Benchmark                                        Mode  Cnt     Score    Error  Units
[info] ConsumerBenchmark.throughput                     avgt   30   205.017 ±  6.977  ms/op
[info] ConsumerBenchmark.throughputWithCommits          avgt   30   284.384 ±  8.733  ms/op
[info] ConsumersComparisonBenchmark.kafkaClients        avgt   30   647.721 ±  2.320  ms/op
[info] ConsumersComparisonBenchmark.manualKafkaClients  avgt   30   641.767 ±  2.508  ms/op
[info] ConsumersComparisonBenchmark.manualZioKafka      avgt   30  2673.759 ± 89.569  ms/op
[info] ConsumersComparisonBenchmark.zioKafka            avgt   30   737.250 ± 22.264  ms/op
```

### After:
```scala
[info] Benchmark                                        Mode  Cnt     Score     Error  Units
[info] ConsumerBenchmark.throughput                     avgt   30   209.078 ±   9.219  ms/op
[info] ConsumerBenchmark.throughputWithCommits          avgt   30   284.131 ±   9.911  ms/op
[info] ConsumersComparisonBenchmark.kafkaClients        avgt   30   645.890 ±   3.103  ms/op
[info] ConsumersComparisonBenchmark.manualKafkaClients  avgt   30   642.727 ±   2.079  ms/op
[info] ConsumersComparisonBenchmark.manualZioKafka      avgt   30  1686.490 ± 219.173  ms/op
[info] ConsumersComparisonBenchmark.zioKafka            avgt   30   722.776 ±  38.728  ms/op
```
### Conclusions

Interestingly and unexpectedly it's making the manual consumption with zio-kafka kind of way faster 🤔

Here are the details about these changes in `manualZioKafka`:
Before:
```scala
sbt:zio-kafka> zioKafkaBench/Jmh/run -wi 10 -i 10 -r 1 -w 1 -t 1 -f 5 -foe true .*ConsumersComparisonBenchmark.manualZioKafka*
...
[info] Result "zio.kafka.bench.ConsumersComparisonBenchmark.manualZioKafka":
[...]

PR #910, opened 2023-06-07
Use a Promise in the `RunloopState.Started` as suggested by @erikvanoosten

guizmaii:
See https://github.com/zio/zio-kafka/pull/857#discussion_r1218434608

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
@@ -7,13 +7,24 @@ import zio.kafka.consumer.internal.Runloop.ByteArrayCommittableRecord
 import zio.kafka.consumer.internal.RunloopAccess.PartitionAssignment
 import zio.kafka.consumer.{ ConsumerSettings, Subscription }
 import zio.stream.{ Stream, Take, UStream, ZStream }
-import zio.{ durationInt, Hub, Ref, Scope, UIO, ZIO, ZLayer }
+import zio.{ Hub, Promise, Ref, Scope, UIO, URIO, ZIO, ZLayer }
 
 private[internal] sealed trait RunloopState
 private[internal] object RunloopState {
-  case object NotStarted                     extends RunloopState
-  final case class Started(runloop: Runloop) extends RunloopState
-  case object Stopped                        extends RunloopState
+
+  /**
+   * Why do we need a Promise here?
+   *
+   * If the user starts a forked consumption session and just after the fork, calls `consumer.stopConsumption`, the
+   * consumption needs to be stopped even if the runloop is still booting up.
+   *
+   * For all the details, see discussion: https://github.com/zio/zio-kafka/pull/857#discussion_r1218434608
+   */
+  final case class Started(promise: Promise[Nothing, Runloop]) extends RunloopState {
+    @inline def runloop: UIO[Runloop] = promise.await
+  }
+  case object NotStarted extends RunloopState
+  case object Stopped    extends RunloopState
 }
 
 /**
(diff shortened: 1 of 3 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopExecutor.scala
```diff
@@ -9,16 +9,16 @@ private[consumer] object RunloopExecutor {
...
+      ZIO.succeed {
         val javaExecutor =
           Executors.newSingleThreadExecutor(runnable => new Thread(runnable, s"zio-kafka-runloop-thread-$i"))
 
         Executor.fromJavaExecutor(javaExecutor) -> javaExecutor
       }
     } { case (_, executor) => ZIO.attempt(executor.shutdown()).orDie }.map(_._1)
 
-  val newInstance: ZIO[Scope, Throwable, Executor] = newSingleThreadedExecutor(counter.getAndIncrement())
+  val newInstance: URIO[Scope, Executor] = newSingleThreadedExecutor(counter.getAndIncrement())
```
guizmaii, line 22: We need this to be a `UIO` so the `Runloop.make` is an UIO to, so everything is simpler in the `RunloopAccess` code. 
The `newSingleThreadExecutor` cannot fail anyway. 
If it does, it's not something we should handle. We should let things die.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
```diff
@@ -37,40 +49,28 @@ private[internal] object RunloopState {
...
+  makeRunloop: UIO[Runloop],
   diagnostics: Diagnostics
 ) {
   private def runloop(shouldStartIfNot: Boolean): UIO[RunloopState] =
-    runloopStateRef.updateSomeAndGetZIO { case RunloopState.NotStarted if shouldStartIfNot => makeRunloop }
+    runloopStateRef.updateSomeAndGetZIO {
+      case RunloopState.NotStarted if shouldStartIfNot =>
+        for {
+          promise <- Promise.make[Nothing, Runloop]
+          _       <- makeRunloop.map(promise.succeed).fork
```
guizmaii, line 60: I think tests are failing because of this fork. It might stop to early. I'm not sure how to wait for the end of the `makeRunloop.map(promise.succeed)` IO before to close the fiber without blocking the outer IO 🤔

@erikvanoosten Any idea?

guizmaii: I tried:
```scala
makeRunloop.map(promise.succeed).forkIn(scope)
```

The `scope` coming from the `RunloopAccess.make` (so being the Consumer scope), but it doesn't work either 🤔

guizmaii: Stupid mistake: I used `.map(promise.succeed)` instead of `.flatMap(promise.succeed)` 🤦‍♂️

guizmaii: Well, now that this stupid mistake is fixed, your solution is not passing the tests:
```scala
  + issue #856
    + Booting a Consumer to do something else than consuming should not fail with `RunloopTimeout` exception
    + Ordering of finalizers matters. If subscriptions are finalized after the runloop, it creates a deadlock
      + When not consuming, the Runloop is not started so only the Consumer is finalized
      + When consuming, the Runloop is started. The finalization orders matters to avoid a deadlock
      - Calling `Consumer::stopConsumption` just after starting a forked consumption session should stop the consumption
        ✗ 100000 was not less than 100000
        consumed_0 did not satisfy isLessThan(numberOfMessages.toLong)
        consumed_0 = 100000
        at /Users/jules/conduktor/workspace/zio-kafka/zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala:1163
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
```diff
@@ -37,40 +48,37 @@ private[internal] object RunloopState {
...
-    @inline def next: UIO[Unit] = stopConsumption(retry - 1, initialCall = false)
-
-    runloop(shouldStartIfNot = false).flatMap {
-      case RunloopState.Stopped          => ZIO.unit
-      case RunloopState.Started(runloop) => runloop.stopConsumption
-      case RunloopState.NotStarted =>
-        if (retry <= 0) ZIO.unit
-        else if (initialCall) next
-        else next.delay(10.millis)
+  def stopConsumption: UIO[Boolean] =
```
guizmaii, line 76: I need this `Boolean` in the returned type so that the caller can retry it if needed.

The test was failing because it's calling the `stopConsumption` before the Runloop had enough time to start

guizmaii:
Closing because the test solution (using a Promise) doesn't solve the issue

Issue #918, opened 2023-06-11
OOM Exception in version 2.3.1


jgordijn:
I am getting OOM exceptions while processing a kafka topic with 4mln backlog. Whenever I start the app with version `2.3.1`, I get the exception below. It works fine when I revert back to version `2.0.7`

The code:
```scala
def run(): UIO[Unit] =
    ZIO
      .scoped {
        ZIO.logInfo(s"Starting consumer on: ${topic.value}") *>
        Consumer.make(KafkaEventStreamImpl.consumerSettings(configuration)).flatMap { consumer =>
          consumer
            .plainStream(Subscription.topics(topic.value), Serde.string, Serde.string)
            .map { cr => cr.offset }
            .mapZIO(batch => batch.commitOrRetry(Schedule.exponential(1.second) && Schedule.recurs(3)))
            .runDrain
            .tapError(e => ZIO.logErrorCause(s"Error", Cause.fail(e)))
        }
      }
      .sandbox
      .tapError(e => ZIO.logErrorCause(s"Error", Cause.fail(e)))
      .ignore
```

```
java.lang.OutOfMemoryError: Java heap space
        at java.base/java.nio.HeapByteBuffer.<init>(Unknown Source)
        at java.base/java.nio.ByteBuffer.allocate(Unknown Source)
        at org.apache.kafka.common.memory.MemoryPool$1.tryAllocate(MemoryPool.java:30)
        at org.apache.kafka.common.network.NetworkReceive.readFrom(NetworkReceive.java:102)
        at org.apache.kafka.common.network.KafkaChannel.receive(KafkaChannel.java:452)
        at org.apache.kafka.common.network.KafkaChannel.read(KafkaChannel.java:402)
        at org.apache.kafka.common.network.Selector.attemptRead(Selector.java:674)
        at org.apache.kafka.common.network.Selector.pollSelectionKeys(Selector.java:576)
        at org.apache.kafka.common.network.Selector.poll(Selector.java:481)
        at org.apache.kafka.clients.NetworkClient.poll(NetworkClient.java:560)
        at org.apache.kafka.clients.consumer.internals.ConsumerNetworkClient.poll(ConsumerNetworkClient.java:280)
        at org.apache.kafka.clients.consumer.internals.ConsumerNetworkClient.pollNoWakeup(ConsumerNetworkClient.java:321)
[...]

PR #920, opened 2023-06-12
Prevent runaway feedback loop in optimistic resume

erikvanoosten:
At the cost of some performance, we are no longer including the predicted resumes in the poll history. This prevents runaway feedback loops which prevent out of memory problems.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -245,9 +245,10 @@ private[consumer] final class Runloop private (
     val resumeTps = new java.util.ArrayList[TopicPartition](streams.size)
     val pauseTps  = new java.util.ArrayList[TopicPartition](streams.size)
     streams.foreach { stream =>
-      val tp = stream.tp
-      val toResume =
-        requestedPartitions.contains(tp) || (consumerSettings.enableOptimisticResume && stream.optimisticResume)
+      val tp              = stream.tp
+      val requestedResume = requestedPartitions.contains(tp)
+      if (consumerSettings.enableOptimisticResume) stream.addPollHistory(requestedResume)
+      val toResume = requestedResume || (consumerSettings.enableOptimisticResume && stream.optimisticResume)
       if (toResume) resumeTps.add(tp) else pauseTps.add(tp)
       if (consumerSettings.enableOptimisticResume) {
         stream.addPollHistory(toResume)
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -247,12 +247,11 @@ private[consumer] final class Runloop private (
     streams.foreach { stream =>
       val tp              = stream.tp
       val requestedResume = requestedPartitions.contains(tp)
-      if (consumerSettings.enableOptimisticResume) stream.addPollHistory(requestedResume)
+      if (consumerSettings.enableOptimisticResume) {
+        stream.addPollHistory(requestedResume)
+      }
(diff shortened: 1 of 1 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -245,9 +245,10 @@ private[consumer] final class Runloop private (
...
-      val tp = stream.tp
-      val toResume =
-        requestedPartitions.contains(tp) || (consumerSettings.enableOptimisticResume && stream.optimisticResume)
+      val tp              = stream.tp
+      val requestedResume = requestedPartitions.contains(tp)
+      if (consumerSettings.enableOptimisticResume) stream.addPollHistory(requestedResume)
+      val toResume = requestedResume || (consumerSettings.enableOptimisticResume && stream.optimisticResume)
       if (toResume) resumeTps.add(tp) else pauseTps.add(tp)
       if (consumerSettings.enableOptimisticResume) {
         stream.addPollHistory(toResume)
```
guizmaii, line 254: Don't you need to remove these lines, then?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -245,9 +245,10 @@ private[consumer] final class Runloop private (
     val resumeTps = new java.util.ArrayList[TopicPartition](streams.size)
     val pauseTps  = new java.util.ArrayList[TopicPartition](streams.size)
     streams.foreach { stream =>
-      val tp = stream.tp
-      val toResume =
-        requestedPartitions.contains(tp) || (consumerSettings.enableOptimisticResume && stream.optimisticResume)
+      val tp              = stream.tp
+      val requestedResume = requestedPartitions.contains(tp)
+      if (consumerSettings.enableOptimisticResume) stream.addPollHistory(requestedResume)
```
guizmaii, line 250:
```suggestion
      if (consumerSettings.enableOptimisticResume) {
        stream.addPollHistory(requestedResume)
      }
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: It is still essential to record when more data was requested.

guizmaii: Aren't you adding to the history twice? I'm not sure to understand, sorry

erikvanoosten: Fixed!

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Okay!

guizmaii:
@erikvanoosten Do you think i'd be possible to add a test for this maybe?

erikvanoosten:
> @erikvanoosten Do you think i'd be possible to add a test for this maybe?

Not sure, we don't have any white box testing of runloop, it will be difficult. But I will take a look.

erikvanoosten:
Unfortunately, this doesn't work. Performance degrades considerably: from 1050 to 1550 ms/op. Marked as draft for now.

svroonland:
@erikvanoosten @guizmaii  Given the current issues with this approach and the fact that it's conceptually harder and more error-prone, I'd vote for removing it and reinstating #803.

guizmaii:
@svroonland, I agree. I already suggested to remove it here https://github.com/zio/zio-kafka/issues/918#issuecomment-1587089854

svroonland:
Right, sorry I missed that.

erikvanoosten:
No longer needed due to #803 and #944.

PR #921, opened 2023-06-13
Avoid using `ZIO.succeed` when wrapping Kafka code

guizmaii:


zio-kafka/src/main/scala/zio/kafka/consumer/internal/ConsumerAccess.scala
```diff
@@ -52,7 +52,7 @@ private[consumer] object ConsumerAccess {
                       )
                     }
                   } { consumer =>
-                    ZIO.blocking(access.withPermit(ZIO.succeed(consumer.close(settings.closeTimeout))))
+                    ZIO.blocking(access.withPermit(ZIO.attempt(consumer.close(settings.closeTimeout)))).orDie
```
adamgfraser, line 55: This is the same as `succeed`.

guizmaii: Good to know. Thanks! :)

erikvanoosten (approved):

PR #929, opened 2023-06-17
[Intenal] Reduce `OffsetImpl` memory footprint

guizmaii:


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -218,7 +218,7 @@ private[consumer] final class Runloop private (
     }
   }
 
-  private def getConsumerGroupMetadataIfAny: UIO[Option[ConsumerGroupMetadata]] =
```
guizmaii, line 221: So that we only evaluate the `if` once

erikvanoosten: The wonders of IO 👍 .

zio-kafka/src/main/scala/zio/kafka/consumer/Offset.scala
```diff
@@ -36,12 +36,14 @@ object Offset {
 }
 
 private final case class OffsetImpl(
-  topic: String,
-  partition: Int,
-  offset: Long,
-  commitHandle: Map[TopicPartition, Long] => Task[Unit],
+  private val record: ConsumerRecord[_, _],
```
erikvanoosten, line 39: On one side we need less CPU because we can skip copying 3 values. On the other side `record` carries a lot of data such as the key and value of the message. Especially when the values are large, this means more pressure on the garbage collector.

In one of my applications the offsets of all ~2000 streams are merged into a single stream for committing. So this application keeps the offsets in memory for quite a bit longer than that it needs the values of the messages. I am afraid that this application will need much more heap due to this change.

guizmaii: Yeah. I realized this morning that It's actually a bad idea. It can prevent the GC to collect the `ConsumerRecord` and be more heavyweight in terms of memory consumption. My bad

erikvanoosten:
@guizmaii this PR is still good for changing getConsumerGroupMetadataIfAny to a val.

guizmaii:
@erikvanoosten Ok. I'll open a new PR then :)

guizmaii:
See https://github.com/zio/zio-kafka/pull/931

PR #931, opened 2023-06-18
[Internal] Only evaluate `if (hasGroupId)` once in `getConsumerGroupMetadataIfAny`

guizmaii:


svroonland (approved):
I wonder if the JVM does these kinds of optimizations by itself

guizmaii:
> I wonder if the JVM does these kinds of optimizations by itself

It probably does. But if we do it for it, it can focus on something else

yarosman:
> I wonder if the JVM does these kinds of optimizations by itself

the optimizer can't change the method signature from `def` to `val`, therefore it will create (allocate) a new object every time `getConsumerGroupMetadataIfAny` is called.

PR #932, opened 2023-06-18
Revert "Optimistically resume partitions early"

svroonland:
This reverts commit dc46d3e9b98d589768f31d4051ad17f326f5b0b8 and #809


zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -20,22 +21,17 @@ import zio.kafka.security.KafkaCredentialStore
...
- *   pauses partitions for which the associated stream is processing previously fetched data. When the stream needs more
- *   data, the partition is resumed. With this feature enabled, partitions are also resumed when it is likely that the
- *   stream needs more data in the next poll.
  */
 final case class ConsumerSettings(
   bootstrapServers: List[String],
   properties: Map[String, AnyRef] = Map.empty,
   closeTimeout: Duration = 30.seconds,
   pollTimeout: Duration = 50.millis,
+  perPartitionChunkPrefetch: Int = 2,
```
guizmaii, line 30: Do we really want to put this back?

svroonland: My plan is to remove it in 803 again, this is just a revert of 809 (as clean as possible)

guizmaii: I'll leave the approval of this PR to Erik at it's his work if that's ok for you

erikvanoosten: If we go back to this, we should also bring back this setting.
However, I still think we can do better. E.g. by making this maximum prefetch (like in fs2-kafka):

```suggestion
  perPartitionChunkMaxPrefetch: Int = 2,
```

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
```diff
@@ -121,6 +121,7 @@ object KafkaTestUtils {
           ConsumerConfig.MAX_POLL_RECORDS_CONFIG         -> s"${`max.poll.records`}",
           ConsumerConfig.ALLOW_AUTO_CREATE_TOPICS_CONFIG -> allowAutoCreateTopics.toString
         )
+        .withPerPartitionChunkPrefetch(16)
```
erikvanoosten, line 124: For more realistic testing this should stays on the default.
BTW, shouldn't this be configurable, I mean as a parameter to the method?

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
erikvanoosten: We'll leave it for another PR.

erikvanoosten (approved):

PR #933, opened 2023-06-18
Pre-fetch per consumer, not per stream

erikvanoosten:
Resume some partitions when there is no other demand, limited by the maxPrefetchPolls setting.

We select partitions based on the estimate number of polls to the next resume. This algorithm is much faster and more accurate than the previous bit-pattern based algorithm.

The advantages of pre-fetching per consumer are:
 - No more OOM; only a fixed number of polls are pre-fetched so additional memory pressure is very limited.
 - Read-starvation problem is back to pre-2.1 zio-kafka. This is because the pre-fetching only kicks in when no other streams have demand.

Todo before going out of draft:
 - [ ] Check performance.
 - [ ] Move magic values into a constant.
 - [ ] See if we need more unit tests.


erikvanoosten:
Some notes about the implementation at this point:
 - the estimated polls to resume gives a conservative estimate because pre-fetches are not added to the poll history,
 - pre-fetching only happens when no other streams need data, this can limit throughput when there is no data for those streams that have demand.

erikvanoosten:
No longer needed due to #944.

PR #940, opened 2023-06-20
Use less heap by calling offer directly

erikvanoosten:
Instead of first building a sequence of committable records and then offering those to the streams, we now offer the records directly after they are built. This saves an intermediate data structure while keeping the same number of ZIOs.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -188,9 +188,8 @@ private[consumer] final class Runloop private (
         _ <- ZIO.foreachDiscard(streams) { streamControl =>
                val tp      = streamControl.tp
                val records = polledRecords.records(tp)
-               if (records.isEmpty) {
-                 ZIO.unit
-               } else {
+               if (records.isEmpty) ZIO.unit
+               else {
                  val builder  = ChunkBuilder.make[Record](records.size())
                  val iterator = records.iterator()
                  while (iterator.hasNext) {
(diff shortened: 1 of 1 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -184,34 +185,26 @@ private[consumer] final class Runloop private (
...
-              }
-              acc += (streamControl -> builder.result())
-            }
-          }
-          acc.result()
-        }
-        _ <- ZIO
-               .foreachDiscard(committableRecords) { case (streamControl, records) =>
-                 streamControl.offerRecords(records)
+        _ <- ZIO.foreachDiscard(streams) { streamControl =>
```
erikvanoosten, line 188: We should also test this with:
```suggestion
        _ <- ZIO.foreachParDiscard(streams) { streamControl =>
```

guizmaii: +1

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -184,34 +185,26 @@ private[consumer] final class Runloop private (
...
+               val tp      = streamControl.tp
+               val records = polledRecords.records(tp)
+               if (records.isEmpty) {
+                 ZIO.unit
+               } else {
```
guizmaii, line 193:
```suggestion
               if (records.isEmpty) ZIO.unit
               else {
```

guizmaii (approved):
LGTM

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -15,6 +15,7 @@ import java.util
 import scala.jdk.CollectionConverters._
 
 //noinspection SimplifyWhenInspection
+//noinspection SimplifyUnlessInspection
```
guizmaii, line 18: +1

PR #942, opened 2023-06-24
Review #803: Optimize `partitionsToFetch` computation

guizmaii:
Review of #803


svroonland (approved):

erikvanoosten (approved):
Although I have nothing against this change, I do not understand why this is an optimization. Can you explain that please?

guizmaii:
@erikvanoosten

Here's how I see it:
- only one collection built: previously, we were building `queueSizes` and `partitionsToFetch`, we now only building and having in memory `partitionsToFetch`

PR #950, opened 2023-06-27
Fix: Make the `SslHelper.validateEndpoint` as resilient as Kafka. If some Kafka bootstraps servers are down, that shouldn't affect the result of the validation 

### LINKED ISSUE #949: Cannot start any of zio-kafka clients with security.protocol = SASL_PLAINTEXT when one of Kafka nodes is down
Hi, I've got issue in production with `SslHelper.validateEndpoint` - one of nodes was removed from Kafka cluster temporarily and I was unable to restart app - because `validateEndpoint` was failing endlessly on this node. 

```
org.apache.kafka.common.KafkaException: Failed to create new KafkaAdminClient
	at zio.kafka.utils.SslHelper$.$anonfun$validateEndpoint$21(SslHelper.scala:72)
	at zio.Cause.$anonfun$map$1(Cause.scala:418)
	at zio.Cause.$anonfun$flatMap$2(Cause.scala:173)
	at zio.Cause$$anon$2.failCase(Cause.scala:288)
	at zio.Cause$$anon$2.failCase(Cause.scala:276)
	at zio.Cause.loop$2(Cause.scala:221)
	at zio.Cause.foldContext(Cause.scala:248)
	at zio.Cause.foldLog(Cause.scala:276)
	at zio.Cause.flatMap(Cause.scala:179)
	at zio.Cause.map(Cause.scala:418)
	at zio.ZIO.$anonfun$mapError$1(ZIO.scala:981)
	at zio.ZIO.$anonfun$mapErrorCause$1(ZIO.scala:993)
	at zio.internal.FiberRuntime.runLoop(FiberRuntime.scala:1121)
	at zio.internal.FiberRuntime.evaluateEffect(FiberRuntime.scala:381)
	at zio.internal.FiberRuntime.evaluateMessageWhileSuspended(FiberRuntime.scala:504)
	at zio.internal.FiberRuntime.drainQueueOnCurrentThread(FiberRuntime.scala:220)
	at zio.internal.FiberRuntime.run(FiberRuntime.scala:139)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: java.net.ConnectException: Connection refused
	at java.base/sun.nio.ch.Net.connect0(Native Method)
	at java.base/sun.nio.ch.Net.connect(Net.java:579)
	at java.base/sun.nio.ch.Net.connect(Net.java:586)
	at java.base/sun.nio.ch.SocketChannelImpl.connect(SocketChannelImpl.java:853)
	at java.base/java.nio.channels.SocketChannel.open(SocketChannel.java:285)
	at zio.kafka.utils.SslHelper$.$anonfun$validateEndpoint$11(SslHelper.scala:47)
	at zio.ZIOCompanionVersionSpecific.$anonfun$attempt$1(ZIOCompanionVersionSpecific.scala:100)
	at zio.internal.FiberRuntime.runLoop(FiberRuntime.scala:904)
	at zio.internal.FiberRuntime.runLoop(FiberRuntime.scala:890)
```

guizmaii:
Fixes #949

I want to publicly thank @vladimirkl for his invaluable help on this work! ❤️

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/utils/SslHelper.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/utils/SslHelper.scala
@@ -162,7 +162,7 @@ object SslHelper {
     unsafeOpenSocket: InetSocketAddress => SocketChannel,
     socketTimeout: Duration
   )(address: InetSocketAddress): Task[Unit] = {
-    @inline def error: IO[IllegalArgumentException, Nothing] =
+    @inline def unexpectedSslPacketError: IO[IllegalArgumentException, Nothing] =
       ZIO.fail(
         new IllegalArgumentException(
           "Received an unexpected SSL packet from the server. Please ensure the client is properly configured with SSL enabled"
@@ -190,7 +190,7 @@ object SslHelper {
       } finally channel.close()
     }
       .timeoutFail(timeoutException)(socketTimeout)
-      .flatMap(isTLS => if (isTLS) error else ZIO.unit)
+      .flatMap(isTLS => if (isTLS) unexpectedSslPacketError else ZIO.unit)
   }
 
   /**


zio-kafka/src/main/scala/zio/kafka/utils/SslHelper.scala
```diff
@@ -21,56 +23,94 @@ import scala.jdk.CollectionConverters._
...
+          s"Received an unexpected SSL packet from the server. Please ensure the client is properly configured with SSL enabled"
+        )
+      )
+
+    ZIO.scoped {
+      for {
+        channel <- ZIO.acquireRelease(
+                     ZIO.attempt(openSocket(address)).mapError(ConnectExceptionWrapper)
+                   )(channel => ZIO.attempt(channel.close()).orDie)
+        tls <- ZIO.attempt {
```
vladimirkl, line 104: Should we use `ZIO.attemptBlocking` here?

guizmaii: All the code is wrapped in a `ZIO.blocking` ;)

vladimirkl: Oh, I see, thank you. Other question - do we need some kind of timeout here? Depending on network conditions, reply from disabled node could be delayed.

guizmaii: There's already a timeout. If you try to login to `localhost:1234`, it'll timeout

zio-kafka/src/main/scala/zio/kafka/utils/SslHelper.scala
```diff
@@ -21,56 +23,94 @@ import scala.jdk.CollectionConverters._
...
+  private final case class ConnectExceptionWrapper(cause: Throwable) extends NoStackTrace
+
+  // ⚠️ Must not do anything else than calling `_validateEndpoint`.
+  def validateEndpoint(bootstrapServers: List[String], props: Map[String, AnyRef]): IO[KafkaException, Unit] =
+    _validateEndpoint(SocketChannel.open)(bootstrapServers, props)
+
+  /**
+   * We use this private function so that we can easily manipulate the `openSocket` function in unit-tests.
+   */
+  private[utils] def _validateEndpoint(
```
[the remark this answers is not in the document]
guizmaii: Renamed `doValidateEndpoint` :)

zio-kafka/src/main/scala/zio/kafka/utils/SslHelper.scala
```diff
@@ -21,56 +23,94 @@ import scala.jdk.CollectionConverters._
...
-                           .toList
+  private final case class ConnectExceptionWrapper(cause: Throwable) extends NoStackTrace
+
+  // ⚠️ Must not do anything else than calling `_validateEndpoint`.
+  def validateEndpoint(bootstrapServers: List[String], props: Map[String, AnyRef]): IO[KafkaException, Unit] =
+    _validateEndpoint(SocketChannel.open)(bootstrapServers, props)
+
+  /**
+   * We use this private function so that we can easily manipulate the `openSocket` function in unit-tests.
+   */
```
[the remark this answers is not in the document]
guizmaii: This is a private function, so this is a private documentation

zio-kafka/src/main/scala/zio/kafka/utils/SslHelper.scala
```diff
@@ -21,56 +23,94 @@ import scala.jdk.CollectionConverters._
...
-        ZIO.blocking {
-          for {
-            address <- ZIO.attempt {
-                         ClientUtils
-                           .parseAndValidateAddresses(bootstrapServers.asJava, ClientDnsLookup.USE_ALL_DNS_IPS)
-                           .asScala
-                           .toList
+  private final case class ConnectExceptionWrapper(cause: Throwable) extends NoStackTrace
+
+  // ⚠️ Must not do anything else than calling `_validateEndpoint`.
```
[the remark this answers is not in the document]
guizmaii: Because the code of this function is contained in `_validateEndpoint` and should only be contained into `_validateEndpoint `

zio-kafka/src/main/scala/zio/kafka/utils/SslHelper.scala
```diff
@@ -21,56 +23,94 @@ import scala.jdk.CollectionConverters._
...
+                         // We don't propagate the internal wrapper
+                         case error: ConnectExceptionWrapper => ZIO.fail(error.cause)
+                         case e                              => ZIO.fail(e)
+                       }
+                     case _ => // Some calls failed
+                       errors.collect { case e if !e.isInstanceOf[ConnectExceptionWrapper] => e } match {
+                         // No "real errors", we ignore the internally wrapped errors
+                         case Nil => ZIO.unit
+                         // Some "real errors", we propagate the first one
+                         case realError :: _ => ZIO.fail(realError)
```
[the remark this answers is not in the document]
guizmaii: FYI, I gave a try to your algorithm here: https://github.com/zio/zio-kafka/pull/954
Conclusion: I think I prefer yours :)
Thanks!

zio-kafka/src/main/scala/zio/kafka/utils/SslHelper.scala
guizmaii: Implemented here: https://github.com/zio/zio-kafka/pull/950/commits/674114edabec347f4ca30cdbdc94887cb9dbd55e

vpavkin:
Is it possible to make this check opt-in? I also have trouble starting my producer with seemingly correct configuration

guizmaii:
> Is it possible to make this check opt-in? 

I'd prefer not to. This PR will make the algorithm as resilient as Kafka clients: if some nodes are down/inaccessible, it'll ignore them

vpavkin:
I have different kind of problem when `validateEndpoint` produces a false positive (thinks it's tls while in reality it's not).
up to you of course, but I don't see harm in a configuration flag, hidden under a default `true` value

guizmaii:
@vpavkin 
> thinks it's tls while in reality it's not

What do you mean by that? This check is only performed if your client is configured for an SSL connection to the server.

vpavkin:
we're connecting using SASL_PLAINTEXT, so the check triggers but fails

guizmaii:
Should it be triggered with `SASL_PLAINTEXT`? 🤔

vpavkin:
hey @guizmaii , sorry for confusion, we discovered an issue in our config, so you can disregard all of my comments above.

Thank you for working on this issue ❤️

guizmaii:
@vpavkin Thanks 🙂

Can you share maybe more details about the issue you had in your config?

zio-kafka/src/main/scala/zio/kafka/utils/SslHelper.scala
```diff
@@ -21,61 +23,179 @@ import scala.jdk.CollectionConverters._
...
+  )(bootstrapServers: List[String], props: Map[String, AnyRef]): IO[KafkaException, Unit] = {
+    @inline def `request.timeout.ms`: Duration = {
+      val defaultValue = 30.seconds
+
+      props.get("request.timeout.ms") match {
+        case None => defaultValue
+        case Some(raw) =>
+          try {
+            val v = raw.toString.toLong
+            if (v <= 0) defaultValue else v.millis
```
erikvanoosten, line 52: Often `0` is used to disable timeouts. Is that the case here as well (if you ask me: no)? But if its 'yes', the code should be changed to:

```suggestion
            if (v < 0) defaultValue else v.millis
```

zio-kafka/src/main/scala/zio/kafka/utils/SslHelper.scala
```diff
@@ -21,61 +23,179 @@ import scala.jdk.CollectionConverters._
...
+                           }
+              errors <- ZIO.collectAllFailuresPar(
+                          addresses.map(validateSslConfigOf(unsafeOpenSocket, socketTimeout = `request.timeout.ms`))
+                        )
+              atLeastOneBootstrapServerIsUp = errors.size < addresses.size
+              _ <- errors.partition(_.isInstanceOf[ConnectionError]) match {
+                     // If we have at least one "real error" (not our internal "ConnectionError"), we fail with the first one
+                     case (_, head :: _) => ZIO.fail(head)
+                     // If we have no errors or if we have some of our internal "ConnectionError" but at least one bootstrap server is up, we succeed
+                     case (Nil, Nil)                         => ZIO.unit
```
erikvanoosten, line 86: Maybe make `case (Nil, Nil)` the first case.

BTW, please be careful with using `Nil`. `Nil` is an empty `List` and will therefore not match other empty collections. It all works here because of line 76.

zio-kafka/src/main/scala/zio/kafka/utils/SslHelper.scala
```diff
@@ -21,61 +23,179 @@ import scala.jdk.CollectionConverters._
...
+   *   - Discord discussion/thread with Vladimir Klyushnikov (https://github.com/vladimirkl) starting here:
+   *     https://discord.com/channels/629491597070827530/1122924033827164282/1124719212246610023
+   *   - https://stackoverflow.com/questions/2866557/timeout-for-socketchannel-doesnt-work#comment11463818_2866557
+   *   - https://stackoverflow.com/a/18375293/2431728
+   */
+  private def validateSslConfigOf(
+    unsafeOpenSocket: InetSocketAddress => SocketChannel,
+    socketTimeout: Duration
+  )(address: InetSocketAddress): Task[Unit] = {
+    @inline def error: IO[IllegalArgumentException, Nothing] =
```
erikvanoosten, line 164:
```suggestion
    @inline def unexpectedSslPacketError: IO[IllegalArgumentException, Nothing] =
```

zio-kafka/src/main/scala/zio/kafka/utils/SslHelper.scala
```diff
@@ -21,61 +23,179 @@ import scala.jdk.CollectionConverters._
...
+      ZIO.fail(
+        new IllegalArgumentException(
+          s"Received an unexpected SSL packet from the server. Please ensure the client is properly configured with SSL enabled"
```
erikvanoosten, line 167:
```suggestion
          "Received an unexpected SSL packet from the server. Please ensure the client is properly configured with SSL enabled"
```

zio-kafka/src/main/scala/zio/kafka/utils/SslHelper.scala
```diff
@@ -21,61 +23,179 @@ import scala.jdk.CollectionConverters._
...
+   *      most tricky/weird part, to close the socket, we interrupt the thread running it.
+   *
+   * Why does interrupting the thread running the networking exchange closes the socket? Because the `SocketChannel`
+   * class implements the `InterruptibleChannel` interface, and that's a property of this interface.
+   *
+   * From the documentation of `InterruptibleChannel`, we can read:
+   * {{{
+   * > A channel that implements this interface is also interruptible:
+   * > If a thread is blocked in an I/O operation on an interruptible channel then another thread may invoke the blocked thread's interrupt method.
+   * > This will cause the channel to be closed, the blocked thread to receive a ClosedByInterruptException, and the blocked thread's interrupt status to be set.
```
erikvanoosten, line 140: Nifty!

erikvanoosten (approved):
Nice! Some small remarks.

zio-kafka/src/main/scala/zio/kafka/utils/SslHelper.scala
guizmaii: We can't use `0` here as it'd mean "infinite"

zio-kafka/src/main/scala/zio/kafka/utils/SslHelper.scala
guizmaii: > Maybe make case (Nil, Nil) the first case.

Things are easier to explain with the `case (Nil, Nil)` and the following one are next to each other.
The comment just above explains both lines.

> BTW, please be careful with using Nil. Nil is an empty List and will therefore not match other empty collections. It all works here because of line 76.

If it was not a `List` the code would not compile

erikvanoosten: > If it was not a `List` the code would not compile

Unfortunately, that is not the case. I found this out the hard way. E.g. this compiles just fine even though `s` is not a `List`:
```scala
    val s: Seq[Int] = Seq.empty[Int]
    s match {
      case Nil => println("empty list")
      case Seq() => println("empty seq")
      case _ => println("something else")
    }
```

There must have been some improvements though. I am unable to come with an expression of type Seq[Int] that doesn't match `Nil` but does match `Seq()`. All these expressions match `Nil`: `Seq(1).drop(1)`, `Vector(1).filter(_ % 2 == 0)`, `"".map(_ => 1)`, `ArraySeq[Int]()`.

So I guess Scala has become more safe in this regard!

guizmaii: Well, it's generally advised to never use `Seq`. This interface has many issues.

PR #953, opened 2023-06-30
Introduce FetchStrategy to allow custom fetch strategies

erikvanoosten:
Extracting the fetch strategy from runloop allow anyone to experiment more freely with different way to determine which partitions are allowed to poll.


svroonland (commented):
A good first step. Let's think if this is a sufficiently powerful interface for future FetchStrategies. Would it be enough to implement the optimistic resume from the previous zio-kafka version for example?

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/FetchStrategy.scala
```diff
@@ -0,0 +1,85 @@
...
+ *   Maximum number of records to be buffered per partition. This buffer improves throughput and supports varying
+ *   downstream message processing time, while maintaining some backpressure. Large values effectively disable
+ *   backpressure at the cost of high memory usage, low values will effectively disable prefetching in favour of low
+ *   memory consumption. The number of records that is fetched on every poll is controlled by the `max.poll.records`
+ *   setting, the number of records fetched for every partition is somewhere between 0 and `max.poll.records`. A value
+ *   that is a power of 2 offers somewhat better queueing performance.
+ *
+ * The default value for this parameter is 2 * the default `max.poll.records` of 500, rounded to the nearest power of 2.
+ */
+case class DefaultFetchStrategy(maxPartitionQueueSize: Int = 1024) extends FetchStrategy {
```
guizmaii, line 33:
```suggestion
final case class DefaultFetchStrategy(maxPartitionQueueSize: Int = 1024) extends FetchStrategy {
```

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/FetchStrategy.scala
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * The default value for this parameter is 2 * the default `max.poll.records` of 500, rounded to the nearest power of 2.
+ * @param maxTotalQueueSize
+ *   Maximum number of records to be buffered over all partitions together. This can be used to limit memory usage when
+ *   consuming a large number of partitions.
+ *
+ * The default value is 20 * the default for `maxTotalQueueSize`, allowing approximately 20 partitions to do
+ * pre-fetching in each poll.
+ */
+case class ManyPartitionsFetchStrategy(
```
guizmaii, line 65:
```suggestion
final case class ManyPartitionsFetchStrategy(
```

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/FetchStrategy.scala
```diff
@@ -0,0 +1,85 @@
+package zio.kafka.consumer.fetch
+
+import org.apache.kafka.common.TopicPartition
+import zio.kafka.consumer.internal.PartitionStreamControl
+import zio.{ Chunk, ZIO }
+
+trait FetchStrategy {
```
guizmaii, line 7:
```suggestion
sealed trait FetchStrategy {
```

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/FetchStrategy.scala
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * Selects which partitions can fetch data in the next poll.
+   *
+   * @param streams
+   *   all stream for this consumer
+   * @return
+   *   the partitions that may fetch in the next poll
+   */
+  def selectPartitionsToFetch(streams: Chunk[PartitionStreamControl]): ZIO[Any, Nothing, Set[TopicPartition]]
```
guizmaii, line 17: Can we try to avoid OO code design maybe? 

If possible, I'd prefer that we stay with a simple ADT and a function taking a strategy in input:
```scala
sealed trait FetchStrategy
object FetchStrategy {
  final case class A(...) extends FetchStrategy
  final case class B(...) extends FetchStrategy
}

and later:
def selectPartitionsToFetch(strategy: FetchStrategy) = ..
```

Do you think that'd be possible?

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/FetchStrategy.scala
guizmaii, line 33: Can we find a better, more descriptive of what it does, name maybe?
Then the "default" would be implemented like this:
```scala
sealed trait FetchStrategy
object FetchStrategy {
  val default = PrefetchStrategy(maxPartitionQueueSize = 1024)

  // If we imagine that we rename this strategy "PrefetchStrategy"
  final case class PrefetchStrategy(maxPartitionQueueSize: Int) extends FetchStrategy
  final case class ManyPartitionsFetchStrategy(...) extends FetchStrategy
```

guizmaii:
@erikvanoosten See my different comments

I think that this "OO" approach isn't a good one. 
I think that the "future-proofness" of the interface that @svroonland is worried about comes from this OO design.
If we take my suggested "ADT+function taking a strategy as input", we don't have this "future-proofness" issue

svroonland:
Yeah perhaps indeed good to make it sealed for now until we know better how this generalizes, and if we want to open it up to users to make their own implementations.

erikvanoosten:
> A good first step. Let's think if this is a sufficiently powerful interface for future FetchStrategies. Would it be enough to implement the optimistic resume from the previous zio-kafka version for example?

Yes, I think it is 😄

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/FetchStrategy.scala
erikvanoosten: One of the design goals (IMHO) is that anyone can implement a fetch strategy. A sealed trait would prevent that.

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/FetchStrategy.scala
erikvanoosten: Yes! I was lazy 😄 
A good name is `QueueSizeBasedFetchStrategy`.

I would prefer we avoid the word 'pre-fetch'; because of batching the difference between 'fetch' and 'pre-fetch' is a bit blurry. Better avoid the confusion.

erikvanoosten:
> Yeah perhaps indeed good to make it sealed for now until we know better how this generalizes, and if we want to open it up to users to make their own implementations.

Another approach is that we mark the interface as experimental; so no guarantees for its stability. This has the advantage that anyone can experiment. When someone finds something that works really well it could be shared back 😄

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/FetchStrategy.scala
erikvanoosten: See https://github.com/zio/zio-kafka/pull/953#issuecomment-1616436849

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/FetchStrategy.scala
erikvanoosten: Thanks.

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/FetchStrategy.scala
erikvanoosten: Done also.

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/FetchStrategy.scala
erikvanoosten: Changed it to `QueueSizeBasedFetchStrategy`.

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/FetchStrategy.scala
erikvanoosten: Solved this by marking the trait as experimental.

Even though this exposes `PartitionStreamControl`, almost all its methods and fields are hidden. Only `tp` and `queueSize` are available.

svroonland (approved):

PR #957, opened 2023-07-01
Add a timeout mechanism for the `Socket.open` in the `SslHelper.validateEndpoint` helper

guizmaii:


zio-kafka/src/main/scala/zio/kafka/utils/SslHelper.scala
```diff
@@ -96,9 +115,30 @@ object SslHelper {
...
+            ZIO.uninterruptible {
+              // It's imperative that this algorithm is not interruptible because of the timeout mechanism.
+              // Otherwise, we might leak a socket channel.
+              // If the timeout interruption happens between the `openSocket` and the `ref.set`, we would leak the socket channel.
+              ZIO.attempt(openSocket(address)).tap(ref.set)
+            }.timeoutFail(new java.util.concurrent.TimeoutException(s"Failed to contact $address"))(socketTimeout)
+              .mapError(ConnectionError.apply)
+          )(
+            ref.update { socket =>
+              if (socket != null) {
```
vladimirkl, line 130: Better to use `socket ne null` here

Issue #902, opened 2023-06-05
Replace runloop timout with per stream halting detection


erikvanoosten:
Kafka's process to detect halted consumers is a kind of crude. Because it is based on polls, it works per consumer. We can make it more fine grained. E.g. by doing the detection per subscription. When one of the streams associated with that subscription halts, we can unsubscribe just that subscription.

For the Kafka client 'halting' means poll is not called for more than `max.poll.interval.ms`. Translated to zio-kafka 'halting' means that a stream does not require more data for more then `max.poll.interval.ms`.

A halted stream should be interrupted so that it can fail properly.

As discussed on [Discord](https://discord.com/channels/629491597070827530/629497941719121960/1114868628156858400).

Replaces #686.

guizmaii:
🔥

Issue #686, opened 2023-03-05
Close consumer when it is too slow for a few partitions


erikvanoosten:
When the consumer is progressing too slowly (or not at all), we want the Kafka broker to learn about it so that the broker can revoke the partitions from this consumer. The needed mechanism already exists for the case where all streams are too slow to request data for their partition. In that case, poll will not be invoked since there are no requests for data. After `max.poll.interval.ms` without polls, the partitions will be revoked by the broker.

However, if at least one stream is still making progress, poll will continue to be invoked and therefore no revocation happens.

The idea to fix this is by keeping track of the delay between data requests from each stream. When a stream needs more than `max.poll.interval.ms`, we should close the consumer. Alternatively, we can close only the slow stream.

This is a follow-up from #664.

PR #960, opened 2023-07-02
Add processing timeout at the stream level

erikvanoosten:
Kafka's process to detect halted consumers is crude. Because it is based on polls, it works per consumer. Here we make it more fine grained by doing slow consumer detection per stream.

For the Kafka client 'halting' means poll is not called for longer than `max.poll.interval.ms`. Translated to zio-kafka 'halting' means that a stream does not consume any records for `max.poll.interval.ms` or longer, while records are available.

The halted stream is interrupted so that it can fail properly. In addition, we shut down the entire consumer. This will help the broker to detect slow consumers. Ideally we would only stop the subscription that the stream belongs to. Unfortunately, we do not maintain a topic belongs-to subscription relation.

As discussed on [Discord](https://discord.com/channels/629491597070827530/629497941719121960/1114868628156858400).

ExtraZStreamOps is based on a code example from @adamgfraser. Earlier versions were based on code from @paulpdaniels.

Replaces https://github.com/zio/zio-kafka/issues/686.
Implements #902.

Co-authored-by: Adam Fraser <adam.fraser@gmail.com>
Co-authored-by: Paul Daniels <paulpdaniels@gmail.com>

### DIFF
--- a/zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
+++ b/zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
@@ -103,7 +103,6 @@ object KafkaTestUtils {
     offsetRetrieval: OffsetRetrieval = OffsetRetrieval.Auto(),
     restartStreamOnRebalancing: Boolean = false,
     `max.poll.records`: Int = 100, // settings this higher can cause concurrency bugs to go unnoticed
-    runloopTimeout: Duration = ConsumerSettings.defaultRunloopTimeout,
     commitTimeout: Duration = ConsumerSettings.defaultCommitTimeout,
     properties: Map[String, String] = Map.empty
   ): URIO[Kafka, ConsumerSettings] =
@@ -113,7 +112,6 @@ object KafkaTestUtils {
         .withCloseTimeout(5.seconds)
         .withPollTimeout(100.millis)
         .withCommitTimeout(commitTimeout)
-        .withRunloopTimeout(runloopTimeout)
         .withProperties(
           ConsumerConfig.AUTO_OFFSET_RESET_CONFIG        -> "earliest",
           ConsumerConfig.METADATA_MAX_AGE_CONFIG         -> "100",
--- a/zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
+++ b/zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
@@ -116,6 +116,7 @@ object KafkaTestUtils {
     allowAutoCreateTopics: Boolean = true,
     offsetRetrieval: OffsetRetrieval = OffsetRetrieval.Auto(),
     restartStreamOnRebalancing: Boolean = false,
+    maxPollInterval: Duration = 10.seconds,
     `max.poll.records`: Int = 100, // settings this higher can cause concurrency bugs to go unnoticed
     commitTimeout: Duration = ConsumerSettings.defaultCommitTimeout,
     properties: Map[String, String] = Map.empty
@@ -125,14 +126,14 @@ object KafkaTestUtils {
         .withClientId(clientId)
         .withCloseTimeout(5.seconds)
         .withPollTimeout(100.millis)
+        .withMaxPollInterval(maxPollInterval)
         .withCommitTimeout(commitTimeout)
         .withProperties(
           ConsumerConfig.AUTO_OFFSET_RESET_CONFIG        -> "earliest",
           ConsumerConfig.METADATA_MAX_AGE_CONFIG         -> "100",
           ConsumerConfig.SESSION_TIMEOUT_MS_CONFIG       -> "3000",
-          ConsumerConfig.MAX_POLL_INTERVAL_MS_CONFIG     -> "10000",
           ConsumerConfig.HEARTBEAT_INTERVAL_MS_CONFIG    -> "1000",
-          ConsumerConfig.MAX_POLL_RECORDS_CONFIG         -> s"${`max.poll.records`}",
+          ConsumerConfig.MAX_POLL_RECORDS_CONFIG         -> `max.poll.records`.toString,
           ConsumerConfig.ALLOW_AUTO_CREATE_TOPICS_CONFIG -> allowAutoCreateTopics.toString
         )
         .withOffsetRetrieval(offsetRetrieval)
--- a/zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
+++ b/zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
@@ -96,7 +96,7 @@ object KafkaTestUtils {
   /**
    * A stream that produces messages to a Topic on a schedule for as long as it is running.
    */
-  def scheduledProducer[R](
+  def scheduledProduce[R](
     topic: String,
     schedule: Schedule[R, Any, Long]
   ): ZStream[R with Producer, Throwable, RecordMetadata] =
--- a/zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
+++ b/zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
@@ -9,6 +9,7 @@ import zio.kafka.consumer._
 import zio.kafka.consumer.diagnostics.Diagnostics
 import zio.kafka.producer._
 import zio.kafka.serde.{ Deserializer, Serde }
+import zio.stream.ZStream
 
 import java.io.File
 import java.nio.file.{ Files, StandardCopyOption }
@@ -91,6 +92,19 @@ object KafkaTestUtils {
         Serde.string,
         Serde.string
       )
+
+  /**
+   * A stream that produces messages to a Topic on a schedule for as long as it is running.
+   */
+  def scheduledProducer[R](
+    topic: String,
+    schedule: Schedule[R, Any, Long]
+  ): ZStream[R with Producer, Throwable, RecordMetadata] =
+    ZStream
+      .fromSchedule(schedule)
+      .mapZIO { i =>
+        produceOne(topic, s"key$i", s"msg$i")
+      }
 
   /**
    * Utility function to make a Consumer settings set.
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
@@ -119,5 +119,5 @@ final case class ConsumerSettings(
 }
 
 object ConsumerSettings {
-  val defaultCommitTimeout: Duration  = 15.seconds
+  val defaultCommitTimeout: Duration = 15.seconds
 }
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
@@ -25,7 +25,6 @@ final case class ConsumerSettings(
   properties: Map[String, AnyRef] = Map.empty,
   closeTimeout: Duration = 30.seconds,
   pollTimeout: Duration = 50.millis,
-  maxPollInterval: Duration = 5.minutes,
   commitTimeout: Duration = ConsumerSettings.defaultCommitTimeout,
   offsetRetrieval: OffsetRetrieval = OffsetRetrieval.Auto(),
   rebalanceListener: RebalanceListener = RebalanceListener.noop,
@@ -39,9 +38,8 @@ final case class ConsumerSettings(
 
   def driverSettings: Map[String, AnyRef] =
     Map(
-      ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG    -> bootstrapServers.mkString(","),
-      ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG   -> "false",
-      ConsumerConfig.MAX_POLL_INTERVAL_MS_CONFIG -> maxPollInterval.toMillis.toString
+      ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG  -> bootstrapServers.mkString(","),
+      ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG -> "false"
     ) ++ autoOffsetResetConfig ++ properties
 
   def withBootstrapServers(servers: List[String]): ConsumerSettings =
@@ -72,19 +70,18 @@ final case class ConsumerSettings(
     copy(pollTimeout = timeout)
 
   /**
-   * The maximum delay between pulling chunks from a partition stream. This places an upper bound on the amount of time
-   * that a stream can be processing records. If no chunks are pulled before this timeout, then the entire consumer is
-   * considered failed and it will shutdown. When a group is in use the group will then rebalance in order to reassign
-   * the partitions to other members of the group.
+   * Set Kafka's `max.poll.interval.ms` configuration. See
+   * https://kafka.apache.org/documentation/#consumerconfigs_max.poll.interval.ms for more information.
    *
-   * This configuration is copied into Kafka's `max.poll.interval.ms` configuration. For more information see
-   * https://kafka.apache.org/documentation/#consumerconfigs_max.poll.interval.ms.
+   * Zio-kafka uses this value also to determine whether a stream stopped processing. If no chunks are pulled from a
+   * stream for this interval we consider the stream to be halted. When this happens we shutdown the entire consumer. In
+   * future versions of zio-kafka we may stop only the affected subscription.
    *
    * The default is 5 minutes. Make sure that all records from a single poll (see
    * https://kafka.apache.org/documentation/#consumerconfigs_max.poll.records) can be processed in this interval.
    */
-  def withMaxPollInterval(interval: Duration): ConsumerSettings =
-    copy(maxPollInterval = interval)
+  def withMaxPollInterval(maxPollInterval: Duration): ConsumerSettings =
+    withProperty(ConsumerConfig.MAX_POLL_INTERVAL_MS_CONFIG, maxPollInterval.toMillis.toString)
 
   def withProperty(key: String, value: AnyRef): ConsumerSettings =
     copy(properties = properties + (key -> value))
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
@@ -29,7 +29,6 @@ final case class ConsumerSettings(
   offsetRetrieval: OffsetRetrieval = OffsetRetrieval.Auto(),
   rebalanceListener: RebalanceListener = RebalanceListener.noop,
   restartStreamOnRebalancing: Boolean = false,
-  runloopTimeout: Duration = ConsumerSettings.defaultRunloopTimeout,
   fetchStrategy: FetchStrategy = QueueSizeBasedFetchStrategy()
 ) {
   private[this] def autoOffsetResetConfig: Map[String, String] = offsetRetrieval match {
@@ -94,16 +93,6 @@ final case class ConsumerSettings(
     withProperties(credentialsStore.properties)
 
   /**
-   * @param timeout
-   *   Internal timeout for each iteration of the command processing and polling loop, use to detect stalling. This
-   *   should be much larger than the pollTimeout and the time it takes to process chunks of records. If your consumer
-   *   is not subscribed for long periods during its lifetime, this timeout should take that into account as well. When
-   *   the timeout expires, the plainStream/partitionedStream/etc will fail with a [[Consumer.RunloopTimeout]].
-   */
-  def withRunloopTimeout(timeout: Duration): ConsumerSettings =
-    copy(runloopTimeout = timeout)
-
-  /**
    * @param maxPartitionQueueSize
    *   Maximum number of records to be buffered per partition. This buffer improves throughput and supports varying
    *   downstream message processing time, while maintaining some backpressure. Large values effectively disable
@@ -130,6 +119,5 @@ final case class ConsumerSettings(
 }
 
 object ConsumerSettings {
-  val defaultRunloopTimeout: Duration = 4.minutes
   val defaultCommitTimeout: Duration  = 15.seconds
 }
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
@@ -25,6 +25,7 @@ final case class ConsumerSettings(
   properties: Map[String, AnyRef] = Map.empty,
   closeTimeout: Duration = 30.seconds,
   pollTimeout: Duration = 50.millis,
+  maxPollInterval: Duration = 5.minutes,
   commitTimeout: Duration = ConsumerSettings.defaultCommitTimeout,
   offsetRetrieval: OffsetRetrieval = OffsetRetrieval.Auto(),
   rebalanceListener: RebalanceListener = RebalanceListener.noop,
@@ -38,8 +39,9 @@ final case class ConsumerSettings(
 
   def driverSettings: Map[String, AnyRef] =
     Map(
-      ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG  -> bootstrapServers.mkString(","),
-      ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG -> "false"
+      ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG    -> bootstrapServers.mkString(","),
+      ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG   -> "false",
+      ConsumerConfig.MAX_POLL_INTERVAL_MS_CONFIG -> maxPollInterval.toMillis.toString
     ) ++ autoOffsetResetConfig ++ properties
 
   def withBootstrapServers(servers: List[String]): ConsumerSettings =
@@ -68,6 +70,21 @@ final case class ConsumerSettings(
 
   def withPollTimeout(timeout: Duration): ConsumerSettings =
     copy(pollTimeout = timeout)
+
+  /**
+   * The maximum delay between pulling chunks from a partition stream. This places an upper bound on the amount of time
+   * that a stream can be processing records. If no chunks are pulled before this timeout, then the entire consumer is
+   * considered failed and it will shutdown. When a group is in use the group will then rebalance in order to reassign
+   * the partitions to other members of the group.
+   *
+   * This configuration is copied into Kafka's `max.poll.interval.ms` configuration. For more information see
+   * https://kafka.apache.org/documentation/#consumerconfigs_max.poll.interval.ms.
+   *
+   * The default is 5 minutes. Make sure that all records from a single poll (see
+   * https://kafka.apache.org/documentation/#consumerconfigs_max.poll.records) can be processed in this interval.
+   */
+  def withMaxPollInterval(interval: Duration): ConsumerSettings =
+    copy(maxPollInterval = interval)
 
   def withProperty(key: String, value: AnyRef): ConsumerSettings =
     copy(properties = properties + (key -> value))
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
@@ -62,11 +62,11 @@ object PartitionStreamControl {
     maxPollInterval: Duration,
     onConsumeTimeout: UIO[Unit]
   ): UIO[PartitionStreamControl] = {
-    val consumeTimeout = new TimeoutException(
-      s"No records were polled for more than $maxPollInterval for topic partition $tp. " +
-        "Set kafka configuration 'max.poll.interval.ms' to a higher value " +
-        "if processing a batch of records needs more time."
-    ) with NoStackTrace
+    val timeOutMessage = s"No records were polled for more than $maxPollInterval for topic partition $tp. " +
+      "Set kafka configuration 'max.poll.interval.ms' to a higher value " +
+      "if processing a batch of records needs more time."
+    val consumeTimeout = new TimeoutException(timeOutMessage) with NoStackTrace
+    val onTimeout = ZIO.logError(timeOutMessage) *> onConsumeTimeout
 
     for {
       _                   <- ZIO.logDebug(s"Creating partition stream ${tp.toString}")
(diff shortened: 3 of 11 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -84,6 +87,31 @@ private[internal] object PartitionStreamControl {
                  }.flattenTake
                    .chunksWith(_.tap(records => queueSize.update(_ - records.size)))
                    .interruptWhen(interruptionPromise)
+                   .pollTimeoutFail(
```
svroonland, line 90: This works based on the assumption that if a partition is resumed but no data is available on the topic, empty chunks will be passed through this point of the stream, right?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -84,6 +87,31 @@ private[internal] object PartitionStreamControl {
...
+        } yield {
+          val timeoutStream = ZStream
+            .fromQueue(queue)
+            .timeoutFail(e)(timeout)
+            .tapError(timeoutReached.fail)
+          val pollStream = stream.chunks
+            .tap(_ => queue.offer(()).unit)
+            .flattenChunks
+            .interruptWhen(timeoutReached)
+          timeoutStream *> pollStream
```
svroonland, line 112: What's the difference with `stream.chunks.timeoutFail(e)(timeout).flattenChunks` ?

erikvanoosten: Scaladoc of `timeoutFail`: Fails the stream with given error if it does not _produce_ a value after d duration.

To be added scaladoc of `pollTimeoutFail`: Fails the stream with given error if it does not _poll_ a value after d duration.

The difference is in italics.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
erikvanoosten: Not really. This code doesn't care what gets passed through. It only cares about passing _something_ through. Any item will reset the timer of the `timeoutStream`.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
svroonland: What exactly is polling in the semantics of ZChannels then..?

erikvanoosten: I was using the wrong word. The term is `pull` not `poll` :face_palm:

zio-kafka/src/main/scala/zio/kafka/utils/ExtraZStreamOps.scala
```diff
@@ -0,0 +1,40 @@
...
+                  ZChannel.write(in) *> ZChannel.fromZIO(runningTimer.interrupt) *> ZChannel.unwrap(timer.map(loop)),
+                (cause: Cause[ZNothing]) => ZChannel.refailCause(cause),
+                (_: Any) => ZChannel.unit
+              )
+
+            ZPipeline.fromChannel(loop(initialTimer).interruptWhen(p))
+          }
+        }
+      )
+  }
```
paulpdaniels, line 38: Now that it is working, it can be simplified to:

```scala
    ZPipeline.unwrapScoped(for {
      scope   <- ZIO.scope
      p       <- Promise.make[E, Unit]
      timer    = p.fail(e).delay(after).forkScoped.provideEnvironment(ZEnvironment[Scope](scope))
      initial <- timer
    } yield {
      def loop(f: Fiber[Nothing, Boolean]): ZChannel[Any, ZNothing, Chunk[In], Any, E, Chunk[In], Unit] =
        ZChannel.readWithCause(
          (in: Chunk[In]) => ZChannel.write(in) *> ZChannel.unwrap(f.interrupt *> timer.map(loop)),
          (cause: Cause[ZNothing]) => ZChannel.refailCause(cause),
          (_: Any) => ZChannel.unit
        )

      ZPipeline.fromChannel(loop(initial).interruptWhen(p))
    })
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -84,6 +96,9 @@ object PartitionStreamControl {
                  }.flattenTake
                    .chunksWith(_.tap(records => queueSize.update(_ - records.size)))
                    .interruptWhen(interruptionPromise)
+                   .consumeTimeoutFail(consumeTimeout)(maxPollInterval)
```
svroonland, line 99: What is the behavior if there is no new data available on the partition for the duration of `consumeTimeout`?

erikvanoosten: In that case the stream will just wait. However, let's add a test for that!

svroonland: That would be good to see, I can't quite determine the desired behavior from the code.

erikvanoosten: Thanks! So there was a actual bug for this scenario. It would also let the stream time out on a slow producer.

The bug is fixed, a test has been added.

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
```diff
@@ -92,6 +93,19 @@ object KafkaTestUtils {
         Serde.string
       )
 
+  /**
+   * A stream that produces messages to a Topic on a schedule for as long as it is running.
+   */
+  def scheduledProducer[R](
```
guizmaii, line 99:
```suggestion
  def scheduledProduce[R](
```

as it's not a Producer but a producing action?

zio-kafka/src/main/scala/zio/kafka/admin/AdminClient.scala
```diff
@@ -1212,15 +1212,9 @@ object AdminClient {
...
-
-      if (configs.nonEmpty) {
-        val _ = jn.configs(configs.asJava)
-      }
-
-      jn
-    }
+    def asJava: JNewTopic =
+      new JNewTopic(name, numPartitions, replicationFactor)
+        .configs(if (configs.nonEmpty) configs.asJava else null)
```
guizmaii, line 1217: It'd prefer we avoid the use of `null`
I prefer the previous version of the code.

zio-kafka/src/main/scala/zio/kafka/utils/ExtraZStreamOps.scala
```diff
@@ -0,0 +1,40 @@
+package zio.kafka.utils
+
+import zio._
+import zio.stream._
+
+object ExtraZStreamOps {
+
+  implicit class ZStreamOps[R, E, A](val stream: ZStream[R, E, A]) {
```
guizmaii, line 8:
```suggestion
  implicit final class ZStreamOps[R, E, A](private val stream: ZStream[R, E, A]) extends AnyVal {
```

svroonland, line 8: I'm not familiar enough with the ZChannel API to verify the behavior. Given that ZStreams are pull-based, won't the §(in: Chunk[A]) => ...  ` only be called when there is a downstream pull? So if downstream is no longer making progress, it will not be called. 

You're starting a timer under the assumption of this code being called push-based, it seems.

zio-kafka/src/main/scala/zio/kafka/admin/AdminClient.scala
erikvanoosten: Working with Java code sucks. At leas this makes it visible. But I don't case enough. I'll remove it.

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
```diff
@@ -92,6 +93,19 @@ object KafkaTestUtils {
...
+    topic: String,
+    schedule: Schedule[R, Any, Long]
+  ): ZStream[R with Producer, Throwable, RecordMetadata] =
+    ZStream
+      .fromSchedule(schedule)
+      .mapZIO { i =>
+        produceOne(topic, s"key$i", s"msg$i")
```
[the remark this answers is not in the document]
erikvanoosten: `fromSchedule` already creates a stream, so repeating again doesn't make sense.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
```diff
@@ -88,15 +89,25 @@ private[consumer] object RunloopAccess {
...
     } yield new RunloopAccess(runloopStateRef, partitionsHub, makeRunloop, diagnostics)
+
+  private def maxPollIntervalConfig(settings: ConsumerSettings): Duration = {
+    def defaultMaxPollInterval: AnyRef = ConsumerConfig
+      .configDef()
+      .defaultValues()
+      .getOrDefault(ConsumerConfig.MAX_POLL_INTERVAL_MS_CONFIG, "300000")
+    val maxPollIntervalStr = settings.properties
+      .getOrElse(ConsumerConfig.MAX_POLL_INTERVAL_MS_CONFIG, defaultMaxPollInterval)
+    maxPollIntervalStr.toString.toInt.millis
```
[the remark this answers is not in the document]
erikvanoosten: Unfortunately yes. You never know what you get exactly in Kafka properties. It can be a String, a java.lang.Integer or a java.lang.Long.

zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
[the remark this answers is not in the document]
svroonland: Ah you're creating a stream. 👍

zio-kafka/src/main/scala/zio/kafka/utils/ExtraZStreamOps.scala
erikvanoosten: Indeed, I feel you. It took several iterations before we realized that this is the correct way. This discovery was mostly driven by the unit tests. To be honest, I do not understand how ZStreams can be pull based and still make use of this API. I can only speculate that readWithCause is only invoked when there is demand. Anyways, the tests don't lie 😄

svroonland: Tests can be fault as well of course. Can we verify with Adam perhaps?

erikvanoosten: Since this is an _implicit_ class adding `final` and `private` don't make sense.

Adding `extends anyVal` make the method be inlined which is a good idea for small methods. However, this method is not small.

guizmaii: The goal here is to avoid an allocation of the `ZStreamOps` class, not anything else ;)

zio-kafka/src/main/scala/zio/kafka/utils/ExtraZStreamOps.scala
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * Fails the stream with given error if it is not consumed (pulled) from, for some duration.
+     *
+     * Also see [[zio.stream.ZStream#timeoutFail]] for failing the stream doesn't _produce_ a value.
```
guizmaii, line 13:
```suggestion
     * Also see [[zio.stream.ZStream.timeoutFail]] for failing the stream doesn't _produce_ a value.
```

So that we can CTRL+click on it to access the method directly

zio-kafka/src/main/scala/zio/kafka/utils/ExtraZStreamOps.scala
```diff
@@ -0,0 +1,40 @@
...
+    def consumeTimeoutFail[E1 >: E](e: => E1)(after: Duration): ZStream[R, E1, A] =
+      // For every incoming chunk a timer is started. When the chunk is consumed, the timer is stopped by interrupting
+      // it. When the timer completes, the stream gets interrupted.
+      stream.via(
+        ZPipeline.unwrapScoped(
+          for {
+            scope <- ZIO.scope
+            p     <- Promise.make[E1, Unit]
+          } yield {
+            val timer = p.fail(e).delay(after).forkScoped.provideEnvironment(ZEnvironment[Scope](scope))
```
guizmaii, line 24: This can be simplified a bit:
```suggestion
          for {
            scope <- ZIO.environment[Scope]
            p     <- Promise.make[E1, Unit]
          } yield {
            val timer = p.fail(e).delay(after).forkScoped.provideEnvironment(scope)
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
```diff
@@ -88,15 +89,25 @@ private[consumer] object RunloopAccess {
...
+  }
```
guizmaii, line 112: To me, this isn't safe enough. The `toInt` can fail too.
Here's what I'd like to suggest:
```suggestion
  private def maxPollIntervalConfig(settings: ConsumerSettings): Duration = {
    // 5 minutes.
    // See https://docs.confluent.io/platform/current/installation/configuration/consumer-configs.html#max-poll-interval-ms
    def kafkaDefaultValue = 5.minutes

    def defaultMaxPollInterval: AnyRef =
      ConsumerConfig
        .configDef()
        .defaultValues()
        .getOrDefault(ConsumerConfig.MAX_POLL_INTERVAL_MS_CONFIG, kafkaDefaultValue.toMillis)

    try
      settings.properties
        .getOrElse(ConsumerConfig.MAX_POLL_INTERVAL_MS_CONFIG, defaultMaxPollInterval)
        .toString
        .toInt
        .millis
    catch {
      case NonFatal(_) => kafkaDefaultValue
    }
  }
```

zio-kafka/src/main/scala/zio/kafka/utils/ExtraZStreamOps.scala
erikvanoosten: Indeed. I am mistaken; extending AnyVal doesn't mean the method is inlined (see https://docs.scala-lang.org/overviews/core/value-classes.html#extension-methods).

AnyVal Classes cannot be extended so they are implicitly final.

Also, I am reading (and clearly remember) that the class must have a single public parameter. However, when I try it out the compiler doesn't complain. So perhaps this rule no longer applies.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -54,8 +58,16 @@ object PartitionStreamControl {
   private[internal] def newPartitionStream(
     tp: TopicPartition,
     commandQueue: Queue[RunloopCommand],
-    diagnostics: Diagnostics
-  ): UIO[PartitionStreamControl] =
+    diagnostics: Diagnostics,
+    maxPollInterval: Duration,
+    onConsumeTimeout: UIO[Unit]
+  ): UIO[PartitionStreamControl] = {
+    val consumeTimeout = new TimeoutException(
```
guizmaii, line 65:
```suggestion
    def consumeTimeout = new TimeoutException(
```

zio-kafka/src/main/scala/zio/kafka/utils/ExtraZStreamOps.scala
erikvanoosten: https://discord.com/channels/629491597070827530/630498701860929559/1129744143246164088

zio-kafka/src/main/scala/zio/kafka/utils/ExtraZStreamOps.scala
erikvanoosten: It has to be `#` for scala 2. In Scala 3 you can also use `.`.

zio-kafka/src/main/scala/zio/kafka/utils/ExtraZStreamOps.scala
erikvanoosten: Nice!

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
erikvanoosten: Fixed, but a little bit differently in https://github.com/zio/zio-kafka/pull/960/commits/6830e938d49884e722cb15fdfe5b61775e6b9dfe

zio-kafka/src/main/scala/zio/kafka/utils/ExtraZStreamOps.scala
guizmaii: I'm using Scala 2 and `#` doesn't work for me 🤔

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
```diff
@@ -102,12 +102,18 @@ private[consumer] object RunloopAccess {
...
       .defaultValues()
-      .getOrDefault(ConsumerConfig.MAX_POLL_INTERVAL_MS_CONFIG, "300000")
-    val maxPollIntervalStr = settings.properties
-      .getOrElse(ConsumerConfig.MAX_POLL_INTERVAL_MS_CONFIG, defaultMaxPollInterval)
-    maxPollIntervalStr.toString.toInt.millis
+      .getOrDefault(ConsumerConfig.MAX_POLL_INTERVAL_MS_CONFIG, Int.box(fallbackDefault))
+      .asInstanceOf[Integer]
+    settings.properties
+      .get(ConsumerConfig.MAX_POLL_INTERVAL_MS_CONFIG)
+      .flatMap(_.toString.toIntOption) // Ignore invalid
```
guizmaii, line 115: `toIntOption` doesn't exist in Scala 2.12, thus my suggestion ;)

zio-kafka/src/main/scala/zio/kafka/utils/ExtraZStreamOps.scala
```diff
@@ -0,0 +1,40 @@
+package zio.kafka.utils
+
+import zio._
+import zio.stream._
+
+object ExtraZStreamOps {
+
+  implicit class ZStreamOps[R, E, A](val stream: ZStream[R, E, A]) extends AnyVal {
```
guizmaii, line 8: `final` and `private` are missing
```suggestion
  implicit final class ZStreamOps[R, E, A](private val stream: ZStream[R, E, A]) extends AnyVal {
```

See for example: https://github.com/zio/zio/blob/bdc08ea21dce1485e9280ba76ee9d74d8e0b202f/core/shared/src/main/scala/zio/ZIO.scala#L5131

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
erikvanoosten: O haha!

erikvanoosten: Fixed by using scala compat.

zio-kafka/src/main/scala/zio/kafka/utils/ExtraZStreamOps.scala
erikvanoosten: Hmmm. It doesn't work for me either. The documentation must be wrong :(
I'll change it.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
erikvanoosten: That didn't work. I changed it completely and now it delegates to the Kafka library. https://github.com/zio/zio-kafka/pull/960/commits/542bd22b9b16161571a9c495541ddcbcb79d8c96

zio-kafka/src/main/scala/zio/kafka/utils/ExtraZStreamOps.scala
erikvanoosten: Okay, this is against what I learned in the past. O well, times change...

zio-kafka/src/main/scala/zio/kafka/utils/ExtraZStreamOps.scala
erikvanoosten: Okay, this is against what I learned in the past. O well, times change...

zio-kafka/src/main/scala/zio/kafka/utils/ExtraZStreamOps.scala
erikvanoosten: From @adamgfraser on [Discord](https://discord.com/channels/629491597070827530/630498701860929559/1129814749027250247):

> I think there are a few things to think about with your implementation. Basically what you are doing is starting this timer immediately before the upstream writes to the downstream and then interrupting it immediately after execution resumes after the write, which is fair as far as it goes because you can think of control passing to the downstream when the upstream writes to it and then passing back to the upstream when the downstream reads from it. However, there are a couple of situations to consider. One is if the downstream never reads. In this case the timer never starts. Not sure whether this is a problem. Another is that even if the timer does start, failure isn't signaled until the downstream reads again after the timeout.If the downstream never reads again action is never taken based on the failure even though the timeout has expired. Again, not sure if this is a problem for your use case. You might alternatively think of providing some action you want to run if there isn't a pull within a certain interval.

erikvanoosten: In https://github.com/zio/zio-kafka/pull/960/commits/a0cd35aa5b522d93e381cce1a1303137b7589ac8 you will find a completely revamped implementation which is based on `ZStream.toPull`. This implementation does not need to start a background ZIO with a timer on each chunk. Instead, it times the time between pulls and fails the stream when the next pull takes too long to arrive.
This is not ideal yet. We want to fail the stream as soon as the timeout occurs, not when the delayed pull arrives.

guizmaii:
>  In addition, we shut down the entire consumer. This will help the broker to detect slow consumers. Ideally we would only stop the subscription that the stream belongs to. Unfortunately, we do not maintain a topic belongs-to subscription relation.

I'm not a fan of this. To me, the purpose of these changes was to avoid stopping the full Consumer and only stop the subscription.
What does your PR improve over the previous `RunlooptTimeout` mechanism if both are killing the Consumer? 🤔

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -70,6 +71,21 @@ final case class ConsumerSettings(
...
+   * that a stream can be processing records. If no chunks are pulled before this timeout, then the entire consumer is
+   * considered failed and it will shutdown. When a group is in use the group will then rebalance in order to reassign
+   * the partitions to other members of the group.
+   *
+   * This configuration is copied into Kafka's `max.poll.interval.ms` configuration. For more information see
+   * https://kafka.apache.org/documentation/#consumerconfigs_max.poll.interval.ms.
+   *
+   * The default is 5 minutes. Make sure that all records from a single poll (see
+   * https://kafka.apache.org/documentation/#consumerconfigs_max.poll.records) can be processed in this interval.
+   */
```
guizmaii, line 85: Can we document the `ConsumerSettings::maxPollInterval` field instead of this method, please?

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -25,11 +25,11 @@ final case class ConsumerSettings(
   properties: Map[String, AnyRef] = Map.empty,
   closeTimeout: Duration = 30.seconds,
   pollTimeout: Duration = 50.millis,
+  maxPollInterval: Duration = 5.minutes,
```
guizmaii, line 28: I don't like that we duplicate Kafka configurations in our configuration.
I think it's confusing for users.
What if they configure it here and in the `properties` map?
We're also duplicating the default value, which I don't like either.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -54,8 +58,16 @@ object PartitionStreamControl {
...
-  ): UIO[PartitionStreamControl] =
+    diagnostics: Diagnostics,
+    maxPollInterval: Duration,
+    onConsumeTimeout: UIO[Unit]
+  ): UIO[PartitionStreamControl] = {
+    val timeOutMessage = s"No records were polled for more than $maxPollInterval for topic partition $tp. " +
+      "Use ConsumerSettings.withMaxPollInterval to set a longer interval if processing a batch of records " +
+      "needs more time."
+    val consumeTimeout = new TimeoutException(timeOutMessage) with NoStackTrace
+    val onTimeout      = ZIO.logError(timeOutMessage) *> onConsumeTimeout
```
guizmaii, line 69: If everything goes well, these things should never be instantiated, no?

```suggestion
    def timeOutMessage = s"No records were polled for more than $maxPollInterval for topic partition $tp. " +
      "Use ConsumerSettings.withMaxPollInterval to set a longer interval if processing a batch of records " +
      "needs more time."
    def consumeTimeout = new TimeoutException(timeOutMessage) with NoStackTrace
    def onTimeout      = ZIO.logError(timeOutMessage) *> onConsumeTimeout
```

zio-kafka/src/main/scala/zio/kafka/utils/ExtraZStreamOps.scala
```diff
@@ -0,0 +1,89 @@
...
+     * Will also run workflow `onTimeout` when the timeout occurs.
+     *
+     * Note: since the wrapped stream only starts when pulling starts, we never time out waiting for the first pull
+     *
+     * Also see [[zio.stream.ZStream.timeoutFail]] for failing the stream doesn't _produce_ a value.
+     */
+    // noinspection SimplifySleepInspection
+    def consumeTimeoutFail[E1 >: E](
+      e: => E1
+    )(after: Duration)(onTimeout: ZIO[R, Nothing, Any])(implicit trace: Trace): ZStream[R, E1, A] =
```
guizmaii, line 29: Avoids allocating things if the timeout never happens

```suggestion
    )(after: Duration)(onTimeout: => ZIO[R, Nothing, Any])(implicit trace: Trace): ZStream[R, E1, A] =
```

erikvanoosten:
> I'm not a fan of this. To me, the purpose of these changes was to avoid stopping the full Consumer and only stop the subscription.
> What does your PR improve over the previous `RunlooptTimeout` mechanism if both are killing the Consumer? 🤔

The improvements are:
- the offending stream gets interrupted,
- we can stop the subscription only in a later PR.

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
erikvanoosten: I was wondering about this as well. I like the helper approach. We can use that.

zio-kafka/src/main/scala/zio/kafka/utils/ExtraZStreamOps.scala
```diff
@@ -0,0 +1,89 @@
...
+          }
+
+        for {
+          producerDone <- Promise.make[Nothing, Unit]
+          stateRef     <- Ref.make[State](WaitingForProducer(producerDone))
+          p            <- Promise.make[E1, Nothing]
+          _            <- deadlineChecker(stateRef, p.fail(e) *> onTimeout).interruptible.forkScoped
+        } yield {
+          val timedOutState = (ZIO.fail(e), ConsumerTimedOut)
+          val illegalState  = (ZIO.die(new IllegalStateException()), ConsumerTimedOut)
```
guizmaii, line 53: We need an error message in `new IllegalStateException()`

erikvanoosten: It never happens. Perhaps we should change it for `new AssertionError` instead.

zio-kafka/src/main/scala/zio/kafka/utils/ExtraZStreamOps.scala
```diff
@@ -0,0 +1,89 @@
...
+     * Fails the stream with given error if it is not consumed (pulled) from, for some duration.
+     *
+     * Will also run workflow `onTimeout` when the timeout occurs.
+     *
+     * Note: since the wrapped stream only starts when pulling starts, we never time out waiting for the first pull
+     *
+     * Also see [[zio.stream.ZStream.timeoutFail]] for failing the stream doesn't _produce_ a value.
+     */
+    // noinspection SimplifySleepInspection
+    def consumeTimeoutFail[E1 >: E](
```
guizmaii, line 27: Why didn't you just copy Adam's solution https://discord.com/channels/629491597070827530/629497941719121960/1130198205120905348? 🤔

erikvanoosten: I had to make changes because:
- It doesn't interrupt pulls when the timeout already occurred.
- It doesn't work with the test clock.
- It assumes acquire/release semantics which we don't need. We only need to interrupt the stream. In addition there is an 'onTimeout' callback, this is different from 'release' callback which is also called when the scope closes without a timeout.
- Because `onTimeout` is only called from the checker fiber, we do not need the `once` method any more.
- We already have the stream, so a `uses` parameter doesn't make sense in our use case.
- There is no need to interrupt the checked fiber because it is tied to the scope.

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
erikvanoosten: BTW, we do interpret max.poll.interval differently in zio-kafka. That is why this description is needed.

erikvanoosten: Fixed in https://github.com/zio/zio-kafka/pull/960/commits/7f1d196b1d42b76969ccf485b92bdb3bd0cd2b65

zio-kafka/src/main/scala/zio/kafka/utils/ExtraZStreamOps.scala
erikvanoosten: Fixed in https://github.com/zio/zio-kafka/pull/960/commits/7f1d196b1d42b76969ccf485b92bdb3bd0cd2b65

zio-kafka/src/main/scala/zio/kafka/utils/ExtraZStreamOps.scala
erikvanoosten: The current version contains another implementation, this time based on an example from Adam.

svroonland:
A comment on the PR description:

> Translated to zio-kafka 'halting' means that a stream does not consume more data for longer than max.poll.interval.ms.
That is not true in case there are no new messages on the topic.

svroonland:
To consider as an alternative: with the prefetch queue, could we detected stalled partition streams by checking when the queue size last changed or something?

erikvanoosten:
> A comment on the PR description:
> 
> > Translated to zio-kafka 'halting' means that a stream does not consume more data for longer than max.poll.interval.ms.
> > That is not true in case there are no new messages on the topic.

Indeed. I will update the description.

> To consider as an alternative: with the prefetch queue, could we detected stalled partition streams by checking when the queue size last changed or something?

I love this idea. The stream operator that is introduced in this PR looks quite heavy.

Let's put this PR on hold while we explore that option.

erikvanoosten:
Superseded by #1009.

PR #962, opened 2023-07-02
Only calculate initial state once

erikvanoosten:


svroonland (approved):
Not so much about calculating, since it references the same val, but a single declaration of the state used initially

erikvanoosten:
> Not so much about calculating, since it references the same val, but a single declaration of the state used initially

Yes, I'll change the commit message a bit.

guizmaii:
Hum. It's was already calculated only once as it was a `val`

PR #963, opened 2023-07-02
Prevent asInstanceOf

erikvanoosten:
Add a bit more type safety by removing the need for `asInstanceOf`.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
@@ -30,7 +30,7 @@ private[consumer] final class RunloopAccess private (
   diagnostics: Diagnostics
 ) {
 
-  private def runloopZIO[E](
+  private def withRunloopZIO[E](
     requireRunning: Boolean
   )(whenRunning: Runloop => IO[E, Unit]): IO[E, Unit] =
     runloopStateRef.updateSomeAndGetZIO {
@@ -44,7 +44,7 @@ private[consumer] final class RunloopAccess private (
   /**
    * No need to call `Runloop::stopConsumption` if the Runloop has not been started or has been stopped.
    */
-  def stopConsumption: UIO[Unit] = runloopZIO(requireRunning = false)(_.stopConsumption)
+  def stopConsumption: UIO[Unit] = withRunloopZIO(requireRunning = false)(_.stopConsumption)
 
   /**
    * We're doing all of these things in this method so that the interface of this class is as simple as possible and
@@ -58,9 +58,9 @@ private[consumer] final class RunloopAccess private (
     for {
       stream <- ZStream.fromHubScoped(partitionHub)
       // starts the Runloop if not already started
-      _ <- runloopZIO(requireRunning = true)(_.addSubscription(subscription))
+      _ <- withRunloopZIO(requireRunning = true)(_.addSubscription(subscription))
       _ <- ZIO.addFinalizer {
-             runloopZIO(requireRunning = false)(_.removeSubscription(subscription)) <*
+             withRunloopZIO(requireRunning = false)(_.removeSubscription(subscription)) <*
                diagnostics.emit(Finalization.SubscriptionFinalized)
            }
     } yield stream
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
@@ -30,30 +30,21 @@ private[consumer] final class RunloopAccess private (
   diagnostics: Diagnostics
 ) {
 
-  private def runloop[E, A](
+  private def runloop[E](
     requireRunning: Boolean
-  )(whenRunning: Runloop => IO[E, A], whenNotRunning: IO[E, A]): IO[E, A] =
+  )(whenRunning: Runloop => IO[E, Unit]): IO[E, Unit] =
     runloopStateRef.updateSomeAndGetZIO {
       case RunloopState.NotStarted if requireRunning => makeRunloop.map(RunloopState.Started.apply)
     }.flatMap {
-      case RunloopState.NotStarted       => whenNotRunning
+      case RunloopState.NotStarted       => ZIO.unit
       case RunloopState.Started(runloop) => whenRunning(runloop)
-      case RunloopState.Finalized        => whenNotRunning
+      case RunloopState.Finalized        => ZIO.unit
     }
-
-  private def withRunloop[E, A](whenRunning: Runloop => IO[E, A]): IO[E, A] =
-    runloop(requireRunning = true)(
-      whenRunning = whenRunning,
-      whenNotRunning = ZIO.die(new IllegalStateException("Trying to use a consumer that is already finalized"))
-    )
-
-  private def whenRunloop[E](whenRunning: Runloop => IO[E, Unit]): IO[E, Unit] =
-    runloop(requireRunning = false)(whenRunning, whenNotRunning = ZIO.unit)
 
   /**
    * No need to call `Runloop::stopConsumption` if the Runloop has not been started or has been stopped.
    */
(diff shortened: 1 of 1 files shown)


guizmaii (commented):
I'd prefer to leave the code as it was. It's simpler and easier to read.
To me, the added complexity is not worth this change
There's no type-safety issue here. The `asInstanceOf` are just here to please the compiler.

erikvanoosten:
Relax!
Its all your code and ideas, only a few things now have a name which should make it *more* readable.
The amount of complexity is exactly the same. Please look at the end result and not the diff.

> There's no type-safety issue here. The `asInstanceOf` are just here to please the compiler.

There is and no that is not true. Right now it only goes right by accident; the method is always used in such a way that the types line up.

Here is a minimizer that show what happens when they are not lined up:

```scala
import zio._
import zio.Console._

object App extends ZIOAppDefault {

  private def runloop[A](p: Boolean)(f: UIO[A]): UIO[A] = if (p) f else ZIO.unit.asInstanceOf[UIO[A]]

  def run: ZIO[Any, Throwable, ExitCode] = for {
    value <- runloop(false)(ZIO.succeed("Hello, world!"))
    _ <- printLine(value)
  } yield ExitCode.success
}
```

It fails with:
```
timestamp=2023-07-03T07:32:43.676441Z level=ERROR thread=#zio-fiber-1 message="" cause="Exception in thread "zio-fiber-4" java.lang.ClassCastException: class scala.runtime.BoxedUnit cannot be cast to class java.lang.String (scala.runtime.BoxedUnit is in unnamed module of loader 'app'; java.lang.String is in module java.base of loader 'bootstrap')
	at replicator.App.run(App.scala:11)"
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
```diff
@@ -29,21 +29,34 @@ private[consumer] final class RunloopAccess private (
...
+      case RunloopState.NotStarted if requireRunning => makeRunloop.map(RunloopState.Started.apply)
+    }.flatMap {
+      case RunloopState.NotStarted       => whenNotRunning
+      case RunloopState.Started(runloop) => whenRunning(runloop)
+      case RunloopState.Finalized        => whenNotRunning
     }
 
+  private def withRunloopZIO[E, A](whenRunning: Runloop => IO[E, A]): IO[E, A] =
+    runloop(
+      requireRunning = true,
```
yarosman, line 48: therefore it's better to add naming for all parameters

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
```diff
@@ -29,21 +29,34 @@ private[consumer] final class RunloopAccess private (
...
-      case RunloopState.NotStarted       => ZIO.unit.asInstanceOf[IO[E, A]]
-      case RunloopState.Started(runloop) => f(runloop)
+      case RunloopState.NotStarted if requireRunning => makeRunloop.map(RunloopState.Started.apply)
+    }.flatMap {
+      case RunloopState.NotStarted       => whenNotRunning
+      case RunloopState.Started(runloop) => whenRunning(runloop)
+      case RunloopState.Finalized        => whenNotRunning
     }
 
+  private def withRunloopZIO[E, A](whenRunning: Runloop => IO[E, A]): IO[E, A] =
```
yarosman, line 46: Does it necessary to add `ZIO` suffix everywhere?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
```diff
@@ -29,21 +29,34 @@ private[consumer] final class RunloopAccess private (
   makeRunloop: UIO[Runloop],
   diagnostics: Diagnostics
 ) {
-  private def runloop(shouldStartIfNot: Boolean): UIO[RunloopState] =
+
+  private def runloop[E, A](
+    requireRunning: Boolean,
+    whenRunning: Runloop => IO[E, A],
```
yarosman, line 35: It would be better to separate call-by-name and call-by-value args, like it was in the previous version of `withRunloopZIO`

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
erikvanoosten: Neh, not a requirement at all.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
erikvanoosten: Sure, that is not an issue. Why is that better?

guizmaii:
@erikvanoosten 
> Right now it only goes right by accident

Indeed, my bad.
Then the simplest fix is this:
```scala
  private def withRunloopZIO[E, A](shouldStartIfNot: Boolean)(f: Runloop => UIO[Unit]): UIO[Unit] =
    runloop(shouldStartIfNot).flatMap {
      case RunloopState.Finalized        => ZIO.unit
      case RunloopState.NotStarted       => ZIO.unit
      case RunloopState.Started(runloop) => f(runloop)
    }
```

It's good enough for now.

erikvanoosten:
> Then the simplest fix is this:
Yes, that works as well.
Done in https://github.com/zio/zio-kafka/pull/963/commits/e7409401b19503e563a1ed918253ac97b59534d1

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
erikvanoosten: It is convention taken over from ZIO 2 itself.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
erikvanoosten: Its restored.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
```diff
@@ -29,21 +29,22 @@ private[consumer] final class RunloopAccess private (
...
+
+  private def withRunloopZIO[E](
+    requireRunning: Boolean
+  )(whenRunning: Runloop => IO[E, Unit]): IO[E, Unit] =
     runloopStateRef.updateSomeAndGetZIO {
-      case RunloopState.NotStarted if shouldStartIfNot => makeRunloop.map(RunloopState.Started.apply)
-    }
-  private def withRunloopZIO[E, A](shouldStartIfNot: Boolean)(f: Runloop => IO[E, A]): IO[E, A] =
-    runloop(shouldStartIfNot).flatMap {
-      case RunloopState.Finalized        => ZIO.unit.asInstanceOf[IO[E, A]]
```
svroonland, line 38: Played around with the types here and I can see where the problem comes from, but casting `()` to an `A` doesn't appear to me to be the way to go. You don't actually need the `A` here, just replace it with this and it works fine:


```
  private def withRunloopZIO[E](shouldStartIfNot: Boolean)(f: Runloop => IO[E, Unit]): IO[E, Unit] =
    runloop(shouldStartIfNot).flatMap {
      case RunloopState.Finalized        => ZIO.unit
      case RunloopState.NotStarted       => ZIO.unit
      case RunloopState.Started(runloop) => f(runloop)
    }
```

svroonland (approved):

svroonland:
Since this is @guizmaii's code, let's wait for his approval as well.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
erikvanoosten: Yes, in the end that is what I did as well. In addition, I merged `runloop` and `withRunloopZIO` because `runloop` by itself didn't make sense.

guizmaii (approved):

PR #970, opened 2023-07-05
Add experimental alternative fetch strategies

erikvanoosten:
1. `ManyPartitionsQueueSizeBasedFetchStrategy`, a variation on the default `QueueSizeBasedFetchStrategy` which limits total memory usage.
2. `PredictiveFetchStrategy` an improved predictive fetching strategy (compared to the predictive strategy from zio-kafka 2.3.x) which uses history to calculate the average number of polls the stream needed to process, and uses that to estimate when the stream needs more data.

To do:
 - [x] Add unit tests

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/fetch/PredictiveFetchStrategy.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/fetch/PredictiveFetchStrategy.scala
@@ -1,0 +1,53 @@
+package zio.kafka.consumer.fetch
+
+import org.apache.kafka.common.TopicPartition
+import zio.kafka.consumer.internal.PartitionStream
+import zio.{ Chunk, ZIO }
+
+import scala.collection.mutable
+
+/**
+ * A fetch strategy that predicts when a stream needs more data by analyzing its history.
+ *
+ * The prediction is based on the average number of polls the stream needed to process data in the recent past. In
+ * addition, a stream can always fetch when it is out of data.
+ *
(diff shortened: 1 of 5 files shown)


erikvanoosten:
`ManyPartitionsQueueSizeBasedFetchStrategy` has been split of to #1281.

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/PredictiveFetchStrategy.scala
```diff
@@ -0,0 +1,53 @@
+package zio.kafka.consumer.fetch
+
+import org.apache.kafka.common.TopicPartition
+import zio.kafka.consumer.internal.PartitionStream
+import zio.{ Chunk, ZIO }
+
+import scala.collection.mutable
+
+/**
```
svroonland, line 9: Number of polls is quite a discrete measurement, what do you think of converting this into a running average (like exponentially weighed moving average) of the number of records dequeued each poll?

erikvanoosten: That will work against people that use something like Grafana. First you need to collect and sum the raw counters from each instance of your service. Then, and only then, you can calculate a running average, an integral, or whatever other operation.

erikvanoosten:
> @erikvanoosten Shall we close this PR? It's well over a year old now

Sure. I have put a lot of effort in `PredictiveFetchStrategy` but if nobody is going to use it, it doesn't make sense to keep it.

PR #981, opened 2023-07-15
Fix CI

guizmaii:


erikvanoosten (commented):
I fixed this also in #960 which is awaiting a review.

guizmaii:
> I fixed this also in https://github.com/zio/zio-kafka/pull/960 which is awaiting a review.

Would have been better to make a dedicated PR like this one

PR #983, opened 2023-07-15
Try to implement the "one timer per thread"

guizmaii:


svroonland (commented):
Do we have reason to assume ZIO#delay is inefficient?

erikvanoosten:
@guizmaii note that there are 2 more tests as recommended by Adam in #960.

guizmaii:
@erikvanoosten As explained in Discord, this code generates behaviours I can't get my head around. 

The `ref.updateSomeZIO { Some(...) => <code>.as(None) }` is executed twice. I don't understand how that's possible.
Also, when the `p.fail(e)` is called, that doesn't interrupt the Stream while the `p` is correctly completed. Same, I have no idea why 😖

erikvanoosten:
I started with a completely new implementation and also their I do not see any effect of calling p.fail 😞 
Will research further...

erikvanoosten:
@guizmaii With help from Adam I got a working version. See the latest commits in #960.

PR #986, opened 2023-07-19
Fix: Runloop might be suspended when assigned with many partitions

yaarix:
- The Runloop command queue is a bounded queue, with a fixed size of 1024
 - The Runloop stream takes commands from the queue, process them and offer a new `Poll` command into the stream (when applicable).
 - The `.offer` can suspend the fiber when the command queue is already full. this, in turn, makes the entire stream freeze. the underlying kafka consumer will eventually (after `max.poll.interval.ms`) be dropped from the group without any notification to the zstream.
 - to fix this, we can change the command queue from bounded to unbounded which will not suspend the fiber that adds the `Poll` command

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -554,9 +554,6 @@ private[consumer] object Runloop {
     ) extends RebalanceEvent
   }
 
-  // Internal parameters, should not be necessary to tune
-  private final val commandQueueSize: Int = 1024
-
   def make(
     hasGroupId: Boolean,
     consumer: ConsumerAccess,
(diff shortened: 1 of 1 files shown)


svroonland:
👍 

As expressed on Discord, unbounded memory usage is something to be avoided in general, but in this case it's preferrable over freezing. Should this lead to trouble, it's at least easily diagnosed as OOM errors.

yaarix:
@svroonland any idea why the benchmark fails? it fails on lack of permissions

guizmaii (approved):

erikvanoosten (approved):

guizmaii:
@yaarix benchmarks don't work on forks

svroonland:
@yaarix It's been a long time since this issue, but have you actually experienced this issue or was it just a theoretical one?

yaarix:
I have experienced it and merged a fix a long time ago

svroonland:
Thanks for the quick response. Alright, then we'll cancel plans for #1386 (completely forgot about this issue here) unless we have a good idea what exactly caused it and how we can mitigate that.

PR #993, opened 2023-07-21
Fix signature of Producer#produceAll

vladimirkl:
This tiny PR removes `Producer` service from environment of `Producer#produceAll` - probably it was left there by accident. Without fix method cannot be used with Module 2.0 pattern


erikvanoosten (approved):

erikvanoosten:
Thanks for attention to detail!

PR #995, opened 2023-07-22
Document setting pollTimeout

erikvanoosten:


guizmaii:
@erikvanoosten It's still a draft PR, normal?

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -67,6 +67,14 @@ final case class ConsumerSettings(
   def withOffsetRetrieval(retrieval: OffsetRetrieval): ConsumerSettings =
     copy(offsetRetrieval = retrieval)
 
+  /**
+   * The maximum time to block while polling the Kafka consumer. The Kafka consumer will return earlier when the maximum
```
guizmaii, line 71: Isn't this `pollTimeout` property a Kafka client property? If yes, shouldn't we put the link to the kafka documentation here, as you do for the `max.poll.records`?

I'm not sure documenting twice this property (once here and once in the Kafka documentation is a good idea.

I'm fine with keeping the additional information which are specific to zio-kafka, like:
> The default is `50ms` which to good for low latency applications. Set this higher, e.g. `500ms` for better  throughput.

But I don't think we should explain something already explained in the Kafka documenation

Finally, I think it'd be better to document the field of the case class over this method

erikvanoosten:
> @erikvanoosten It's still a draft PR, normal?

Oops.

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
erikvanoosten: > Isn't this pollTimeout property a Kafka client property? If yes, shouldn't we put the link to the kafka documentation here, as you do for the max.poll.records?

Sure, that is a good idea.

erikvanoosten: Ah! `pollTimeout` is _not_ a Kafka configuration!

I tried to find pollTimeout in the Kafka documentation. It is in the javadocs: https://kafka.apache.org/35/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#poll(java.time.Duration) where it says:

> Parameters:
>     timeout - The maximum time to block (must not be greater than [Long.MAX_VALUE](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Long.html#MAX_VALUE) milliseconds)

svroonland (approved):

guizmaii (approved):

Issue #988, opened 2023-07-19
Duplicate events appear when using single Consumer with multiple subscriptions


myazinn:
Here's a test for [SubscriptionsSpec](https://github.com/zio/zio-kafka/blob/master/zio-kafka-test/src/test/scala/zio/kafka/consumer/SubscriptionsSpec.scala).
In this test, I subscribe to `topic1` and `topic2` using the same `Consumer` but within different streams, and then publish one message to each topic. However, for some reason I get the same event twice from `topic1`.

```scala
    test("consumes from two topic subscriptions without duplicates") {
      def plainStream(topic: String) =
        Consumer.plainStream(Subscription.topics(topic), Serde.string, Serde.string)

      // .delay imitates long processing
      def readCount(topic: String)(counter: Ref[Int]) =
        plainStream(topic).runForeach(_.offset.commit.delay(500.millis) *> counter.update(_ + 1))

      def readOne(topic: String) =
        plainStream(topic).take(1).runDrain

      def writeSingleTo(topics: String*) =
        ZIO.foreachParDiscard(topics)(topic => produceOne(topic, s"key-$topic", s"value-$topic"))

      for {
        topic1 <- randomTopic
        topic2 <- randomTopic
        client <- randomClient
        group  <- randomGroup

        c = consumer(client, Some(group))

        counter <- Ref.make[Int](0)
        _ <- (readCount(topic1)(counter).fork *> readOne(topic2).zipPar(writeSingleTo(topic1, topic2)))
               .zipPar(ZIO.sleep(3.seconds)) // give consumer some time to poll events
               .provideSome[Producer with Kafka](c)

        count <- counter.get
      } yield assertTrue(count == 1)
    }
```
This test fails because `count` is equal to 2.
I couldn't find the root cause yet, but it appears that when we stop reading from `topic2` there's some internal restart, which causes events to be read from Kafka again event though there're some messages pending in memory. Because of it, we might consume the same message several times which is definitely wrong in this case.

erikvanoosten:
What clock is this test using?

erikvanoosten:
Do I read it correctly that this test reads from topic2 and then produces that record to topic1 _and_ topic2. If I read that right, this creates a duplicate on topic2.

I am missing commits on topic2. So consuming from topic2 again might cause duplicates when using the same consumergroup. This might happen depending on the implementation of `randomGroup`.

guizmaii:
[...]

PR #996, opened 2023-07-22
Prevent doube logging in tests

erikvanoosten:
Previously all logging in tests is doubled. This change will make un-double logging.

TODO: validate if this is the right way to fix the problem.

Related to #988.


erikvanoosten:
@guizmaii Another way to solve duplicate logging is by rewriting the file to:

```scala
abstract class ZIOSpecDefaultSlf4j extends ZIOSpecDefault {
  private val loggerLayer = SLF4J.slf4j
  override def aspects: Chunk[TestAspectAtLeastR[TestEnvironment]] =
    super.aspects :+ TestAspect.fromLayer(zio.Runtime.removeDefaultLoggers >>> loggerLayer)
}
```

🤷 please, you decide which solution to take.

erikvanoosten:
This PR is no longer relevant.

PR #1009, opened 2023-08-05
Add processing timeout at the stream level

erikvanoosten:
Kafka's process to detect halted consumers is crude. Because it is based on polls, it works per consumer. Here we make it more fine-grained by doing slow consumer detection per stream.

For the Kafka client 'halting' means poll is not called for longer than `max.poll.interval.ms`. Translated to zio-kafka 'halting' means that a stream does not consume any records for `max.poll.interval.ms` or longer, while records are available.

The halted stream is interrupted so that it can fail properly. In addition, we shut down the entire consumer. This will help the broker to detect slow consumers. Ideally we would only stop the subscription that the stream belongs to. Unfortunately, we do not maintain a topic belongs-to subscription relation. Therefore, stopping only the affected subscription is left for a future change.

As discussed on [Discord](https://discord.com/channels/629491597070827530/629497941719121960/1114868628156858400).

Replaces https://github.com/zio/zio-kafka/issues/686 and https://github.com/zio/zio-kafka/pull/960. Implements #902.


erikvanoosten:
From https://github.com/zio/zio-kafka/pull/960#issuecomment-1666531945:

> > To consider as an alternative: with the prefetch queue, could we detected stalled partition streams by checking when the queue size last changed or something?
> 
> I love this idea. The stream operator that is introduced in this PR looks quite heavy.
> 
> Let's put this PR on hold while we explore that option.

@svroonland can you review this PR please? For now it only contains the detection mechanism. Once we think this is a good idea I can copy over the other changes (like removing runloopTimeout) and the tests.

svroonland:
Nice, yeah something like this should work.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
```diff
@@ -88,16 +92,30 @@ private[consumer] object RunloopAccess {
...
                         partitionsHub = partitionsHub,
-                        runloopTimeout = settings.runloopTimeout,
                         fetchStrategy = settings.fetchStrategy
                       )
                       .withFinalizer(_ => runloopStateRef.set(RunloopState.Finalized))
                       .provide(ZLayer.succeed(consumerScope))
     } yield new RunloopAccess(runloopStateRef, partitionsHub, makeRunloop, diagnostics)
+
+  private def maxPollIntervalConfig(settings: ConsumerSettings): Task[Duration] = ZIO.attempt {
+    def defaultMaxPollInterval: Int = ConsumerConfig
```
svroonland, line 109: We could make ConsumerSettings.maxPollInterval a proper case class value and use it to fill in the consumer properties, instead of retrieving it from the consumer properties map with somewhat complicated code. Like with the `bootstrapServers` property.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -2,34 +2,65 @@ package zio.kafka.consumer.internal
...
 
-  def queueSize: UIO[Int] = queueSizeRef.get
+  def queueSize: UIO[Int] = queueInfoRef.get.map(_.size)
+
+  /**
+   * @return
+   *   `true` when the stream has data available, but none has been pulled for more than `maxPollInterval` (since data
+   *   became available), `false` otherwise
+   */
+  private[internal] def maxPollIntervalExceeded(): UIO[Boolean] =
```
svroonland, line 44: Nitpicky, but queueSize does not have parens while this method does. Personally I prefer ZIO methods without parentheses, as they don't have side effects anyway.

svroonland (commented):
Nice, this is a lot more straightforward. A few things to consider

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
erikvanoosten: It was kind of like that in #960 until @guizmaii convinced me to do it like this 🤔

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
erikvanoosten: Good catch. I'll change it.

erikvanoosten: Should I remove them from `lost`, `halted`, and `end` as well?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
svroonland: Alright, let's leave it as is then.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
svroonland: That would be consistent :)

erikvanoosten: And I keep doubting whether I should have named `halted` `halt` instead... 🤔

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
erikvanoosten: Here is a link to the comment: https://github.com/zio/zio-kafka/pull/960#discussion_r1271298633

svroonland: Valid arguments.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
erikvanoosten: Okay. Done.

svroonland (approved):

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -359,6 +359,9 @@ private[consumer] final class Runloop private (
                          pollResult.records
                        )
       updatedPendingCommits <- ZIO.filter(state.pendingCommits)(_.isPending)
+      _ <- ZIO.whenZIO(ZIO.exists(updatedStreams) { stream =>
+             ZIO.ifZIO(stream.maxPollIntervalExceeded)(stream.halted.as(true), ZIO.succeed(false))
```
guizmaii, line 363: To make things more explicit I'd prefer that we use named arguments here:
```suggestion
             ZIO.ifZIO(stream.maxPollIntervalExceeded)(onTrue = stream.halted.as(true), onFalse = ZIO.succeed(false))
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -359,6 +359,9 @@ private[consumer] final class Runloop private (
...
+           })(shutdown)
```
guizmaii, line 364: I think this code combining so many ZIO operators isn't visible enough.
IMO, it'd be better if we split it a bit a give a proper name to the boolean controlling the shutdown call (to have a clearer separation of the "what" and the "how": What are we testing here? How do we know this what?)

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
guizmaii: @erikvanoosten 
> And I keep doubting whether I should have named halted halt instead... 🤔
I prefer `halt` as this is not only returning a state but also performing an action. `halted` suggests to me that this method only returns me a state

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -359,6 +359,9 @@ private[consumer] final class Runloop private (
                          pollResult.records
                        )
       updatedPendingCommits <- ZIO.filter(state.pendingCommits)(_.isPending)
+      _ <- ZIO.whenZIO(ZIO.exists(updatedStreams) { stream =>
```
guizmaii, line 362: `updatedStreams` is `runningStreams ++ startingStreams`. I don't think we need to loop over the `startingStreams` as they're just been started so the `maxPollIntervalExceeded` is probably always `false`

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: You are right about the legibility. It is a bit too dense. I will extract it to a method and work on adding some names as you suggested.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Good catch.

guizmaii (commented):
See comments and questions :)

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
guizmaii: > I will extract it to a method

I don't think we need a method here. Just a nice name for the boolean things will already be much better :)

erikvanoosten: I agree that shutting down twice should not be an issue.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Fixed in https://github.com/zio/zio-kafka/pull/1009/commits/1cfd33bf2ee86ffd30ccda7e8077c64069db7f9e.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Fixed in https://github.com/zio/zio-kafka/pull/1009/commits/1cfd33bf2ee86ffd30ccda7e8077c64069db7f9e.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: I have used the 'more elegant way' in https://github.com/zio/zio-kafka/pull/1009/commits/1cfd33bf2ee86ffd30ccda7e8077c64069db7f9e.

Issue #694, opened 2023-03-10
[Feature Request] provide access to the underlying kafka consumer


tnielens:
At my workplace we have lots of functionality working on the java consumer for concerns like healthcheck, consumer lag monitoring etc. 
Could the zio-kafka consumer provide access to the underlying consumer? That would allow us to reuse the functionality without rewriting it against the zio-kafka Consumer methods.

guizmaii:
I need to finish this PR which allows you to have access to the underlying Java consumer. We already this approach for a while where I work

svroonland:
Now available in https://github.com/zio/zio-kafka/releases/tag/v2.4.2

Issue #452, opened 2022-04-08
Access to underlying kafka Consumer/Producer


newffy:
Sometimes it would be great to have ability to configure underlying consumer/producer directly.
Right now it can be configured via settings, but sometimes you need access to original object.
Probably it can be done in the same way as it's implemented in `alpakka kafka` project, via smth like 
`ProducerSettings.withProducerFactory` or `ProducerSettings.withProducer`

guizmaii:
@newffy Can you please give us a concrete example where you'd need this?
I'm not sure to see the need

PR #1011, opened 2023-08-06
Create Consumer and Producer from existing apache kafka Consumer/Producer

svroonland:
Implements #452 
Implements #694

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/ConsumerAccess.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/ConsumerAccess.scala
@@ -42,7 +42,6 @@ private[consumer] object ConsumerAccess {
 
   def make(settings: ConsumerSettings): ZIO[Scope, Throwable, ConsumerAccess] =
     for {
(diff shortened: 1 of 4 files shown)


erikvanoosten (approved):
Nice.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/ConsumerAccess.scala
```diff
@@ -52,7 +51,13 @@ private[consumer] object ConsumerAccess {
                       )
                     }
                   } { consumer =>
-                    ZIO.blocking(access.withPermit(ZIO.attempt(consumer.close(settings.closeTimeout)))).orDie
+                    ZIO.blocking(ZIO.attempt(consumer.close(settings.closeTimeout))).orDie
```
guizmaii, line 54: Is this change safe? Isn't there a risk that `.close` is called in parallel with other Consumer methods?

svroonland: Well I was wrong there. See #1238

PR #1017, opened 2023-08-09
Add unit test for `QueueSizeBasedFetchStrategy`

erikvanoosten:
This PR introduces abstract class `PartitionStream` to only expose the relevant fields to `FetchStrategy` (also it makes it easier to create test values).

### DIFF
--- a/zio-kafka-test/src/test/scala/zio/kafka/consumer/fetch/QueueSizeBasedFetchStrategySpec.scala
+++ b/zio-kafka-test/src/test/scala/zio/kafka/consumer/fetch/QueueSizeBasedFetchStrategySpec.scala
@@ -8,8 +8,8 @@ import zio.{ Chunk, Scope, UIO, ZIO }
 
 object QueueSizeBasedFetchStrategySpec extends ZIOSpecDefaultSlf4j {
 
-  private val maxPartitionQueueSize = 50
-  private val fetchStrategy         = QueueSizeBasedFetchStrategy(maxPartitionQueueSize)
+  private val partitionPreFetchBufferLimit = 50
+  private val fetchStrategy                = QueueSizeBasedFetchStrategy(partitionPreFetchBufferLimit)
 
   private val tp10 = new TopicPartition("topic1", 0)
   private val tp11 = new TopicPartition("topic1", 1)
@@ -17,13 +17,19 @@ object QueueSizeBasedFetchStrategySpec extends ZIOSpecDefaultSlf4j {
 
   override def spec: Spec[TestEnvironment with Scope, Any] =
     suite("QueueSizeBasedFetchStrategySpec")(
-      test("stream with queue size above maxSize is paused") {
+      test("stream with queue size above limit is paused") {
         val streams = Chunk(newStream(tp10, currentQueueSize = 100))
         for {
           result <- fetchStrategy.selectPartitionsToFetch(streams)
         } yield assertTrue(result.isEmpty)
       },
-      test("stream with queue size less-equal maxSize may resume") {
+      test("stream with queue size equal to limit is paused") {
+        val streams = Chunk(newStream(tp10, currentQueueSize = partitionPreFetchBufferLimit))
+        for {
+          result <- fetchStrategy.selectPartitionsToFetch(streams)
+        } yield assertTrue(result.isEmpty)
+      },
+      test("stream with queue size below limit may resume") {
         val streams = Chunk(newStream(tp10, currentQueueSize = 10))
         for {
           result <- fetchStrategy.selectPartitionsToFetch(streams)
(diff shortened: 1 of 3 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -6,14 +6,19 @@ import zio.kafka.consumer.internal.Runloop.ByteArrayCommittableRecord
 import zio.stream.{ Take, ZStream }
 import zio.{ Chunk, LogAnnotation, Promise, Queue, Ref, UIO, ZIO }
 
+trait PartitionStreamControlLike {
```
guizmaii, line 9: I think we should prefer abstract classes if we want to improve on binary compatibility

See https://nrinaudo.github.io/scala-best-practices/binary_compat/abstract_over_trait.html

zio-kafka-test/src/test/scala/zio/kafka/consumer/fetch/QueueSizeBasedFetchStrategySpec.scala
```diff
@@ -0,0 +1,49 @@
...
+        for {
+          result <- fetchStrategy.selectPartitionsToFetch(streams)
+        } yield assertTrue(result == Set(tp10, tp11))
+      }
+    )
+
+  private def newStream(topicPartition: TopicPartition, currentQueueSize: Int): PartitionStreamControlLike =
+    new PartitionStreamControlLike {
+      override def tp: TopicPartition  = topicPartition
+      override def queueSize: UIO[Int] = ZIO.succeed(currentQueueSize)
```
svroonland, line 47: I agree with the principle of making code easy to test, but  in this case it would not be that much work to provide dummy values for the other values of PartitionStreamControl. Also see Jules' remark on binary compatibility.

erikvanoosten: Okay, I'll try a bit harder then :)

erikvanoosten: Yeah its possible. It will make constructing the PartitionStreamControl effectful though...

erikvanoosten: There is another reason for a trait/abstract class: it defines the public interface of PartitionStreamControl.

erikvanoosten: @svroonland It would look like this: https://github.com/zio/zio-kafka/pull/1017/commits/54cf417ceba97411aff5c0e38f227fb6b6de2f5c

I am a bit annoyed that we are exposing `PartitionStreamControl` now. I would prefer that we find a better name for `PartitionStreamControlLike`, something that expresses that it is the public interface of `PartitionStreams`.
E.g. like this: https://github.com/zio/zio-kafka/pull/1017/commits/9cc5f4f2983fdb533971427718467f9de758f70f

@svroonland @guizmaii WDYT?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
erikvanoosten: Okay, I have no strong opinion on this so I'll go with abstract class.

zio-kafka-test/src/test/scala/zio/kafka/consumer/fetch/QueueSizeBasedFetchStrategySpec.scala
guizmaii: Why do we need to expose this publicly? Aren't these an implementation detail?

erikvanoosten: > Why do we need to expose this publicly? Aren't these an implementation detail?

Because we're using the constructor of `PartitionStreamControl` from a different package than where it is defined.

But more importantly, because `FetchStrategy` uses it. Anyone who want to build their own `FetchStrategy`, and want to test it, needs to construct `PartitionStreamControl` instances. The only public methods in `PartitionStreamControl` are `tp` and `queueSize`. That is why I think its better to have an interface (or abstract class) that exposes just that. I have called the abstract class `PartitionStream` in https://github.com/zio/zio-kafka/commit/9cc5f4f2983fdb533971427718467f9de758f70f.

erikvanoosten:
@svroonland @guizmaii Can you take at this PR again please?

svroonland (approved):
LGTM, let's wait for Jules' review though

erikvanoosten:
@guizmaii Sorry. I just noticed that we are still missing your approval as requested by @svroonland . Would like me to roll this back?

guizmaii:
@erikvanoosten Steve's approval is enough. Thanks for asking 🙂

PR #1018, opened 2023-08-09
add MIMA checks to ensure binary compatibility across different versions

myazinn:


build.sbt
```diff
@@ -204,3 +209,8 @@ lazy val docs = project
   )
   .enablePlugins(WebsitePlugin)
   .dependsOn(zioKafka, zioKafkaTestkit)
+
+lazy val lint = {
+  val defaultLint = zio.sbt.Commands.ComposableCommand.lint
+  defaultLint.copy(commandStrings = defaultLint.commandStrings :+ "mimaCheck").toCommand
```
myazinn, line 215: This is an awkward hack that I'm not proud of. But it's the only one that I can think of that works ¯\_(ツ)_/¯

build.sbt
```diff
@@ -123,6 +126,7 @@ lazy val zioKafkaTestkit =
     .dependsOn(zioKafka)
     .enablePlugins(BuildInfoPlugin)
     .settings(stdSettings("zio-kafka-testkit"))
+    .settings(mimaSettings(failOnProblem = false))
```
myazinn, line 129: I guess for test kit binary compatibility is not a big deal. So I've decided to enable it for this project for those who want to check it manually, but it won't cause the build to fail.

erikvanoosten:
I propose that (for now) we only keep patch versions binary compatible. So minor version may break it.
As we gain confidence (and/or find the need) we may extend this to keep minor versions binary compatible.

@myazinn is there a way to automate this? Right now I see that Mima will always compare to 2.4.2.

erikvanoosten:
🚀  @myazinn Thanks for your contribution!

myazinn:
Hi @erikvanoosten :) Apologies for not replying for so long.

> @myazinn is there a way to automate this? Right now I see that Mima will always compare to 2.4.2.

Yes, but only up to a point. We can always use the latest version to compare like this
```scala
lazy val bincompatVersionToCompare = "latest.integration"
```
Or choose the latest patch like this (maybe there's an easier way though).
```scala
  def mimaSettings(failOnProblem: Boolean) =
    Seq(
      mimaPreviousArtifacts := {
        version.value match {
          case VersionNumber(Seq(major, minor, patch), _, _) =>
            Set(organization.value %% name.value % s"$major.$minor.$patch")
          case _ =>
            Set.empty
        }
      },
      mimaBinaryIssueFilters ++= Seq(
        exclude[Problem]("zio.kafka.consumer.internal.*")
      ),
      mimaFailOnProblem := failOnProblem
    )

```
Or using [ivy wildcards](https://ant.apache.org/ivy/history/2.3.0/ivyfile/dependency.html).

But
>  propose that (for now) we only keep patch versions binary compatible.

This is a tricky part. When you make a MR, you don't actually know which next version will be released, so the checker can't decide whether to fail the build or not. It should be possible to do something like this
```scala
  def mimaSettings(failOnProblem: Boolean) =
    Seq(
      mimaPreviousArtifacts := {
        version.value match {
          // patch == 0 means there's a new minor or major version, which doesn't have to maintain binary compatibility
          case VersionNumber(Seq(_, _, 0), _, _) =>
            Set.empty
          case VersionNumber(Seq(major, minor, patch), _, _) =>
            Set(organization.value %% name.value % s"$major.$minor.+")
          case _ =>
            Set.empty
        }
      },
      mimaBinaryIssueFilters ++= Seq(
        exclude[Problem]("zio.kafka.consumer.internal.*")
      ),
      mimaFailOnProblem := failOnProblem
    )
```
and run this when the new version was cut and is being build. It's a safer option, but far less convenient, since the release might be blocked because of some compatibility issues caused by MR merged weeks ago.

erikvanoosten:
Replaced by #1214.

erikvanoosten:
Thanks @myazinn !

PR #1021, opened 2023-08-12
[Internal] Micro-optimisation: Only compute the `requestedPartitions` Set if necessary

guizmaii:


erikvanoosten:
Nice find! But IMHO we should go further: `requestedPartitions` should be removed completely. Instead `partitionsToFetch` can be used.

erikvanoosten (approved):

Issue #852, opened 2023-05-18
Deadlock with parallel processing of single partition stream


vladimirkl:
Hi, I need to take n records from single partition, process in parallel batches, and commit. Unfortunately this code causes a deadlock in Consumer on commit:

```scala
 val consumerLayer: ZLayer[Kafka, Throwable, Consumer] =
     ZLayer.scoped ( ZIO.serviceWithZIO[Kafka] { kafka =>
       Consumer.make(ConsumerSettings(kafka.bootstrapServers).withGroupId("group1").withOffsetRetrieval(Consumer.OffsetRetrieval.Auto(AutoOffsetStrategy.Earliest)))
     })

...

Producer.produceChunk(Chunk.fromIterable(1 to 1000).map(n => new ProducerRecord(topic, n, n.toString)), Serde.int, Serde.string) *>
  Consumer.plainStream(Subscription.topics(topic), Serde.int, Serde.string)
    .take(100)
    .groupedWithin(10, 100.millis)
    .mapZIOPar(2)(c => ZIO.debug(c.size) as c.map(_.offset))
    .map(OffsetBatch.apply)
    .debug("Offset")
    .mapZIO(_.commit)
    .debug("Commit")
    .runDrain zipPar Fiber.dumpAll.delay(20.seconds)

```
Replacing `groupedWithin` with `grouped` makes this code working. Also I can leave `groupedWithin` in place and remove `take` -  no deadlock occurs but I need to terminate stream exactly for n records.  It looks like race condition on stream termination with `take`.  Tested with zio-kafka 2.3.0, Scala 2.13.10  and embedded Kafka

guizmaii:
Thanks for your report @vladimirkl 🙂

I'm wondering if this is a bug related to zio-kafka or zio-streams 🤔
Because, for example, changing `groupedWithin` with `grouped` as you did, has nothing to do with zio-kafka.
Why do you think it comes from zio-kafka?

Sorry for the questions. I'm trying to better understand your issue.

vladimirkl:
It may be some zio-streams issue, but I never encountered it without zio-kafka - while I have similar aggregations in other places . It locks exactly on `commit`. If you wait some time after deadlock you will see RunloopTimeout exception. Without `commit` take terminates stream correctly

vladimirkl:
I even tried to replace `commit` with something non interruptible  like `ZIO.attemptBlocking(Thread.sleep(2000))` and it works correctly. So it's very likely  Consumer issue

[...]

PR #1022, opened 2023-08-12
Add `Consumer.commit` and `Consumer.commitOrRetry` methods

guizmaii:
First step toward removing the `Offset` trait atrocity and the current OOP orientation of the commit interface proposed by zio-kafka (ie. `record.offset.commit`. This is non sense which pisses me off TBH)

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
@@ -5,9 +5,9 @@ import zio.kafka.consumer.diagnostics.DiagnosticEvent.Finalization
 import zio.kafka.consumer.diagnostics.Diagnostics
 import zio.kafka.consumer.internal.Runloop.ByteArrayCommittableRecord
 import zio.kafka.consumer.internal.RunloopAccess.PartitionAssignment
-import zio.kafka.consumer.{ ConsumerSettings, InvalidSubscriptionUnion, Subscription }
+import zio.kafka.consumer.{ CommittableRecord, ConsumerSettings, InvalidSubscriptionUnion, Subscription }
 import zio.stream.{ Stream, Take, UStream, ZStream }
-import zio.{ Hub, IO, Ref, Scope, UIO, ZIO, ZLayer }
+import zio._
 
 private[internal] sealed trait RunloopState
 private[internal] object RunloopState {
@@ -31,10 +31,10 @@ private[consumer] final class RunloopAccess private (
 ) {
 
   private def withRunloopZIO[E](
-    requireRunning: Boolean
+    shouldStartIfNot: Boolean
   )(whenRunning: Runloop => IO[E, Unit]): IO[E, Unit] =
     runloopStateRef.updateSomeAndGetZIO {
-      case RunloopState.NotStarted if requireRunning => makeRunloop.map(RunloopState.Started.apply)
+      case RunloopState.NotStarted if shouldStartIfNot => makeRunloop.map(RunloopState.Started.apply)
     }.flatMap {
       case RunloopState.NotStarted       => ZIO.unit
       case RunloopState.Started(runloop) => whenRunning(runloop)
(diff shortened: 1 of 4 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
```diff
@@ -31,10 +31,10 @@ private[consumer] final class RunloopAccess private (
 ) {
 
   private def withRunloopZIO[E](
-    requireRunning: Boolean
+    shouldStartIfNot: Boolean
```
guizmaii, line 34: I'm sorry @erikvanoosten but I really disliked the name you used. I prefer the one I was using. IMO, it's more explicit about what this boolean is controlling

erikvanoosten: 😄  My major gripes with `shouldStartIfNot` is that it is like an unfinished sentence. If not, .. if not what?
That is how I went to `shouldStartIfNotRunning` and from there to `requireRunning`.

guizmaii: Well, it's "should start if not started" but there's IMO no need to repeat the "start" word with "started"

erikvanoosten:
If I understand the code right, after this change one needs a `CommittableRecord` to do a commit. Unfortunately, for high throughput services it is not feasible to keep all `CommittableRecord`s in memory until the offset needs to be committed.

I think we should keep the notion of `Offset` and `Offsets`. Then we expect the user to pass an `Offset` or `Offsets` into the `commit` method. This is also more in line with the Kafka model (one commits offsets, not a record).

Or did I miss something?

svroonland:
> I think we should keep the notion of Offset and Offsets.
👍

guizmaii:
@erikvanoosten 

> I think we should keep the notion of Offset and Offsets. Then we expect the user to pass an Offsets into the commit method. This is also more in line with the Kafka model (one commits offsets, not a record).
>
>Or did I miss something?

We agree on the need not to cache the records and to only cache the minimum information needed to avoid explode memory usage 🙂

This PR is just a first step towards the goal I want to achieve

These 2 functions I add here are only here to commit **one record at a time** - if it's what the user needs - replacing, by terms, the `Offet::commit` method. 

Replacing the `OffsetBatch::commit` method requires more work that I wanna do later
My idea, for now, will be kind of similar tho. 
The user will have something like this to commit some records:
```scala
 trait Consumer {

  // I was discussing this idea in Discord some time ago
  // see: https://discord.com/channels/629491597070827530/629497941719121960/1144576343330263080

  /**
   * The `UIO` represents the action of putting the Record(s) in the "commit batch"
   * The `Task` represents the action of commiting the batch to Kaka
   */
  def commitAccumBatch(commitSchedule: Schedule)(record: Record): UIO[Task[Unit]]
  def commitAccumBatch(commitschedule: Schedule)(records: Chunk[Record]): UIO[Task[Unit]]

}
```
so the user will replace the aggregation he/she does on his side by calling one of these functions which will do the aggregation and the commits on our side.

Doing the aggregation on our side means that we'll have access to the "list of pending commits" inside zio-kafka, which will help us fix the deadlock some people are experiencing (ie. https://github.com/zio/zio-kafka/issues/852)

Do you see what I wanna do? What do you think about it?

Maybe it's better to implement everything before merging these new functions so that we can bring the complete new interface and deprecate the old one all at once, and so we can ensure the new interface is sound and coherent.

erikvanoosten:
> Do you see what I wanna do? What do you think about it?

Do I understand that correctly that you want the users to add offsets to a commit batch, and then when they think it is time, they can complete the batch, effectively committing it for real?

Yeah, that might work. In fact, I like that idea. If the user indicates that an offset needs to be committed at some point, we know we have to wait for it with closing the consumer (or the partition).

However, I'd prefer we do not smash up committing and this new concept of intending to commit. Can we keep separate API methods for that?

svroonland:
How shall we move forward with the (currently) 4 open PRs related to the commit interface?

Can we summarize the interface changes we would like to see and break it down into small potential PRs?

PR #1028, opened 2023-08-28
Added a failing test for stopConsumption with aggregateAsync

vladimirkl:
Relates to https://github.com/zio/zio-kafka/issues/852


erikvanoosten:
Added this test, but fixed, in #1123.

PR #1057, opened 2023-09-27
Put back the `ProducerSettings::driverSettings` and `AdminClientSettings::driverSettings` methods to avoid a useless breaking change and to provide a more unified interface through all the settings case classes

guizmaii:
Related to https://github.com/zio/zio-kafka/pull/1038


erikvanoosten (approved):
Good idea!

PR #1063, opened 2023-10-01
Replace pending requests by single boolean

erikvanoosten:
Since streams now wait for data by listening to their own queue, we no longer need to keep track of pending requests inside runloop. Instead:
- streams initiate a poll when they need data,
- runloop initiates another poll when not all streams got data.

Runloop still needs to know which streams need data. For this `def awaitingData` was added to the stream.

The advantage of this approach is:
* Less likely to go wrong (calculating pendingRequests is not trivial).
* The 'awaiting data state' is less implicit because it moved from runloop to stream.
* Slightly less pressure on the garbage collector because results from partition calculations are kept for a very short time (they are no longer kept until the next run loop).

Also:
* Removed field `tpRequested` from `DiagnosticEvent.Poll`.
* Class/object `State` is moved into `object Runloop` to hide it for the rest of the package.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -451,7 +451,7 @@ private[consumer] final class Runloop private (
    *   - Poll periodically when we are subscribed but do not have assigned streams yet. This happens after
    *     initialization and rebalancing
    */
-  def run(initialState: State): ZIO[Scope, Throwable, Any] = {
+  private def run(initialState: State): ZIO[Scope, Throwable, Any] = {
     import Runloop.StreamOps
 
     ZStream
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -238,15 +238,16 @@ private[consumer] final class Runloop private (
 
   private def handlePoll(state: State): Task[State] =
     for {
-      _                 <- ZIO.logDebug(s"Starting poll with ${state.pendingCommits.size} pending commits")
-      _                 <- currentStateRef.set(state)
-      partitionsToFetch <- fetchStrategy.selectPartitionsToFetch(state.assignedStreams)
+      _                      <- ZIO.logDebug(s"Starting poll with ${state.pendingCommits.size} pending commits")
+      _                      <- currentStateRef.set(state)
+      partitionsAwaitingData <- ZIO.filter(state.assignedStreams)(_.awaitingData).map(_.map(_.tp))
+      partitionsToFetch      <- fetchStrategy.selectPartitionsToFetch(state.assignedStreams)
       pollResult <-
         consumer.runloopAccess { c =>
           ZIO.suspend {
-            val prevAssigned = c.assignment().asScala.toSet
-
-            resumeAndPausePartitions(c, prevAssigned, partitionsToFetch)
+            val prevAssigned       = c.assignment().asScala.toSet
+            val partitionsToResume = partitionsToFetch ++ partitionsAwaitingData
+            resumeAndPausePartitions(c, prevAssigned, partitionsToResume)
 
             val polledRecords = {
               val records = c.poll(pollTimeout)
@@ -255,8 +256,8 @@ private[consumer] final class Runloop private (
 
             val currentAssigned          = c.assignment().asScala.toSet
             val newlyAssigned            = currentAssigned -- prevAssigned
-            val currentPartitionsToFetch = partitionsToFetch intersect currentAssigned
-            val morePollsNeeded          = currentPartitionsToFetch.size > polledRecords.partitions().size()
+            val currentResumedPartitions = partitionsToResume intersect currentAssigned
+            val morePollsNeeded          = currentResumedPartitions.size > polledRecords.partitions().size()
 
             for {
               ignoreRecordsForTps <- doSeekForNewPartitions(c, newlyAssigned)
@@ -305,7 +306,7 @@ private[consumer] final class Runloop private (
                      val providedTps = polledRecords.partitions().asScala.toSet
                      DiagnosticEvent.Poll(
                        tpWithData = providedTps,
-                       tpWithoutData = currentPartitionsToFetch -- providedTps
+                       tpWithoutData = currentResumedPartitions -- providedTps
                      )
                    }
 
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -10,6 +10,7 @@ import zio.kafka.consumer.diagnostics.DiagnosticEvent.Finalization
 import zio.kafka.consumer.diagnostics.{ DiagnosticEvent, Diagnostics }
 import zio.kafka.consumer.fetch.FetchStrategy
 import zio.kafka.consumer.internal.ConsumerAccess.ByteArrayKafkaConsumer
+import zio.kafka.consumer.internal.Runloop._
 import zio.kafka.consumer.internal.RunloopAccess.PartitionAssignment
 import zio.stream._
 
@@ -86,26 +87,20 @@ private[consumer] final class Runloop private (
               ZIO.fail(new IllegalStateException(s"Multiple onAssigned calls on rebalance listener"))
           },
       onRevoked = (_, _) =>
-        ZIO.logDebug("Rebalancing started") *>
-          currentStateRef.get.flatMap { state =>
-            // End all streams
-            endRevokedPartitions(
-              state.pendingRequests,
-              state.assignedStreams,
-              isRevoked = _ => true
-            ).flatMap { result =>
-              lastRebalanceEvent.updateZIO {
-                case None =>
-                  ZIO.some(Runloop.RebalanceEvent.Revoked(result))
-                case _ =>
-                  ZIO.fail(
-                    new IllegalStateException(
-                      s"onRevoked called on rebalance listener with pending assigned event"
-                    )
-                  )
-              }
-            }
-          }
+        for {
+          _     <- ZIO.logDebug("Rebalancing started")
+          state <- currentStateRef.get
+          // End all streams
+          result <- endRevokedPartitions(state.assignedStreams, isRevoked = _ => true)
+          _ <- lastRebalanceEvent.updateZIO {
+                 case None =>
+                   ZIO.some(Runloop.RebalanceEvent.Revoked(result))
+                 case _ =>
+                   ZIO.fail(
+                     new IllegalStateException(s"onRevoked called on rebalance listener with pending assigned event")
+                   )
+               }
+        } yield ()
     )
 
     if (restartStreamsOnRebalancing) {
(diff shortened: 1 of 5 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -273,8 +253,10 @@ private[consumer] final class Runloop private (
               if (records eq null) ConsumerRecords.empty[Array[Byte], Array[Byte]]() else records
             }
 
-            val currentAssigned = c.assignment().asScala.toSet
-            val newlyAssigned   = currentAssigned -- prevAssigned
+            val currentAssigned          = c.assignment().asScala.toSet
+            val newlyAssigned            = currentAssigned -- prevAssigned
+            val currentPartitionsToFetch = partitionsToFetch intersect currentAssigned
+            val morePollsNeeded          = currentPartitionsToFetch.size > polledRecords.partitions().size()
```
erikvanoosten, line 259: I wondering if we should make it:

```scala
            val morePollsNeeded =
              (currentPartitionsToFetch.size > polledRecords.partitions().size()) || newlyAssigned.nonEmpty
```
to immediately poll again when there are new partitions.

I thinnk it is not necessary because the new streams will very quickly send a poll command. WDYT?

svroonland:
I don't have a concrete example, but I have a feeling that there's a race condition in here somewhere, with `morePollsNeeded` being toggled by Poll commands and `handlePoll`. We need to be sure somehow that if a partition stream makes a request, that that stream's request is fulfilled, instead of another partition fulfillment leading to `morePollsNeeded` being reset. 

Perhaps a `Set[TopicPartition]` is a good middleground between boolean and a `Chunk[Request]`, so we can keep track of individual TPs that are requesting data..?

erikvanoosten:
> I don't have a concrete example, but I have a feeling that there's a race condition in here somewhere, with `morePollsNeeded` being toggled by Poll commands and `handlePoll`.

I am sympathetic to such feelings; they are powerful indicators. Even so, I think we are solid. There are no race conditions because the runloop runs sequentially. Streams run concurrently with the runloop so we could have multiple Poll command arriving in one loop run, but this is okay. When one or more Poll commands arrives in the run loop, they _will_ be handled. Either in this loop run, or in the next, or the next...

When a Poll command is being sent, we need to make sure that `handlePoll` is invoked inside the Runloop's `run` method (line 467 in this PR). This is why the Poll command sets the `morePollsNeeded` flag to true.

Then once `handlePoll` runs, we use the `fetchStrategy` to select the partitions that need data. (This is slightly tricky  because this PR assumes that the fetch strategy will always select partitions that have no data. We may need to make this requirement more clear in the FetchStrategy scaladocs, or we can wrap the fetch strategy to make sure this happens.)

Finally, based on which partitions want data, and which actually got data we set the `morePollsNeeded` flag again (determined in line 259). This time it is used to determine whether we should send another Poll command or not (line 470 in this PR). So in case no data was received for some partition, the loop immediately start again to give it another chance.

erikvanoosten:
Okay, so there is a race condition 😄 

```scala
                 ZStream.repeatZIOChunk {
                   // First try to take all records that are available right now.
                   dataQueue.takeAll.flatMap(data => if (data.isEmpty) requestAndAwaitData else ZIO.succeed(data))  // <--- Poll command send
                 }.flattenTake
                   .chunksWith(_.tap(records => queueSize.update(_ - records.size)))       //  <--- queueSize updated
                   .interruptWhen(interruptionPromise)
```
 This PR only works correctly when the queueSize is set to 0 _strictly before_ the Poll command is send. That is currently not the case.

erikvanoosten:
> so we can keep track of individual TPs that are requesting data..?

@svroonland Forgot to write this earlier: we already track this. The streams that have a queueSize of `0` are also the streams that requested data. (With the caveat that we have a race condition, see previous comment.)

svroonland:
> Forgot to write this earlier: we already track this. The streams that have a queueSize of 0 are also the streams that requested data. (With the caveat that we have a race condition, see previous comment.)

Yes, but that condition should somehow be a guarantee that there is also at least one Poll command on the queue, otherwise it would not lead to any code checking that condition (in handlePoll).

erikvanoosten:
> Yes, but that condition should somehow be a guarantee that there is also at least one Poll command on the queue, otherwise it would not lead to any code checking that condition (in handlePoll).

Indeed. That part is covered by what I wrote in https://github.com/zio/zio-kafka/pull/1063#issuecomment-1742195179

erikvanoosten:
I removed the race condition. Also, we no longer depend on the FetchStrategy to do the right thing.

erikvanoosten:
So in the end this PR makes state tracked by Runloop smaller, in turn it extends the state tracked by PartitionStreamControl. Some compute extensive operation were removed, but some others were introduced.

So this PR should have a positive effect on the garbage collector (heap allocations are used for a shorter time) but overall complexity is about the same.

yarosman:
It's strange, but there were diffs from already merged pr 1065

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -168,44 +161,34 @@ private[consumer] final class Runloop private (
...
     // topic-partition, is by first getting the set of topic-partitions, and
     // then requesting the records per topic-partition.
-    val tps           = polledRecords.partitions().asScala.toSet -- ignoreRecordsForTps
-    val fulfillResult = Runloop.FulfillResult(pendingRequests = pendingRequests.filter(req => !tps.contains(req.tp)))
+    val tps = polledRecords.partitions().asScala.toSet -- ignoreRecordsForTps
     val streams =
       if (tps.isEmpty) Chunk.empty else partitionStreams.filter(streamControl => tps.contains(streamControl.tp))
 
-    if (streams.isEmpty) ZIO.succeed(fulfillResult)
+    if (streams.isEmpty) ZIO.unit
```
svroonland, line 184: Could be replaced by `ZIO.unless(streams.isEmpty) { for { consumerGroupMetaData ... } }`

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -255,26 +238,26 @@ private[consumer] final class Runloop private (
 
   private def handlePoll(state: State): Task[State] =
     for {
-      _ <-
-        ZIO.logDebug(
-          s"Starting poll with ${state.pendingRequests.size} pending requests and ${state.pendingCommits.size} pending commits"
-        )
-      _                 <- currentStateRef.set(state)
-      partitionsToFetch <- fetchStrategy.selectPartitionsToFetch(state.assignedStreams)
+      _                      <- ZIO.logDebug(s"Starting poll with ${state.pendingCommits.size} pending commits")
```
svroonland, line 241: Could we keep the pending requests log message and calculate the nr of pending requests from `partitionsAwaitingData` (below)?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -255,26 +238,26 @@ private[consumer] final class Runloop private (
...
+      _                      <- currentStateRef.set(state)
+      partitionsAwaitingData <- ZIO.filter(state.assignedStreams)(_.awaitingData).map(_.map(_.tp))
+      partitionsToFetch      <- fetchStrategy.selectPartitionsToFetch(state.assignedStreams)
```
svroonland, line 244: Should `partitiosnAwaitingData` be an input to the fetch strategy, so it can decide whether to include or disregard them?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -255,26 +238,26 @@ private[consumer] final class Runloop private (
...
               val records = c.poll(pollTimeout)
               if (records eq null) ConsumerRecords.empty[Array[Byte], Array[Byte]]() else records
             }
 
-            val currentAssigned = c.assignment().asScala.toSet
-            val newlyAssigned   = currentAssigned -- prevAssigned
+            val currentAssigned          = c.assignment().asScala.toSet
+            val newlyAssigned            = currentAssigned -- prevAssigned
+            val currentResumedPartitions = partitionsToResume intersect currentAssigned
+            val morePollsNeeded          = currentResumedPartitions.size > polledRecords.partitions().size()
```
svroonland, line 260: Should this be a set difference instead of size comparison?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -606,26 +571,25 @@ private[consumer] object Runloop {
...
+    subscriptionState: SubscriptionState,
+    morePollsNeeded: Boolean
+  ) {
+    def addCommit(c: RunloopCommand.Commit): State = copy(pendingCommits = pendingCommits :+ c)
 
-  def shouldPoll: Boolean =
-    subscriptionState.isSubscribed && (pendingRequests.nonEmpty || pendingCommits.nonEmpty || assignedStreams.isEmpty)
-}
+    def shouldPoll: Boolean =
+      subscriptionState.isSubscribed && (morePollsNeeded || pendingCommits.nonEmpty || assignedStreams.isEmpty)
```
svroonland, line 584: Perhaps for 'knowledge/terminology compatibility' we can keep the name `pendingRequests` but make it a boolean.

svroonland (commented):
Feeling a lot more reliable this way. Some thoughts

erikvanoosten:
> It's strange, but there were diffs from already merged pr 1065

I extracted the uncontroversial parts from this PR into #1065. I didn't rebase it yet. Depending on @svroonland further comments I may not even bother and just close this PR.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Yes that is possible. However, ZIO.when and ZIO.unless need one more IO and are therefore slightly less performant.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Because we want to do pre-fetching, we already give _all_ partitions to the fetch strategy, including the partitions in `partitionsAwaitingData`.

As to whether we should allow the fetch strategy to disregard a partition that is already requesting new data, that is not something I though about yet. Do we want to do this? It may be beneficial to power users (those that use custom prefetch strategies in the first place) so I am leaning towards allowing it.

There is a problem though. We would need to extend the stream trait. Currently fetch strategy decides only based on queue-size. Oddly enough, this may be positive, even while the stream is starving and requested new data. So we need to also make a new stream method to indicates 'I-need-data'.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Because one is a strict super set of the other a simple size difference is okay.

yarosman: can we use `sizeIs > polledRecords` ?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -532,18 +501,14 @@ private[consumer] object Runloop {
 
   private final case class PollResult(
     startingTps: Set[TopicPartition],
-    pendingRequests: Chunk[RunloopCommand.Request],
     assignedStreams: Chunk[PartitionStreamControl],
     records: ConsumerRecords[Array[Byte], Array[Byte]],
-    ignoreRecordsForTps: Set[TopicPartition]
+    ignoreRecordsForTps: Set[TopicPartition],
+    morePollsNeeded: Boolean
```
yarosman, line 507: nit: `isMorePollsNeeded`

guizmaii:
@erikvanoosten Can you rebase your PR, please?

erikvanoosten:
Let's try another way to tackle this.

PR #1067, opened 2023-10-04
Enable the consumer to commit an offset with metadata

### LINKED ISSUE #1066: Consumer should be able to commit offset with metadata
It could be nice to be able to commit an offset with a metadata.

We could add this method in the **Offset** trait:

`def withMetadata(metadata: String): Self`

This way we don't create a breaking change in the Consumer API.

flavienbert:
Fixes #1066

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,9 +13,7 @@ env:
     branches:
     - master
     - series/0.x
-  pull_request:
-    branches-ignore:
-    - gh-pages
+  pull_request: {}
 jobs:
   build:
     name: Build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,9 @@ env:
     branches:
     - master
     - series/0.x
-  pull_request: {}
+  pull_request:
+    branches-ignore:
+    - gh-pages
 jobs:
   build:
     name: Build
--- a/zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
+++ b/zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
@@ -349,6 +349,29 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
                        .provideSomeLayer[Kafka](consumer(client, Some(group)))
         } yield assert(offsets.values.map(_.map(_.offset)))(forall(isSome(equalTo(nrMessages.toLong / nrPartitions))))
       },
+      test("commits an offset with metadata") {
+        for {
+          topic    <- randomTopic
+          group    <- randomGroup
+          metadata <- randomThing("metadata")
+          client   <- randomClient
+          _        <- ZIO.attempt(EmbeddedKafka.createCustomTopic(topic, partitions = 1))
+          _        <- produceOne(topic, "key", "msg")
+
+          // Consume messages
+          subscription = Subscription.topics(topic)
+          offsets <- (Consumer
+                       .partitionedStream(subscription, Serde.string, Serde.string)
+                       .flatMap(_._2.map(_.offset.withMetadata(metadata)))
+                       .take(1)
+                       .transduce(Consumer.offsetBatches)
+                       .take(1)
+                       .mapZIO(_.commit)
+                       .runDrain *>
+                       Consumer.committed(Set(new TopicPartition(topic, 0))))
+                       .provideSomeLayer[Kafka](consumer(client, Some(group)))
+        } yield assert(offsets.values.headOption.flatten.map(_.metadata))(isSome(equalTo(metadata)))
+      },
       test("handle rebalancing by completing topic-partition streams") {
         val nrMessages   = 50
         val nrPartitions = 6 // Must be even and strictly positive
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/Offset.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/Offset.scala
@@ -1,16 +1,21 @@
 package zio.kafka.consumer
 
-import org.apache.kafka.clients.consumer.{ ConsumerGroupMetadata, RetriableCommitFailedException }
+import org.apache.kafka.clients.consumer.{ ConsumerGroupMetadata, OffsetAndMetadata, RetriableCommitFailedException }
 import org.apache.kafka.common.TopicPartition
 import zio.{ RIO, Schedule, Task }
 
 sealed trait Offset {
+
   def topic: String
   def partition: Int
   def offset: Long
   def commit: Task[Unit]
   def batch: OffsetBatch
   def consumerGroupMetadata: Option[ConsumerGroupMetadata]
+  def withMetadata(metadata: String): Offset
+
+  private[consumer] def metadata: Option[String]
+  private[consumer] def asJavaOffsetAndMetadata: OffsetAndMetadata = new OffsetAndMetadata(offset, metadata.orNull)
 
   /**
    * Attempts to commit and retries according to the given policy when the commit fails with a
@@ -39,9 +44,15 @@ private final case class OffsetImpl(
   topic: String,
   partition: Int,
   offset: Long,
-  commitHandle: Map[TopicPartition, Long] => Task[Unit],
-  consumerGroupMetadata: Option[ConsumerGroupMetadata]
+  commitHandle: Map[TopicPartition, OffsetAndMetadata] => Task[Unit],
+  consumerGroupMetadata: Option[ConsumerGroupMetadata],
+  metadata: Option[String] = None
 ) extends Offset {
-  def commit: Task[Unit] = commitHandle(Map(topicPartition -> offset))
-  def batch: OffsetBatch = OffsetBatchImpl(Map(topicPartition -> offset), commitHandle, consumerGroupMetadata)
+  def commit: Task[Unit] = commitHandle(Map(topicPartition -> asJavaOffsetAndMetadata))
+  def batch: OffsetBatch = OffsetBatchImpl(
+    Map(topicPartition -> asJavaOffsetAndMetadata),
+    commitHandle,
+    consumerGroupMetadata
+  )
+  def withMetadata(metadata: String): OffsetImpl = copy(metadata = Some(metadata))
 }
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/OffsetBatch.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/OffsetBatch.scala
@@ -53,7 +53,7 @@ private final case class OffsetBatchImpl(
     otherOffsets.offsets.foreach { case (tp, offset) =>
       val laterOffset = offsets.get(tp) match {
         case Some(existing) => if (existing.offset < offset.offset) offset else existing
-        case None => offset
+        case None           => offset
       }
       newOffsets += tp -> laterOffset
     }
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/OffsetBatch.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/OffsetBatch.scala
@@ -1,11 +1,11 @@
 package zio.kafka.consumer
 
-import org.apache.kafka.clients.consumer.ConsumerGroupMetadata
+import org.apache.kafka.clients.consumer.{ ConsumerGroupMetadata, OffsetAndMetadata }
 import org.apache.kafka.common.TopicPartition
 import zio.{ RIO, Schedule, Task, ZIO }
 
 sealed trait OffsetBatch {
-  def offsets: Map[TopicPartition, Long]
+  def offsets: Map[TopicPartition, OffsetAndMetadata]
   def commit: Task[Unit]
   def add(offset: Offset): OffsetBatch
   @deprecated("Use add(Offset) instead", "2.1.4")
@@ -28,26 +28,31 @@ object OffsetBatch {
 }
 
 private final case class OffsetBatchImpl(
-  offsets: Map[TopicPartition, Long],
-  commitHandle: Map[TopicPartition, Long] => Task[Unit],
+  offsets: Map[TopicPartition, OffsetAndMetadata],
+  commitHandle: Map[TopicPartition, OffsetAndMetadata] => Task[Unit],
   consumerGroupMetadata: Option[ConsumerGroupMetadata]
 ) extends OffsetBatch {
   override def commit: Task[Unit] = commitHandle(offsets)
 
-  override def add(offset: Offset): OffsetBatch =
+  override def add(offset: Offset): OffsetBatch = {
+    val maxOffsetAndMetadata = offsets.get(offset.topicPartition) match {
+      case Some(existing) if existing.offset > offset.offset => existing
+      case _                                                 => offset.asJavaOffsetAndMetadata
+    }
+
     copy(
-      offsets = offsets + (offset.topicPartition -> (offsets
-        .getOrElse(offset.topicPartition, -1L) max offset.offset))
+      offsets = offsets + (offset.topicPartition -> maxOffsetAndMetadata)
     )
+  }
 
   override def merge(offset: Offset): OffsetBatch = add(offset)
 
   override def merge(otherOffsets: OffsetBatch): OffsetBatch = {
-    val newOffsets = Map.newBuilder[TopicPartition, Long]
+    val newOffsets = Map.newBuilder[TopicPartition, OffsetAndMetadata]
     newOffsets ++= offsets
     otherOffsets.offsets.foreach { case (tp, offset) =>
-      val existing = offsets.getOrElse(tp, -1L)
-      if (existing < offset)
+      val existing = offsets.getOrElse(tp, new OffsetAndMetadata(-1L))
+      if (existing.offset < offset.offset)
         newOffsets += tp -> offset
     }
 
@@ -56,7 +61,7 @@ private final case class OffsetBatchImpl(
 }
 
 case object EmptyOffsetBatch extends OffsetBatch {
-  override val offsets: Map[TopicPartition, Long]                   = Map.empty
+  override val offsets: Map[TopicPartition, OffsetAndMetadata]      = Map.empty
   override val commit: Task[Unit]                                   = ZIO.unit
   override def add(offset: Offset): OffsetBatch                     = offset.batch
   override def merge(offset: Offset): OffsetBatch                   = add(offset)
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/OffsetBatch.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/OffsetBatch.scala
@@ -51,9 +51,11 @@ private final case class OffsetBatchImpl(
     val newOffsets = Map.newBuilder[TopicPartition, OffsetAndMetadata]
     newOffsets ++= offsets
     otherOffsets.offsets.foreach { case (tp, offset) =>
-      val existing = offsets.getOrElse(tp, new OffsetAndMetadata(-1L))
-      if (existing.offset < offset.offset)
-        newOffsets += tp -> offset
+      val laterOffset = offsets.get(tp) match {
+        case Some(existing) => if (existing.offset < offset.offset) offset else existing
+        case None => offset
+      }
+      newOffsets += tp -> laterOffset
     }
 
     copy(offsets = newOffsets.result())
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -113,7 +113,7 @@ private[consumer] final class Runloop private (
   }
 
   /** This is the implementation behind the user facing api `Offset.commit`. */
-  private val commit: Map[TopicPartition, Long] => Task[Unit] =
+  private val commit: Map[TopicPartition, OffsetAndMetadata] => Task[Unit] =
     offsets =>
       for {
         p <- Promise.make[Throwable, Unit]
@@ -127,16 +127,21 @@ private[consumer] final class Runloop private (
     commits: Chunk[RunloopCommand.Commit]
   ): (JavaMap[TopicPartition, OffsetAndMetadata], OffsetCommitCallback, Throwable => UIO[Unit]) = {
     val offsets = commits
-      .foldLeft(mutable.Map.empty[TopicPartition, Long]) { case (acc, commit) =>
+      .foldLeft(mutable.Map.empty[TopicPartition, OffsetAndMetadata]) { case (acc, commit) =>
         commit.offsets.foreach { case (tp, offset) =>
-          acc += (tp -> acc.get(tp).map(_ max offset).getOrElse(offset))
+          acc += (tp -> acc
+            .get(tp)
+            .map(current => if (current.offset() > offset.offset()) current else offset)
+            .getOrElse(offset))
         }
         acc
       }
       .toMap
-    val offsetsWithMetaData = offsets.map { case (tp, offset) => tp -> new OffsetAndMetadata(offset + 1) }
-    val cont                = (e: Exit[Throwable, Unit]) => ZIO.foreachDiscard(commits)(_.cont.done(e))
-    val onSuccess           = cont(Exit.unit) <* diagnostics.emit(DiagnosticEvent.Commit.Success(offsetsWithMetaData))
+    val offsetsWithMetaData = offsets.map { case (tp, offset) =>
+      tp -> new OffsetAndMetadata(offset.offset + 1, offset.leaderEpoch, offset.metadata)
+    }
+    val cont      = (e: Exit[Throwable, Unit]) => ZIO.foreachDiscard(commits)(_.cont.done(e))
+    val onSuccess = cont(Exit.unit) <* diagnostics.emit(DiagnosticEvent.Commit.Success(offsetsWithMetaData))
     val onFailure: Throwable => UIO[Unit] = {
       case _: RebalanceInProgressException =>
         for {
--- a/zio-kafka/src/main/scala/zio/kafka/producer/TransactionalProducer.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/producer/TransactionalProducer.scala
@@ -41,7 +41,7 @@ object TransactionalProducer {
             case Some(consumerGroupMetadata) =>
               val offsets: util.Map[TopicPartition, OffsetAndMetadata] =
                 offsetBatch.offsets.map { case (topicPartition, offset) =>
-                  topicPartition -> new OffsetAndMetadata(offset + 1)
+                  topicPartition -> new OffsetAndMetadata(offset.offset + 1, offset.metadata)
                 }.asJava
 
               ZIO.attemptBlocking(live.p.sendOffsetsToTransaction(offsets, consumerGroupMetadata))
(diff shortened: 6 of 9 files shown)


erikvanoosten:
@flavienbert I have in my plans to merge commits into a single commit. Is that still possible when there is metadata attached to the commit?

flavienbert:
@erikvanoosten yes I think. The apache kafka consumer client take an `OffsetAndMetadata` so it should be OK.

erikvanoosten:
> @erikvanoosten yes I think. The apache kafka consumer client take an `OffsetAndMetadata` so it should be OK.

Ah I see. When we merge commits, per partition we keep the `OffsetAndMetadata` with the highest offset. It means we'll loose the metadata of earlier commits, but that is actually okay; those commits are no longer needed because they have been superseded by a newer one.

flavienbert:
One point is not cover by this PR: the metadata from the `CommittableRecord.offset` will be always empty: https://github.com/zio/zio-kafka/blob/6e451121b148e4997f1318b5807e83b8d0e6fdf9/zio-kafka/src/main/scala/zio/kafka/consumer/CommittableRecord.scala#L48. This PR only able to publish metadata from consumer stream, not retrieve it when you consume a stream. Is it something you want to add too @erikvanoosten? (if it's yes, is it same PR or an other)

flavienbert:
As I mentioned in the previous comment, I think is not possible to get metadata from a ZStream with the Consumer because the API `ConsumerRecord` doesn't expose the metadata.

flavienbert:
Maybe we should put this method private: 
`private[consumer] def metadata: Option[String]` in the Offset trait ? 
To make sure a user didn't try to get the metadata of an offset from the `ConsumerRecord`. Because It will always return None anyway otherwise.

erikvanoosten:
> Maybe we should put this method private: `private[consumer] def metadata: Option[String]` in the Offset trait ? To make sure a user doesn't try to get the metadata of an offset from the `ConsumerRecord`. Because It will always return None anyway otherwise.

Are you referring to a `CommittableRecord`?

I don't like that because it will mean that you can set the metadata but you can't read it back.

erikvanoosten (commented):
Not done yet, but I wanted to submit these comments already.

erikvanoosten:
Okay, I am changing my mind 🙂 The `offset` method from `CommitableRecord` should indeed not have a metadata field because that doesn't make sense. Unfortunately, that also means that the `withMetadata` method (the basis of this PR) doesn't make sense. (Off topic, there is more in `Offset` that doesn't make sense. In particular, there shouldn't be a `commit` method in it. @guizmaii is trying to change this part of zio-kafka for a good reason.)

What about adding methods `commitWithMetadata` and `commitOrRetryWithMetadata`?

flavienbert:
Agree with you `commit` method shouldn't be here. I am not sure `commitWithMetadata` could work because we usually use `commit` after `.aggregateAsync(Consumer.offsetBatches)` this means we call `commit` on the most recent  offset. 

But in my case I need to set the metadata for each offset, and then the most recent offset is commit with the metadata set earlier for sample: 

```
     ...
     offsetOpt  <- offsetState match {
                                 case Committable =>
                                    ZIO.some(record.offset)

                                 case CommittableWithMetadata =>
                                   StorageService
                                       .put(state, record.offset.topic, record.offset.partition)
                                       .map(versionId => Some(record.offset.withMetadata(versionId)))

                                 case NotCommittable => ZIO.none
                               }
        } yield offsetOpt
      }
      .map(_.toSeq)
      .flattenIterables
      .aggregateAsync(Consumer.offsetBatches)
      .mapZIO(_.commit)
``` 

And what about having this model:

```
sealed trait Offset {

  type Self <: Offset

  def topic: String
  def partition: Int
  def offset: Long
  def consumerGroupMetadata: Option[ConsumerGroupMetadata]

  /**
   * Converts this instance of Offset to a CommittableOffset (increment the offset)
   */
  def toCommittableOffset: CommittableOffset

  final lazy val topicPartition: TopicPartition = new TopicPartition(topic, partition)
}

sealed trait CommittableOffset extends Offset {

  type Self <: CommittableOffset

  def commit: Task[Unit]
  def batch: OffsetBatch
  def addMetadata(metadata: String): Self
  def metadata: Option[String]

  private[consumer] def asJavaOffsetAndMetadata: OffsetAndMetadata = new OffsetAndMetadata(offset, metadata.orNull)

  /**
   * Attempts to commit and retries according to the given policy when the commit fails with a
   * RetriableCommitFailedException
   */
  final def commitOrRetry[R](policy: Schedule[R, Throwable, Any]): RIO[R, Unit] =
    Offset.commitOrRetry(commit, policy)
}
```

With this model, we could use `Offset` trait for the offset get from the `CommitableRecord` and then if you want to commit or add metadata you should convert this Offset to a `CommitableOffset` which will be able to commit. And we could introduce an implicit conversion from `Offset` to `CommitableOffset` to avoid breaking changes.

erikvanoosten:
Okay, I understand your use case,
And I like your thinking. `CommittableOffset` is indeed a reasonable way to model this. This is a breaking change though, even with an implicit conversion it will not be binary compatible (we don't really have that as a goal so it could still be fine).
Since not many people use metadata for commits, IMHO source compatibility is required.

@guizmaii WDYT?

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
```diff
@@ -349,6 +349,30 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
...
+          // Consume messages
+          subscription = Subscription.topics(topic)
+          offsets <- (Consumer
+                       .partitionedStream(subscription, Serde.string, Serde.string)
+                       .flatMap(_._2.map(_.offset.withMetadata(metadata)))
+                       .take(1)
+                       .transduce(Consumer.offsetBatches)
+                       .take(1)
+                       .mapZIO(_.commit)
+                       .runDrain *>
```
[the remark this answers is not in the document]
flavienbert: in the function `provideSomeLayer` I need the variable `client` created in the for comprehension. So I can't use `provideSomeLayer` as you mentioned

flavienbert:
I published a version with `CommittableOffset`. To be honest I like this new version. I think it's much more understandable. There is different models between offsets get from the consumer and offsets that you want to commit.

However, there is one breaking change: you have to explicitly convert `ReadOnlyOffset` to `CommitableOffset`in order to commit offsets: `offset.nextCommittableOffset.commit`. But I prefer explicit conversion this way users can understand how it works internally. 

@erikvanoosten let me know your point of view.

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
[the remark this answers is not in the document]
erikvanoosten: Ah yes, I see.

flavienbert:
@erikvanoosten any updates? Could be great to have the metadata commitment feature merged.

erikvanoosten:
> @erikvanoosten any updates? Could be great to have the metadata commitment feature merged.

I have asked @svroonland and @guizmaii to also take a look at this PR.

Since we want to change this API in near future, and changing this many times is not nice for our users, for me source compatibility would be essential. Perhaps the other committers see this differently...

erikvanoosten:
BTW, this is a low risk change. If it doesn't work, we can simply revert and all the data is back! 🙂

zio-kafka/src/main/scala/zio/kafka/consumer/Offset.scala
```diff
@@ -1,25 +1,44 @@
...
+
+  final lazy val topicPartition: TopicPartition = new TopicPartition(topic, partition)
+}
+
+sealed trait ReadOnlyOffset extends Offset {
+
+  /**
+   * Converts this instance of Offset to a CommittableOffset (increment the offset)
+   */
+  def nextCommittableOffset: CommittableOffset
```
guizmaii, line 22: I don't like this name "nextCommittableOffset". The "next" feels like we're getting the next element in a list
I'd probably prefer something like `asCommittableOffset` or `toCommittableOffset`

zio-kafka/src/main/scala/zio/kafka/consumer/OffsetBatch.scala
```diff
@@ -24,30 +24,35 @@ sealed trait OffsetBatch {
...
+  override def merge(offset: CommittableOffset): OffsetBatch = add(offset)
 
   override def merge(otherOffsets: OffsetBatch): OffsetBatch = {
-    val newOffsets = Map.newBuilder[TopicPartition, Long]
+    val newOffsets = Map.newBuilder[TopicPartition, OffsetAndMetadata]
     newOffsets ++= offsets
     otherOffsets.offsets.foreach { case (tp, offset) =>
-      val existing = offsets.getOrElse(tp, -1L)
-      if (existing < offset)
+      val existing = offsets.getOrElse(tp, new OffsetAndMetadata(-1L))
```
guizmaii, line 54: We probably don't want to instantiate more than one `OffsetAndMetadata(-1L)`

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -125,15 +125,14 @@ private[consumer] final class Runloop private (
       } yield ()
 
   private def doCommit(cmd: RunloopCommand.Commit): UIO[Unit] = {
-    val offsets   = cmd.offsets.map { case (tp, offset) => tp -> new OffsetAndMetadata(offset + 1) }
```
guizmaii, line 128: I'm not convinced movint the `+ 1` logic out of this function makes things clearer. 
Making the `+ 1` has the advantage to be explicit about why we need to do this operation: that's part of the "committing process".

While with the new code, this `+ 1` is lost somewhere else and it's harder to make the link between this "commit algorithm" and this "+ 1" operation. 

Unless you convince me why you prefer your solution, I'd prefer to keep this "+ 1 operaton" here

zio-kafka/src/main/scala/zio/kafka/producer/TransactionalProducer.scala
```diff
@@ -39,12 +36,7 @@ object TransactionalProducer {
           offsetBatch.consumerGroupMetadata match {
             case None => invalidGroupIdException
             case Some(consumerGroupMetadata) =>
-              val offsets: util.Map[TopicPartition, OffsetAndMetadata] =
-                offsetBatch.offsets.map { case (topicPartition, offset) =>
-                  topicPartition -> new OffsetAndMetadata(offset + 1)
-                }.asJava
```
guizmaii, line 45: Same remark as https://github.com/zio/zio-kafka/pull/1067/files#r1354310142

zio-kafka-test/src/test/scala/zio/kafka/consumer/SubscriptionsSpec.scala
```diff
@@ -267,7 +267,9 @@ object SubscriptionsSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
             .plainStream(Subscription.topics(topic1), Serde.string, Serde.string)
             .take(40)
             .transduce(
-              Consumer.offsetBatches.contramap[CommittableRecord[String, String]](_.offset) <&> ZSink
+              Consumer.offsetBatches.contramap[CommittableRecord[String, String]](
+                _.offset.nextCommittableOffset
```
guizmaii, line 271: I'm not of a fan of this new interface we're offering to our users with this chance.
What's the point of exposing `CommittableRecord::offset` if they need to use `CommittableRecord::offset::nextCommittableOffset`?

If we follow the Law of Demeter or principle of least knowledge (See https://en.wikipedia.org/wiki/Law_of_Demeter) the users shouldn't have to use more than one `.`.
Something like:
```scala
val r: CommittableRecord = ...
r.nextCommittableOffset
```

Why do we need to continue to expose `.offset` if we use your next classes?

guizmaii (commented):
See comments

I think there is still some refinement work needed to make this PR mergeable

zio-kafka/src/main/scala/zio/kafka/consumer/Offset.scala
```diff
@@ -35,13 +54,32 @@ object Offset {
...
+    CommittableOffsetImpl(topic, partition, offset + 1, commitHandle, consumerGroupMetadata, metadata = None)
+}
+
+private final case class CommittableOffsetImpl(
+  topic: String,
+  partition: Int,
+  offset: Long,
+  commitHandle: Map[TopicPartition, OffsetAndMetadata] => Task[Unit],
+  consumerGroupMetadata: Option[ConsumerGroupMetadata],
+  metadata: Option[String]
```
guizmaii, line 74: Why do we need to create 2 new interfaces and 2 new classes to add this optional field?
Why not just add it to the already existing `Offset` trait? 🤔

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -125,7 +125,9 @@ private[consumer] final class Runloop private (
...
+    val offsets = cmd.offsets.map { case (tp, offset) =>
+      tp -> new OffsetAndMetadata(offset.offset + 1, offset.metadata)
```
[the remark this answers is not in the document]
erikvanoosten: @guizmaii I have no strong opinion on this yet, I was just exploring possibilities. Sorry for that @flavienbert, this aspect is totally not relevant for this PR. If we want this change (probably not) we can do it in another PR.

zio-kafka/src/main/scala/zio/kafka/consumer/Offset.scala
erikvanoosten: @guizmaii something like #1076 is what this PR had at first. @flavienbert then discovered an unfortunate consequence (https://github.com/zio/zio-kafka/pull/1067#issuecomment-1747238802).

Although @flavienbert 's PR fixes that problem now, the price (lower usability and backward incompatibility) is high.

Because of that I also prefer the previous design (like #1076). We can also 'fix' the problem by adding a little warning to the `withMetadata` method.

guizmaii: Quoting @flavienbert in the comment you pointed out:
> a user can believe he can retrieve the metadata offset of the CommittableRecord using _.offset.metadata but in fact it will always return None. We should only able to use _.offset.withMetadata("metadata") from my point of view

In my implementation, the `offset.metadata` is private, only the `.withMetadata` is accessible to the user so I'm not sure to see what's the issue 🤔

zio-kafka/src/main/scala/zio/kafka/consumer/Offset.scala
flavienbert: Yes the next is because of `offset + 1`. I think it's nice to put this logic here.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
flavienbert: We have same issue today: if someone instantiates an `OffsetImpl` with the `offset + 1` because he wants commit this offset, then on your commit method you do again +1. From my point of view it's obvious that if you instantiate a `CommittableOffsetImpl ` it is the offset you want to commit (meaning with offset +1 inside already compute). Then the function commit should only commit this offset. It looks obvious for me but maybe I am wrong

zio-kafka-test/src/test/scala/zio/kafka/consumer/SubscriptionsSpec.scala
flavienbert: Because a user should be able to get the current offset of the record. If you don't like the new interface, I get it. Check my first commit. It's much more easier and let me know if you prefer it.

zio-kafka/src/main/scala/zio/kafka/consumer/Offset.scala
flavienbert: Probably first commit is enough!

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: There is indeed a disconnect with how zio-kafka tracks offsets (last consumed offset) and what kafka expects (offset to consume). However, the core of this PR is to add metadata to the commits. Lets move the offset discussion to another PR, or maybe to Discord.

zio-kafka/src/main/scala/zio/kafka/consumer/Offset.scala
erikvanoosten: Would you mind rolling the branch back to that commit?

flavienbert: Ok should I push force the first commit? (with the first reviews fixed)?

erikvanoosten: Since all PR commits are eventually squashed anyway, you can do whatever works for you: force push the first commit, or add a new commit that reverts to that state...

erikvanoosten: BTW, if you take the force push route, then fon't forget to rebase on master again.

zio-kafka/src/main/scala/zio/kafka/consumer/Offset.scala
```diff
@@ -39,9 +46,17 @@ private final case class OffsetImpl(
   topic: String,
   partition: Int,
   offset: Long,
-  commitHandle: Map[TopicPartition, Long] => Task[Unit],
-  consumerGroupMetadata: Option[ConsumerGroupMetadata]
+  commitHandle: Map[TopicPartition, OffsetAndMetadata] => Task[Unit],
+  consumerGroupMetadata: Option[ConsumerGroupMetadata],
+  metadata: Option[String]
```
guizmaii, line 51: Needs to be `private[consumer]`

guizmaii, line 51:
```suggestion
  private[consumer] metadata: Option[String]
```

zio-kafka/src/main/scala/zio/kafka/consumer/Offset.scala
erikvanoosten: This is done.

zio-kafka/src/main/scala/zio/kafka/producer/TransactionalProducer.scala
```diff
@@ -41,7 +41,7 @@ object TransactionalProducer {
             case Some(consumerGroupMetadata) =>
               val offsets: util.Map[TopicPartition, OffsetAndMetadata] =
                 offsetBatch.offsets.map { case (topicPartition, offset) =>
-                  topicPartition -> new OffsetAndMetadata(offset + 1)
+                  topicPartition -> new OffsetAndMetadata(offset.offset + 1, offset.metadata)
```
erikvanoosten, line 44: We should retain leader epoch here as well.

zio-kafka/src/main/scala/zio/kafka/consumer/Offset.scala
flavienbert: should I remove private ?

erikvanoosten:
@flavienbert master changed quite a bit, would you like me to rebase it for you?

zio-kafka/src/main/scala/zio/kafka/consumer/Offset.scala
guizmaii: you need to add `private[consumer]`, same in the interface

flavienbert: The OffsetImpl class is already private.

zio-kafka/src/main/scala/zio/kafka/consumer/Offset.scala
```diff
@@ -39,9 +44,15 @@ private final case class OffsetImpl(
...
```
guizmaii, line 49: I think it'd be better to give a default value to `metadata: Option[String]`:
```scala
metadata: Option[String] = None
```

flavienbert:
I rebased my branch, But the CI fail for `sbt ciGenerateGithubWorkflow` but it seems to be good

.github/workflows/ci.yml
```diff
@@ -13,9 +13,7 @@ env:
     branches:
     - master
     - series/0.x
-  pull_request:
-    branches-ignore:
-    - gh-pages
```
guizmaii, line 18: You shouldn't have this change

zio-kafka/src/main/scala/zio/kafka/consumer/OffsetBatch.scala
guizmaii: @flavienbert

zio-kafka/src/main/scala/zio/kafka/producer/TransactionalProducer.scala
erikvanoosten: Never mind. I'll add it later.

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
```diff
@@ -349,6 +349,29 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
...
+                       Consumer.committed(Set(new TopicPartition(topic, 0))))
+                       .provideSomeLayer[Kafka](consumer(client, Some(group)))
```
erikvanoosten, line 372:
```suggestion
          // Consume messages
          subscription = Subscription.topics(topic)
          c            = consumer(client, Some(group))
          _ <- Consumer
                 .partitionedStream(subscription, Serde.string, Serde.string)
                 .flatMap(_._2.map(_.offset.withMetadata(metadata)))
                 .take(1)
                 .transduce(Consumer.offsetBatches)
                 .take(1)
                 .mapZIO(_.commit)
                 .runDrain
                 .provideSomeLayer[Kafka](c)
          // Fetch the committed offset
          offsets <- Consumer
                       .committed(Set(new TopicPartition(topic, 0)))
                       .provideSomeLayer[Kafka](c)
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -127,16 +127,21 @@ private[consumer] final class Runloop private (
...
   ): (JavaMap[TopicPartition, OffsetAndMetadata], OffsetCommitCallback, Throwable => UIO[Unit]) = {
     val offsets = commits
-      .foldLeft(mutable.Map.empty[TopicPartition, Long]) { case (acc, commit) =>
+      .foldLeft(mutable.Map.empty[TopicPartition, OffsetAndMetadata]) { case (acc, commit) =>
         commit.offsets.foreach { case (tp, offset) =>
-          acc += (tp -> acc.get(tp).map(_ max offset).getOrElse(offset))
+          acc += (tp -> acc
+            .get(tp)
+            .map(current => if (current.offset() > offset.offset()) current else offset)
+            .getOrElse(offset))
```
erikvanoosten, line 135:
```suggestion
          acc.updateWith(tp) { current =>
            current.map(current => if (current.offset() > offset.offset()) current else offset).orElse(Some(offset))
          }
```

erikvanoosten (changes requested):
Some small suggestions. Otherwise LGTM.

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
erikvanoosten: Never mind, not important.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Never mind, not important.

erikvanoosten (approved):

guizmaii (changes requested):
This one still need to be changed: https://github.com/zio/zio-kafka/pull/1067/files#r1354306081

zio-kafka/src/main/scala/zio/kafka/consumer/OffsetBatch.scala
erikvanoosten: I am not sure this is important. Using a constant means less memory locality (making the L2 useless).
Instead I would propose this logic is rewritten such that creating a dummy OffsetAndMetadata is not even necessary.

zio-kafka/src/main/scala/zio/kafka/consumer/Offset.scala
```diff
@@ -1,16 +1,21 @@
...
 
 sealed trait Offset {
+
   def topic: String
   def partition: Int
   def offset: Long
   def commit: Task[Unit]
   def batch: OffsetBatch
   def consumerGroupMetadata: Option[ConsumerGroupMetadata]
+  def withMetadata(metadata: String): Offset
```
svroonland, line 15: We can copy some documentation from here: https://kafka.apache.org/31/javadoc/org/apache/kafka/clients/consumer/OffsetAndMetadata.html

> The Kafka offset commit API allows users to provide additional metadata (in the form of a string) when an offset is committed. This can be useful (for example) to store information about which node made the commit, what time the commit was made, etc.

zio-kafka/src/main/scala/zio/kafka/consumer/OffsetBatch.scala
```diff
@@ -28,26 +28,31 @@ object OffsetBatch {
...
+      if (existing.offset < offset.offset)
         newOffsets += tp -> offset
```
erikvanoosten, line 56: This is the last thing that is holding the PR from merging. Let's address @guizmaii request by removing the need for the artificial offset completely. @flavienbert WDYT?

```suggestion
      val laterOffset = offsets.get(tp) match {
        case Some(existing) => if (existing.offset < offset.offset) offset else existing
        case None => offset
      }
      newOffsets += tp -> laterOffset
```

guizmaii (approved):

guizmaii:
@erikvanoosten You merge once the CI pass :)

guizmaii:
@flavienbert Thanks for your work :)

erikvanoosten:
@flavienbert Could you run `sbt fmt` please?

zio-kafka/src/main/scala/zio/kafka/consumer/OffsetBatch.scala
```diff
@@ -28,35 +28,42 @@ object OffsetBatch {
...
-    val newOffsets = Map.newBuilder[TopicPartition, Long]
+    val newOffsets = Map.newBuilder[TopicPartition, OffsetAndMetadata]
     newOffsets ++= offsets
     otherOffsets.offsets.foreach { case (tp, offset) =>
-      val existing = offsets.getOrElse(tp, -1L)
-      if (existing < offset)
-        newOffsets += tp -> offset
+      val laterOffset = offsets.get(tp) match {
+        case Some(existing) => if (existing.offset < offset.offset) offset else existing
+        case None => offset
```
erikvanoosten, line 56: To make the linter happy (sigh):

```suggestion
        case None           => offset
```

erikvanoosten: @flavienbert I tried to push this change but I do not have access to your fork.

erikvanoosten:
Thanks for you patience @flavienbert ! This PR is now available in https://github.com/zio/zio-kafka/releases/tag/v2.6.0 🥳

PR #1073, opened 2023-10-08
Automatically merge commits

erikvanoosten:
The commit throughput that is supported by Kafka brokers is much lower than the consume throughput. Therefore, to retain a high consume throughput it is important that not every record's offset is committed. In this PR we automatically merge all commits that were generated in the course of a single run of the runloop. This frees users from having to merge streams and do the commit merging themselves.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -15,6 +15,8 @@ import zio.kafka.consumer.internal.RunloopAccess.PartitionAssignment
 import zio.stream._
 
 import java.util
+import java.util.{ Map => JavaMap }
+import scala.collection.mutable
 import scala.jdk.CollectionConverters._
 
 //noinspection SimplifyWhenInspection,SimplifyUnlessInspection
@@ -25,6 +27,7 @@ private[consumer] final class Runloop private (
   pollTimeout: Duration,
   commitTimeout: Duration,
   runloopTimeout: Duration,
+  commitQueue: Queue[Runloop.Commit],
   commandQueue: Queue[RunloopCommand],
   lastRebalanceEvent: Ref.Synchronized[Option[Runloop.RebalanceEvent]],
   partitionsHub: Hub[Take[Throwable, PartitionAssignment]],
@@ -110,42 +113,16 @@ private[consumer] final class Runloop private (
     }
   }
 
+  /** This is the implementation behind the user facing api `Offset.commit`. */
   private val commit: Map[TopicPartition, Long] => Task[Unit] =
     offsets =>
       for {
         p <- Promise.make[Throwable, Unit]
-        _ <- commandQueue.offer(RunloopCommand.Commit(offsets, p)).unit
+        _ <- commitQueue.offer(Runloop.Commit(offsets, p))
+        _ <- commandQueue.offer(RunloopCommand.CommitAvailable)
         _ <- diagnostics.emit(DiagnosticEvent.Commit.Started(offsets))
         _ <- p.await.timeoutFail(CommitTimeout)(commitTimeout)
       } yield ()
-
-  private def doCommit(cmd: RunloopCommand.Commit): UIO[Unit] = {
-    val offsets   = cmd.offsets.map { case (tp, offset) => tp -> new OffsetAndMetadata(offset + 1) }
-    val cont      = (e: Exit[Throwable, Unit]) => cmd.cont.done(e).asInstanceOf[UIO[Unit]]
-    val onSuccess = cont(Exit.unit) <* diagnostics.emit(DiagnosticEvent.Commit.Success(offsets))
-    val onFailure: Throwable => UIO[Unit] = {
-      case _: RebalanceInProgressException =>
-        ZIO.logDebug(s"Rebalance in progress, retrying commit for offsets $offsets") *>
-          commandQueue.offer(cmd).unit
-      case err =>
-        cont(Exit.fail(err)) <* diagnostics.emit(DiagnosticEvent.Commit.Failure(offsets, err))
-    }
-    val callback =
-      new OffsetCommitCallback {
-        override def onComplete(offsets: util.Map[TopicPartition, OffsetAndMetadata], exception: Exception): Unit =
-          Unsafe.unsafe { implicit u =>
-            runtime.unsafe.run(if (exception eq null) onSuccess else onFailure(exception)).getOrThrowFiberFailure()
-          }
-      }
-
-    // We don't wait for the completion of the commit here, because it
-    // will only complete once we poll again.
-    consumer.runloopAccess { c =>
-      ZIO
-        .attempt(c.commitAsync(offsets.asJava, callback))
-        .catchAll(onFailure)
-    }
-  }
 
   /**
    * Does all needed to end revoked partitions:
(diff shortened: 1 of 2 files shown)


svroonland (changes requested):
Nice change, I agree this will be of benefit to users. Could you reduce the PR diff size though? Since the Runloop's behavior is complex and changes can have subtle unintended effects, small diffs really help to review this well. There's some changed comments, part of the commit code is moved to another place and the commit queue, like you mention, is not necessary for this change.

erikvanoosten:
> Could you reduce the PR diff size though?

Sure, I can split it further.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -555,6 +589,15 @@ private[consumer] object Runloop {
     ) extends RebalanceEvent
   }
 
+  // TODO: make `private` as soon as `State` has been moved into `Runloop`, see https://github.com/zio/zio-kafka/pull/1065
```
guizmaii, line 592: Can be made private now I think 🙂

guizmaii:
@erikvanoosten Can you make the changes @svroonland asked for, please? So we can more easily review your work 🙂

erikvanoosten:
> @erikvanoosten Can you make the changes @svroonland asked for, please? So we can more easily review your work 🙂

Yes, I am working on it!

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: This class is gone now.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -119,33 +122,57 @@ private[consumer] final class Runloop private (
...
 
-  private def doCommit(cmd: RunloopCommand.Commit): UIO[Unit] = {
-    val offsets   = cmd.offsets.map { case (tp, offset) => tp -> new OffsetAndMetadata(offset + 1) }
-    val cont      = (e: Exit[Throwable, Unit]) => cmd.cont.done(e).asInstanceOf[UIO[Unit]]
-    val onSuccess = cont(Exit.unit) <* diagnostics.emit(DiagnosticEvent.Commit.Success(offsets))
+  /** Merge commits and prepare parameters for calling `consumer.commitAsync`. */
+  private def asyncCommitParameters(
+    commits: Chunk[RunloopCommand.Commit]
+  ): (JavaMap[TopicPartition, OffsetAndMetadata], OffsetCommitCallback, Throwable => UIO[Unit]) = {
+    val offsets = commits
```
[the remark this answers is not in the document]
erikvanoosten: `OffsetBatch` is not a straight fit. If `commit` would not be in there, it would be worthwhile to make it work. But I don't see it now. IMHO we can/need to revisit this code together with introducing a more functional commit flow.

Also, IMHO, for now this good enough.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -385,7 +415,6 @@ private[consumer] final class Runloop private (
 
     cmd match {
       case req: RunloopCommand.Request => ZIO.succeed(state.addRequest(req))
-      case cmd: RunloopCommand.Commit  => doCommit(cmd).as(state.addCommit(cmd))
```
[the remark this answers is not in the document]
erikvanoosten: We need to handle commits before the other commands. The other commands might stop the subscription, or stuff like that. We have an open issue (#852) about not handling all the commits when the streams are closed. By handling commits first, this is somewhat prevented (we still need a more fundamental solution though).

Because we need the commit command first, we need to `collect` them.

(Aside: in the past we didn't use `collect` for commands, instead we pattern matched inside the `handleCommand` method. You can ask @guizmaii about this change.)

IMHO splitting commit handling over multiple methods will decrease legibility, and is already pretty complex.

Finally, please keep in mind that in the next PR the commits will come from a separate queue and the `collect` is replaced by pulling from the queue.

svroonland (approved):
Alright

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -119,33 +122,57 @@ private[consumer] final class Runloop private (
...
+    } else {
+      val (offsets, callback, onFailure) = asyncCommitParameters(commits)
+      val newState                       = state.addCommits(commits)
+      consumer.runloopAccess { c =>
+        // We don't wait for the completion of the commit here, because it
+        // will only complete once we poll again.
+        ZIO.attempt(c.commitAsync(offsets, callback))
+      }
         .catchAll(onFailure)
+        .as(newState)
```
guizmaii, line 174: I'd wrap this in a `ZIO.suspendSucced`:
```scala
ZIO.suspendSucced {
      val (offsets, callback, onFailure) = asyncCommitParameters(commits)
      val newState                       = state.addCommits(commits)
      consumer.runloopAccess { c =>
        ZIO.attempt(c.commitAsync(offsets, callback))
      }
        .catchAll(onFailure)
        .as(newState)
}
```

So that all the code is evaluated when the returned `IO` is evaluated. It makes code simpler to follow IMO

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -119,33 +122,57 @@ private[consumer] final class Runloop private (
...
+    commits: Chunk[RunloopCommand.Commit]
+  ): (JavaMap[TopicPartition, OffsetAndMetadata], OffsetCommitCallback, Throwable => UIO[Unit]) = {
+    val offsets = commits
+      .foldLeft(mutable.Map.empty[TopicPartition, Long]) { case (acc, commit) =>
+        commit.offsets.foreach { case (tp, offset) =>
+          acc += (tp -> acc.get(tp).map(_ max offset).getOrElse(offset))
+        }
+        acc
+      }
+      .toMap
```
guizmaii, line 136: I'd not transform it to an immutable Map here if we transform it to a Java Map later. Feels like an unnecessary extra step

guizmaii:
Super smart change! Thanks for this Erik 🙂

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Good idea. I'll prepare another PR with that change.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Sure, we can do that as well. Its often a balance between readability and performance. In this going for a java map directly would be about as readable :)

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Hmmm, this is a pure method, it doesn't return a ZIO so this suggestion doesn't make sense.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: @guizmaii Exploring the idea here: https://github.com/zio/zio-kafka/pull/1079

PR #1077, opened 2023-10-11
Add instructions for pruning the benchmark history

erikvanoosten:
Mentioned shell script was already pushed to the gh-pages branch in https://github.com/zio/zio-kafka/commit/576a196496aaa98a73d679813905628048fd212f

### DIFF
--- a/docs/prune-benchmark-history.md
+++ b/docs/prune-benchmark-history.md
@@ -1,0 +1,16 @@
+# Pruning the benchmarks report page
+
+Benchmarks are run on every commit, even in pull requests. This causes the historic graphs on
+https://zio.github.io/zio-kafka/dev/bench/ to contain so many results that it becomes unreadable. Rather than limiting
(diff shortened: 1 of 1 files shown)


docs/prune-benchmark-history.md
```diff
@@ -0,0 +1,16 @@
+# Pruning the benchmarks report page
```
guizmaii, line 1: This will be put in the doc in the ZIO website. As it's something dedicated to the zio-kafka maintainer I don't think we should put it there but maube just as a markdown file at the root of the repo, WDYT?

erikvanoosten: I see. Yeah, you're probably right that other people don't care much about this. I'll move it the root.

erikvanoosten: We can start a DEVELOPERS.md document... 🙂

guizmaii (approved):

PR #1081, opened 2023-10-14
Add tuning options and docs

erikvanoosten:
Also: remove the no longer needed `defaultRunloopTimeout`.

### DIFF
--- a/docs/tuning.md
+++ b/docs/tuning.md
@@ -9,6 +9,13 @@ val settings = ConsumerSettings(bootstrapServers)
   .... etc.
 ```
 
+## Names
+
+Two very similarly named settings are relevant. Don't mix them up:
+
+* `pollTimeout` — how long a poll may take
+* `max.poll.interval.ms` — the maximum time between polls
+
 ## Throughput and latency
 
 The kafka client can be tuned for either high throughput or low latency, unfortunately not both.
--- a/docs/tuning.md
+++ b/docs/tuning.md
@@ -14,9 +14,13 @@ val settings = ConsumerSettings(bootstrapServers)
 The kafka client can be tuned for either high throughput or low latency, unfortunately not both.
 The most important settings for tuning throughput and latency are:
 
-* poll timeout — This is the maximum time to block while polling the Kafka consumer. Zio-kafka's default is 50ms which is good for low latency applications. Set this higher, e.g. `500ms` for better throughput.
-* [configuration `max.poll.records`](https://kafka.apache.org/documentation/#consumerconfigs_max.poll.records) — The maximum number of records a poll will return. Kafka defaults this to 500. You can set this higher for more throughput, or lower for lower latency.
-* the fetch strategy parameter `maxPartitionQueueSize` — when the number of records in a partition queue is below this value, zio-kafka will start to pre-fetch more records from Kafka. The default value for this parameter is `1024`; 2 * the default `max.poll.records` of 500, rounded to the nearest power of 2.
+* poll timeout — This is the maximum time to block while polling the Kafka consumer. Zio-kafka's default is 50ms which
+  is good for low latency applications. Set this higher, e.g. `500ms` for better throughput.
+* [configuration `max.poll.records`](https://kafka.apache.org/documentation/#consumerconfigs_max.poll.records) — The maximum number of records a poll will return. Kafka defaults this to
+  500. You can set this higher for more throughput, or lower for lower latency.
+* the fetch strategy parameter `partitionPreFetchBufferLimit` — when the number of records in a partition queue is
+  below this value, zio-kafka will start to pre-fetch and buffer more records from Kafka. The default value for this
+  parameter is `1024`; 2 * the default `max.poll.records` of 500, rounded to the nearest power of 2.
 
 Methods `ConsumerSettings.tuneForHighThroughput` and `ConsumerSettings.tuneForLowLatency` make it easy to configure
 these parameters:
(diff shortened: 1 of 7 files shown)


docs/tuning.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+on older zio-kafka version `withMaxPollInterval` is not available, use the following instead:
+
+```scala
+withProperty("max.poll.interval.ms", "900000") // 15 minutes
+```
+
+⚠️In zio-kafka versions 2.2 up to 2.5.0 it may also be necessary to tune the `runloopTimeout` setting.
+When no stream is processing data for this amount of time, the consumer will halt with a failure.
```
svroonland, line 51: This suggests that when you have an idle topic, zio-kafka would stop processing, which is not the case, it will keep on polling.

docs/tuning.md
```diff
@@ -0,0 +1,51 @@
...
+ConsumerSettings.withMaxPollInterval(15.minutes)
+```
+
+on older zio-kafka version `withMaxPollInterval` is not available, use the following instead:
+
+```scala
+withProperty("max.poll.interval.ms", "900000") // 15 minutes
+```
+
+⚠️In zio-kafka versions 2.2 up to 2.5.0 it may also be necessary to tune the `runloopTimeout` setting.
```
svroonland, line 50: tune -> increase, actually

docs/tuning.md
```diff
@@ -0,0 +1,51 @@
...
+* [configuration `max.poll.records`](https://kafka.apache.org/documentation/#consumerconfigs_max.poll.records) — The maximum number of records a poll will return. Kafka defaults this to 500. You can set this higher for more throughput, or lower for lower latency.
+
+Methods `ConsumerSettings.tuneForHighThroughput` and `ConsumerSettings.tuneForLowLatency` make it easy to configure
+these parameters:
+
+```scala
+val highThroughputSettings = ConsumerSettings(bootstrapServers).tuneForHighThroughput
+val lowLatencySettings = ConsumerSettings(bootstrapServers).tuneForLowLatency
+```
+
```
svroonland, line 27: Should we add the `maxPartitionQueueSize` here? At some point it becomes a bottleneck for throughput, especially if you don't have too many partitions

docs/tuning.md
```diff
@@ -0,0 +1,51 @@
...
+Zio-kafka's consumer can be tuned with the `ConsumerSettings` class. 
+
+```scala
+val settings = ConsumerSettings(bootstrapServers)
+  .withGroupId(groupId)
+  .withProperties(properties)
+  .... etc.
+```
+
+## Throughput and latency
```
svroonland, line 12: We can also recommend here to use partitionedStream and do batched commits for higher throughput.

svroonland (commented):
Nice! Some thoughts

docs/tuning.md
erikvanoosten: Correct! I will update the text.

docs/tuning.md
erikvanoosten: Right, when there is a high variability in processing time this might become a bottleneck. Also, when you set `max.poll.records` really high, you don't want to disable pre-fetching with a low `maxPartitionQueueSize`. Does that cover it?

docs/tuning.md
erikvanoosten: Definitely right.
A new can of worms though. Shall we leave that for another PR?

docs/tuning.md
erikvanoosten: I just realized that `maxPartitionQueueSize` is not an absolute maximum. When a chunk comes in that is larger, the queue will contain more than that.

svroonland: Right..! It's the queue size at which no further data fetches will be executed for that partition. If we can summarize that in a nice variable name..

docs/tuning.md
erikvanoosten: Fixed in https://github.com/zio/zio-kafka/pull/1081/commits/7703140375862e10b170ccd81d9529181b747682

docs/tuning.md
erikvanoosten: In https://github.com/zio/zio-kafka/pull/1081/commits/7703140375862e10b170ccd81d9529181b747682

docs/tuning.md
erikvanoosten: WDYT of:
```
* @param fetchMaxQueueSize
*   The queue size below which more records are fetched and buffered (per partition).
```

svroonland: Hmm not necessarily better. `partitionPreFetchBufferLimit` ? 😅

erikvanoosten:
Note: the test are failing, this is fixed in #1082 .

docs/tuning.md
erikvanoosten: That is already better indeed.
Addressed in https://github.com/zio/zio-kafka/pull/1081/commits/b34d73a2dcb68d315cf28430d4c0e858a3a4e021

docs/tuning.md
```diff
@@ -0,0 +1,84 @@
...
+The kafka client can be tuned for either high throughput or low latency, unfortunately not both.
+The most important settings for tuning throughput and latency are:
+
+* poll timeout — This is the maximum time to block while polling the Kafka consumer. Zio-kafka's default is 50ms which
+  is good for low latency applications. Set this higher, e.g. `500ms` for better throughput.
+* [configuration `max.poll.records`](https://kafka.apache.org/documentation/#consumerconfigs_max.poll.records) — The maximum number of records a poll will return. Kafka defaults this to
+  500. You can set this higher for more throughput, or lower for lower latency.
+* the fetch strategy parameter `partitionPreFetchBufferLimit` — when the number of records in a partition queue is
+  below this value, zio-kafka will start to pre-fetch and buffer more records from Kafka. The default value for this
+  parameter is `1024`; 2 * the default `max.poll.records` of 500, rounded to the nearest power of 2.
```
guizmaii, line 30: 🟢 Maybe we could make it clearer which config is a zio-kafka config and which is a Kafka client config

```
* zio-kafka's `pollTimeout` - ...
* Kafka clients' [`max.poll.records`](https://kafka.apache.org/documentation/#consumerconfigs_max.poll.records) - ...
* zio-kafka's `partitionPreFetchBufferLimit` - ...
```

WDYT?

docs/tuning.md
```diff
@@ -0,0 +1,84 @@
...
+
+Methods `ConsumerSettings.tuneForHighThroughput` and `ConsumerSettings.tuneForLowLatency` make it easy to configure
```
guizmaii, line 32: I'd write it like this:
```
Zio-kafka also provides 2 methods to help you tune these parameters of your consumers: `ConsumerSettings::tuneForHighThroughput` and `ConsumerSettings::tuneForLowLatency`
Each one of these methods will tune values of the `pollTimeout`, `max.poll.records` and `partitionPreFetchBufferLimit` configs to optimize your consumer according to your needs
```

WDYT?

docs/tuning.md
```diff
@@ -0,0 +1,84 @@
...
+val lowLatencySettings = ConsumerSettings(bootstrapServers).tuneForLowLatency
+```
+
+Kafka’s performance is not very sensitive to record size. However, when records become very small (< 100 bytes) or very
+large (> 100Kb), increasing or decreasing `max.poll.records` and `partitionPreFetchBufferLimit` can be considered.
+
+## High number of partitions
+
+When a lot of partitions need to be consumed, we need to take into account that heap is needed to store the records in
+the partition queues. A very rough estimate for the maximum amount of heap needed is given by: `average record size` *
```
guizmaii, line 46:
```suggestion
the partitions queue. A very rough estimate for the maximum amount of heap needed is given by: `average record size` *
```

I'm not sure about this one as I'm not a native English speaker

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -35,6 +35,42 @@ final case class ConsumerSettings(
...
+   *   [[ConsumerSettings.withMaxPollRecords]]
+   * @see
+   *   [[zio.kafka.consumer.fetch.QueueSizeBasedFetchStrategy]]
+   */
+  def tuneForLowLatency: ConsumerSettings =
+    this
+      .withPollTimeout(50.millis)
+      .withMaxPollRecords(100)
+      .withFetchStrategy(QueueSizeBasedFetchStrategy(partitionPreFetchBufferLimit = 512))
+
```
guizmaii, line 73: Amazing additions! 🔥😍

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -122,34 +158,39 @@ final case class ConsumerSettings(
...
+   *
+   * The default value for this parameter is 1024. It is calculated by taking 2 * the default `max.poll.records` of 500,
+   * rounded to the nearest power of 2.
    */
-  def withMaxPartitionQueueSize(maxPartitionQueueSize: Int): ConsumerSettings =
-    copy(fetchStrategy = QueueSizeBasedFetchStrategy(maxPartitionQueueSize))
+  def withPartitionPreFetchBufferLimit(partitionPreFetchBufferLimit: Int): ConsumerSettings =
+    copy(fetchStrategy = QueueSizeBasedFetchStrategy(partitionPreFetchBufferLimit))
+
+  @deprecated("Use withPartitionPreFetchBufferLimit instead", "3.5.1")
```
guizmaii, line 175:
```suggestion
  @deprecated("Use withPartitionPreFetchBufferLimit instead", "2.5.1")
```

guizmaii (approved):
Made a few suggestions

Great great great addition to zio-kafka!  
Thank you so much for this Erik! 
It'll make it even easier for beginners to start with Kafka and zio-kafka! 
🔥👏

docs/tuning.md
erikvanoosten: Yep, better.

docs/tuning.md
guizmaii: I'm always confused by this in English 😅

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
guizmaii: I don't mind. Up to you as you'll make the release as you made all the work for this release 😄🙇‍♂️

erikvanoosten: I would be happy to learn how to do a release.

guizmaii: You just need to make a release in GitHub (aka transform the "draft release" automatically made by release-drafter to a "release"). That will trigger the CI and publish everything automaticallly

erikvanoosten: That easy. Nice.

PR #1082, opened 2023-10-14
Fix flaky test

erikvanoosten:
Also: switch to nano clock to avoid the time jumps that come with the regular system clock.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
@@ -82,6 +82,8 @@ final class PartitionStreamControl private (
 
 object PartitionStreamControl {
 
+  private type NanoTime = Long
+
   private[internal] def newPartitionStream(
     tp: TopicPartition,
     commandQueue: Queue[RunloopCommand],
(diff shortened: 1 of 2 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -88,10 +88,12 @@ object PartitionStreamControl {
     diagnostics: Diagnostics,
     maxPollInterval: Duration
   ): UIO[PartitionStreamControl] = {
+    val maxPollIntervalNanos = maxPollInterval.toNanos
```
svroonland, line 91: What's the reason we need it in nano's?

erikvanoosten: Millisecond precision is good enough for our use case. However, the system clock follows the actual time and may make jumps when the clock is adjusted. The nano-clock never makes jumps, it is monotonic, and therefore more suitable for measuring durations.

svroonland: I see. Is there some type from java.time that represents nano time, instead of `Long`?

erikvanoosten: Not that I know off. Even zio gave up and returns Long 🤷 .
I noticed that cats uses FiniteDuration for that, but that is super ugly.
We can create a type alias...

erikvanoosten: Fixed in https://github.com/zio/zio-kafka/pull/1082/commits/88ab0b7fc0b1d31a0a1721b077986c184f1e5724

erikvanoosten:
@svroonland ready for second review.

svroonland (approved):

erikvanoosten:
Unfortunately, the test is still flaky. See #1083.

PR #1088, opened 2023-10-27
Prevent double logging in tests

erikvanoosten:
Also: add some useful loggers in the logback configuration.

### DIFF
--- a/zio-kafka-test/src/test/scala/zio/kafka/ZIOSpecDefaultSlf4j.scala
+++ b/zio-kafka-test/src/test/scala/zio/kafka/ZIOSpecDefaultSlf4j.scala
@@ -11,7 +11,9 @@ import zio.test.{ TestAspect, TestAspectAtLeastR, TestEnvironment, ZIOSpecDefaul
  */
 abstract class ZIOSpecDefaultSlf4j extends ZIOSpecDefault {
 
+  private val loggerLayer = SLF4J.slf4j
+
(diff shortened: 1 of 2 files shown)


zio-kafka-test/src/test/scala/zio/kafka/ZIOSpecDefaultSlf4j.scala
```diff
@@ -11,7 +11,9 @@ import zio.test.{ TestAspect, TestAspectAtLeastR, TestEnvironment, ZIOSpecDefaul
  */
 abstract class ZIOSpecDefaultSlf4j extends ZIOSpecDefault {
 
+  private val loggerLayer = SLF4J.slf4j
+
   override def aspects: Chunk[TestAspectAtLeastR[TestEnvironment]] =
-    super.aspects :+ TestAspect.fromLayer(zio.Runtime.removeDefaultLoggers >>> SLF4J.slf4j)
+    super.aspects :+ TestAspect.fromLayer(zio.Runtime.removeDefaultLoggers >>> loggerLayer)
```
svroonland, line 17: What's the actual change here?

erikvanoosten: I will not pretend that I understand why this works. Before it prints every log twice, and after not any more 🤷

svroonland (approved):

PR #1089, opened 2023-10-27
Always end streams in rebalance listener, support lost partitions

erikvanoosten:
Previously, streams would only be ended in the rebalance listener for revoked partitions. Now, they are ended there even when `restartStreamsOnRebalancing` is used.

Lost partitions are no longer treated as being revoked. With this change, streams of lost partitions are interrupted. Interrupting them prevents these streams from processing and committing more data.

A nice side effect is that Zio-kafka is now faster when the rebalance listener was _not_ called; the 'fast track'.

The main reason for this change is to prepare awaiting commits from within the rebalance listener which will prevent duplicate consuming of records (see #830).

Also: fix test `restartStreamsOnRebalancing mode closes all partition streams` so that it detects rebalances properly on fast computers.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -28,7 +28,7 @@ private[consumer] final class Runloop private (
   maxPollInterval: Duration,
   commitTimeout: Duration,
   commandQueue: Queue[RunloopCommand],
-  lastRebalanceEvent: Ref.Synchronized[Option[Runloop.RebalanceEvent]],
+  lastRebalanceEvent: Ref.Synchronized[Runloop.RebalanceEvent],
   partitionsHub: Hub[Take[Throwable, PartitionAssignment]],
   diagnostics: Diagnostics,
   offsetRetrieval: OffsetRetrieval,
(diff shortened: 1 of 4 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -297,101 +302,86 @@ private[consumer] final class Runloop private (
...
+                  )
 
-            } yield Runloop.PollResult(
-              startingTps = startingTps,
-              pendingRequests = revokeResult.pendingRequests,
-              assignedStreams = revokeResult.assignedStreams,
-              records = polledRecords,
-              ignoreRecordsForTps = ignoreRecordsForTps
-            )
+                case RebalanceEvent(true, assignedTps, revokedTps, lostTps, endedStreams) =>
```
svroonland, line 328: Perhaps I'm overlooking it, is the rebalance event reset somewhere?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -587,14 +576,37 @@ private[consumer] object Runloop {
...
 
-  private sealed trait RebalanceEvent
+  private final case class RebalanceEvent(
+    wasInvoked: Boolean,
+    assignedTps: Set[TopicPartition],
+    revokedTps: Set[TopicPartition],
+    lostTps: Set[TopicPartition],
+    endedStreams: Chunk[PartitionStreamControl]
+  ) {
+    def onAssigned(assigned: Set[TopicPartition], endedStreams: Chunk[PartitionStreamControl]): RebalanceEvent =
```
svroonland, line 586: I would have expected these as methods on the companion object instead of the instance, but probably several of these methods can be called in one poll loop?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
svroonland: Never mind, found it in `handlePoll`

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -71,45 +71,48 @@ private[consumer] final class Runloop private (
...
+          _ <- ZIO.logTrace("onRevoked done")
+        } yield (),
+      onLost = (lostTps, _) =>
+        for {
+          _              <- ZIO.logDebug(s"${lostTps.size} partitions are lost")
+          _              <- diagnostics.emit(DiagnosticEvent.Rebalance.Lost(lostTps))
+          rebalanceEvent <- lastRebalanceEvent.get
+          state          <- currentStateRef.get
+          (lostStreams, remainingStreams) = state.assignedStreams.partition(control => lostTps.contains(control.tp))
+          _ <- ZIO.foreachDiscard(lostStreams)(_.lost)
```
svroonland, line 106: 👍  Makes sense

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -71,45 +71,48 @@ private[consumer] final class Runloop private (
...
+          streamsToEnd = if (restartStreamsOnRebalancing && !rebalanceEvent.wasInvoked) remainingStreams
```
svroonland, line 107: I wonder if it's necessary to restart all other streams in case of a partition lost event. Does the consumer group generation change in that case? Probably hard to test and the behavior is not that important because it's an exceptional event.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -297,101 +302,86 @@ private[consumer] final class Runloop private (
...
+              val providedTps         = polledRecords.partitions().asScala.toSet
+              val requestedPartitions = state.pendingRequests.map(_.tp).toSet
+
+              DiagnosticEvent.Poll(
+                tpRequested = requestedPartitions,
+                tpWithData = providedTps,
+                tpWithoutData = requestedPartitions -- providedTps
+              )
+            } *>
+              lastRebalanceEvent.get.flatMap {
```
svroonland, line 315: If you use `getAndSet(RebalanceEvent.None).flatMap { .. }` it becomes more obvious that it's reset after reading it, where the reset is now located in `handlePoll`

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Yes, line 292.

svroonland (commented):
Nice, looks familiar from #591 of course. Good to have this in a separate PR.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: My thinking was that when we lose a partition it must be a rebalance. But you have a point, we don't have to consider it as one. Indeed, it is so exceptional that I don't even know how to test it.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: I thought that setting it before poll would be more obvious. Though this choice has the downside that the 2 operations are many lines apart. I am fine with either approach. Would you like me to change it?

svroonland: I think it would be a nice benefit yes

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Yes, the idea is that multiple can be called during the same poll.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Done in https://github.com/zio/zio-kafka/pull/1089/commits/1db802e9db6f1029cd5c25791b3a8d0c43c6335c

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Suggestion applied in https://github.com/zio/zio-kafka/pull/1089/commits/a399dfe284de95ac005959f489d3e61cc272fa4f

erikvanoosten:
@svroonland @guizmaii For at least week I have been trying to patch up the test that fails on Github action runner but passes on my laptop. Yesterday I was fed up and decided to take a better look at what the test does and why it could fail. I discovered some interesting things:

* The test was complete rubbish; it didn't test at all what it was supposed to test.
* The test was completely doomed; it is impossible to observe (white box test) the restartStreamsOnRebalance feature because (as far as I can tell) the default partition assignor externally behaves the same as the restartStreamsOnRebalance feature.

(Did you notice I like drama? 😉)  Anyways, I wrote a new test that actually tests whether streams are restarted.

BTW, this PR could use some review love. Thanks in advance!

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -71,45 +71,45 @@ private[consumer] final class Runloop private (
...
-          lastRebalanceEvent.updateZIO {
-            case None =>
-              ZIO.some(Runloop.RebalanceEvent.Assigned(assigned))
-            case Some(Runloop.RebalanceEvent.Revoked(revokeResult)) =>
-              ZIO.some(Runloop.RebalanceEvent.RevokedAndAssigned(revokeResult, assigned))
-            case Some(_) =>
-              ZIO.fail(new IllegalStateException(s"Multiple onAssigned calls on rebalance listener"))
-          },
-      onRevoked = (_, _) =>
+    val recordRebalanceRebalancingListener = RebalanceListener(
```
svroonland, line 74: Shall we add a comment here describing the high level behavior during rebalances

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -302,101 +303,86 @@ private[consumer] final class Runloop private (
...
+                      state.assignedStreams.filter(s => !endedTps.contains(s.tp)) ++ startingStreams
+
+                    // Remove pending requests for all streams that ended:
+                    //  1. streams that were ended because the partition was lost
+                    //  2. streams that were ended because the partition was revoked
+                    //  3. streams that were ended because of `restartStreamsOnRebalancing` being true
+                    updatedPendingRequests =
+                      state.pendingRequests.filter { pendingRequest =>
+                        val tp = pendingRequest.tp
+                        !(lostTps.contains(tp) || revokedTps.contains(tp) || endedStreams.exists(_.tp == tp))
```
svroonland, line 363: Doesn't `endedStreams` contain both lost and revoked TPs?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -302,101 +303,86 @@ private[consumer] final class Runloop private (
...
+                      }
+
+                    updatedAssignedStreams =
+                      state.assignedStreams.filter(s => !endedTps.contains(s.tp)) ++ startingStreams
+
+                    // Remove pending requests for all streams that ended:
+                    //  1. streams that were ended because the partition was lost
+                    //  2. streams that were ended because the partition was revoked
+                    //  3. streams that were ended because of `restartStreamsOnRebalancing` being true
+                    updatedPendingRequests =
```
svroonland, line 360: Could there be some race condition where a new pending request for a lost partition is enqueued (put in the State) after a poll?

svroonland (approved):
Nice work! A few thinks to consider

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Sure!

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: No, `endedStreams` are always from revoked tps, lost tps are not part of it.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Good question! Took me some time to figure this out.

If we look at the stream that is constructed in `PartitionStreamControl`, we see that the place where the requests are created are upstream from the place where the 'lost' interrupt happens. So we know for certain that requests are not sent after the stream interrupt took place.

_However,_ the stream runs independent from the runloop so it can happen that a `Request` was put in the command queue just before the stream interrupt. This is also possible for streams that end in the regular way!

In short, while handling commands, we need to filter out `Request`s for streams that no longer exist. If we don't do this, `pendingRequests` will for ever be non empty, meaning that `State.shouldPoll` will forever return `true`...

erikvanoosten: I am wondering, isn't this already a problem?

erikvanoosten: Fixed in https://github.com/zio/zio-kafka/pull/1089/commits/e9aa756a87406e5e1f9ec55afb8d21461db5e9a7

PR #1091, opened 2023-10-28
Make `partitionPreFetchBufferLimit` value `0` disable prefetching

erikvanoosten:
Before you need to use `ConsumerSettings.partitionPreFetchBufferLimit(1)` to disable pre-fetching. I found that very confusing. After this change you can disable pre-fetching with `ConsumerSettings.partitionPreFetchBufferLimit(0)`, or simply with `ConsumerSettings.withoutPartitionPreFetching`.

Also: attempt to make test `a consumer timeout interrupts the stream and shuts down the consumer` stable by disabling prefetching

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
@@ -177,7 +177,7 @@ final case class ConsumerSettings(
   /**
    * Disables partition record pre-fetching.
    */
-  def withoutPartitionPreFetching(): ConsumerSettings =
+  def withoutPartitionPreFetching: ConsumerSettings =
     withPartitionPreFetchBufferLimit(0)
 
   @deprecated("Use withPartitionPreFetchBufferLimit instead", "2.6.0")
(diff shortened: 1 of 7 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -159,19 +159,27 @@ final case class ConsumerSettings(
...
-   *   controlled by the `max.poll.records` setting, the number of records fetched for every partition is somewhere
-   *   between 0 and `max.poll.records`. A value that is a power of 2 offers somewhat better queueing performance.
+   *   The queue size at or below which more records are fetched and buffered (per partition). This buffer improves
+   *   throughput and supports varying downstream message processing time, while maintaining some backpressure. Large
+   *   values effectively disable backpressure at the cost of high memory usage, low values will effectively disable
+   *   prefetching in favor of low memory consumption. The number of records that is fetched on every poll is controlled
+   *   by the `max.poll.records` setting, the number of records fetched for every partition is somewhere between 0 and
+   *   `max.poll.records`. A value that is a power of 2 offers somewhat better queueing performance.
    *
    * The default value for this parameter is 1024. It is calculated by taking 2 * the default `max.poll.records` of 500,
```
svroonland, line 169: Not quite related, but I just noticed that according to the zio docs, it looks like the rounding to the power of 2 only applies to bounded queues, while we're using an unbounded queue.

svroonland (approved):
👍

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
erikvanoosten: Fixed in https://github.com/zio/zio-kafka/pull/1091/commits/82696d5869720decbc3170a650b35c45ab65c392

PR #1042, opened 2023-09-09
[POC - 3] New commit interface

guizmaii:

PR #1093, opened 2023-10-29
Consumer.commit method instead of record.offset.commit

svroonland:
Enforces that a Consumer is in scope when committing an offset, instead of `record.offset.commit`. 

Also:
* Remove deprecated merge method from OffsetBatch
* Combine OffsetBatch.add and OffsetBatch.commit

See also #1022, #1042.


erikvanoosten:
I like where this is going... :)

svroonland:
Cool. This would be a backwards-incompatible change, so it has to be worth the benefits. We could perhaps deprecate the old commit method for a while.

erikvanoosten:
> Cool. This would be a backwards-incompatible change, so it has to be worth the benefits. We could perhaps deprecate the old commit method for a while.

Yes, definitely.

PR #1097, opened 2023-11-05
Track latest completed commit offset per partition

erikvanoosten:
By tracking these offsets we can skip awaiting already completed commits from the rebalance listener in #830.

To prevent unbounded memory usage, after a rebalance we remove the committed offset for partitions that are no longer assigned to this consumer.

Note that a commit might complete just after a partition was revoked. This is not a big issue; the offset will still be removed in the next rebalance. When the `rebalanceSafeCommits` feature is available (see #830) commits will complete in the rebalance listener and this cannot happen anymore.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -14,6 +14,7 @@ import zio.kafka.consumer.internal.Runloop._
 import zio.kafka.consumer.internal.RunloopAccess.PartitionAssignment
 import zio.stream._
 
+import java.lang.Math.max
 import java.util
 import java.util.{ Map => JavaMap }
 import scala.collection.mutable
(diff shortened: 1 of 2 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -700,4 +712,24 @@ private[consumer] object Runloop {
       subscriptionState = SubscriptionState.NotSubscribed
     )
   }
+
+  // package private for unit testing
+  private[internal] final case class CommitOffsets(offsets: Map[TopicPartition, Long]) {
```
svroonland, line 717: Would it help to reduce allocations here with `extends AnyVal`?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -700,4 +712,24 @@ private[consumer] object Runloop {
...
+    def addCommits(c: Chunk[RunloopCommand.Commit]): CommitOffsets =
```
svroonland, line 718: Looks a lot like OffsetBatch.merge, could you reuse that somehow?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -35,6 +35,7 @@ private[consumer] final class Runloop private (
   userRebalanceListener: RebalanceListener,
   restartStreamsOnRebalancing: Boolean,
   currentStateRef: Ref[State],
+  completedCommitsRef: Ref[CommitOffsets],
```
svroonland, line 38: Instead of 'completed commits', what do you think about 'committed offsets'?

erikvanoosten: You're right that is better. I named it 'completed commits' because it is nicely symmetric with 'pending commits'. However, they store slightly different things so no symmetry is needed.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Neh, not at all. We store `CommitOffsets` instances in a `Ref` and the update methods do not generate some intermediate instance.

BTW, I wrapped the map in a case class for 2 reasons:
* It provides a very nice place to put the updating methods.
* Having updating methods makes the code that uses `CommitOffsets` very concise.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: It is virtually the same operation but with different types.
BTW, using a map builder is probably faster (as in `OffsetBatch.merge`), I'll rewrite `addCommits` to look more like `OffsetBatch.merge`.

erikvanoosten: Nope, my comment above is wrong. `addCommits` is quite different because it needs to combine many offset maps while `OffsetBatch.merge` only needs to merge 2. As a consequence we can't even use a map builder, map builders do not have the `get` operation (every time you add an offset, you need to lookup the highest offset so far).
Maybe we can use a mutable map. Lets see...

erikvanoosten: Fixed in https://github.com/zio/zio-kafka/pull/1097/commits/7292e740e1db2b4f12b5f514ed688532eb06703a

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Fixed in https://github.com/zio/zio-kafka/pull/1097/commits/cb68c82159192dd70ea7fc6cdfeb3b7e65a927ed

svroonland (approved):

Issue #1087, opened 2023-10-26
Unexpected behavior during graceful shutdown


bsaletti:
Hi,

We are observing some unexpected behavior when we do a graceful shutdown on applications running zio-kafka.

What we expected or were hopping to see was that a graceful shutdown would finish processing any messages already in flight but would stop consuming new messages. Once all in flight messages are committed it should shut down the consumer and
cleanup all ressources.

What we are observing is that partitions are reallocated right away and therefore picked up by other instances while some messages are still being processed and are not yet committed.
This results in the same message being processed twice, once in the instance that is shutting down and once in whatever instance gets assigned the partitions.

Could it be that the consumer gets closed too early ?

erikvanoosten:
Thanks for your report. The problem is that zio-kafka does not take into account what the streams are doing.

Right now we are fixing the related problem of preventing duplicates during a rebalance (#830). Once this is in place, it will be fairly straight forward to fix this problem as well.

bsaletti-fgh:
Thank you for your reply, we're waiting for it :)

bsaletti:
I'm testing the commit, the results are pretty nice.
I have a few duplicates (1-10 for 20000 records) but not on the same pod, I think it's on my side.

erikvanoosten:
Duplicated by #1132

adeibnugandiSH:
How are you

PR #1098, opened 2023-11-05
Await commit during a rebalance

### LINKED ISSUE #590: Many records duplicately processed after rebalancing

erikvanoosten:
Fixes #590 "Many records duplicately processed after rebalancing"

In this change we introduce a new mode that holds up a rebalance until all messages that were provided to the stream of a revoked partition, have been committed.

### Motivation

Here is a common (single partition) scenario around rebalances:

1. a consumer polls some messages and puts them in the streams (let's say messages with offsets 0 to 100)
1. asynchronously, the user processes these messages. Some of them are committed (let's say up to offset 50), the rest is still being processed when...
1. a rebalance happens, the partition is revoked and assigned to another consumer on another instance
1. the consumer continues to process the remaining messages with offsets 50 to 100, and tries to commit those offsets
1. _at the same time,_ another consumer on another instance, starts consuming from the last committed offset (which is 50) and will process the same messages with offsets 50 to 100

Messages with offsets 50 to 100 are being processed by both consumers simultaneously. Note that both consumers will try to commit these offsets. Until the first consumer is ready, the stored offsets can go up and down and are therefore unreliable.

After merging this change, the scenario will unfold as follows:

1. a consumer polls some messages and puts them in the streams (let's say messages with offsets 0 to 100). Zio-kafka keeps track of the highest provided offset
1. asynchronously, the user processes these messages. Some of them are committed (let's say up to offset 50), the rest is still being processed when...
1. a rebalance happens, the partition is revoked and assigned to another consumer on another instance
   * the consumer continues to process the remaining messages with offsets 50 to 100, and tries to commit those offsets
   * inside the onRevoked callback, zio-kafka continues to process commit commands from the user
   * zio-kafka continues to do so until the commit with the highest provided offset (offset 100) completes
   * the onRevoked callback completes, signalling to Kafka that the next consumer may start consuming from the partition
1. another consumer on another instance, starts consuming from the last committed offset (which is now 100, problem solved!)

### Commit queue

Because both the main runloop, and the rebalance listener need to process (and thus receive) commits commands, the commit commands were moved to a separate queue. Because the main runloop may still need to be kickstarted when it is no longer polling, a new command `CommitAvailable` was introduced.

### Complications

1. The chosen solution is not suitable for all consumers.
   - There are use cases where not all messages are read from the stream. For example, some want to read exactly 100 messages from a topic and then stop consuming. In that case the user has no intention to commit all messages, and therefore we should not wait for that to happen. Since stream consumers can basically do whatever they want, the only way we can support such use cases is by letting the consumer tell zio-kafka that they are done with committing. This requires an API change. For example, we can let the user tell zio-kafka that a given commit is the last one.
   - Not all consumers commit offsets (to Kafka) in the first place. In a future change we could make it work for commits to other stores though. As a workaround, these users can commit to both places.
   - Some users need access to the consumer from their stream. During a poll the consumer is in use and therefore not available for the stream. For some use cases the admin api, or a second consumer could function as a workaround.
1. It requires Kafka client 3.6.0. In earlier versions there was no way to wait for async commits to complete. (See [KAFKA-10337](https://issues.apache.org/jira/browse/KAFKA-10337).)

### Same thread executor

The Kafka client requires that any nested invocations (that is, from the rebalance listener callback) to the java consumer happens from the same thread. This is very much at odds with how ZIO works. Attempts to convince the Kafka committers to relax this requirement (with [KIP-944](https://cwiki.apache.org/confluence/x/chw0Dw)) failed; they could not be convinced that this is a problem. This is circumvented by using a special same-thread-runtime which runs on the thread of the caller. However, some operations such as `ZIO.timeout` and anything with `Schedules` will still shift work to another thread. We work around this by using blocking time and checking whether a `Promise` is done before awaiting it.

### Collateral

This change also:
- fixes order of `private` and `final`
- removes some completely useless tests

### Future

In the future [KIP-983](https://cwiki.apache.org/confluence/x/9Z6zDw) might become available. Once it is, we can dramatically simplify the implementation of this feature.

### Related

The same issue is present in:
- f2s-kafka: https://github.com/fd4s/fs2-kafka/issues/1200
- alpakka-kafka: https://github.com/akka/alpakka-kafka/issues/1038

In fact, every program that does polls and commits asynchronously is likely affected.

### Non-goals

This change does not try to solve the following goals. However, these can be addressed in future PRs.

- Awaiting commits after stopping the consumer, e.g. due to program shutdown (see #1087).
- Support consumers that want to commit only a portion of the given messages.
- Support transactional consumer/producer.
- Support external commits.

This branch is based on the work of abandoned PRs #788 and #830 and builds on preparatory work in PRs #744, #1068, #1073 #1086, #1089 and #1097.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
@@ -28,6 +28,7 @@ final case class ConsumerSettings(
   offsetRetrieval: OffsetRetrieval = OffsetRetrieval.Auto(),
   rebalanceListener: RebalanceListener = RebalanceListener.noop,
   restartStreamOnRebalancing: Boolean = false,
+  rebalanceSafeCommits: Boolean = false,
   fetchStrategy: FetchStrategy = QueueSizeBasedFetchStrategy()
 ) {
   private[this] def autoOffsetResetConfig: Map[String, String] = offsetRetrieval match {
@@ -154,6 +155,40 @@ final case class ConsumerSettings(
   def withRestartStreamOnRebalancing(value: Boolean): ConsumerSettings =
     copy(restartStreamOnRebalancing = value)
 
+  /**
+   * WARNING: 'rebalanceSafeCommits' is an EXPERIMENTAL feature. It is not recommended for production use yet.
+   *
+   * @param value
+   *   Whether to hold up a rebalance until all offsets of consumed messages have been committed. The default is
+   *   `false`, but the recommended value is `true` as it prevents duplicate messages.
+   *
+   * Use `false` when:
+   *   - your streams do not commit, or
+   *   - your streams require access to the consumer (the consumer is not available until the rebalance is done), or
+   *   - when it is okay to process records twice (possibly concurrently), for example, because processing is
+   *     idempotent.
+   *
+   * When `true`, messages consumed from revoked partitions must be committed before we allow the rebalance to continue.
+   *
+   * When a partition is revoked, consuming the messages will be taken over by another consumer. The other consumer will
+   * continue from the committed offset. It is therefore important that this consumer commits offsets of all consumed
+   * messages. Therefore, by holding up the rebalance until these commits are done, we ensure that the new consumer will
+   * start from the correct offset.
+   *
+   * During a rebalance no new messages can be received _for any stream_. Therefore, _all_ streams are deprived of new
+   * messages until the revoked streams are ready committing.
+   *
+   * Rebalances are held up for at most 3/5 of `maxPollInterval` (see [[withMaxPollInterval]]), by default this
+   * calculates to 3 minutes.
+   *
+   * When `false`, streams for revoked partitions may continue to run even though the rebalance is not held up. Any
+   * offset commits from these streams have a high chance of being delayed (commits are not possible during some phases
+   * of a rebalance). The consumer that takes over the partition will likely not see these delayed commits and will
+   * start from an earlier offset. The result is that some messages are processed twice and concurrently.
+   */
+  def withRebalanceSafeCommits(value: Boolean): ConsumerSettings =
+    copy(rebalanceSafeCommits = value)
+
   def withCredentials(credentialsStore: KafkaCredentialStore): ConsumerSettings =
     withProperties(credentialsStore.properties)
 
@@ -200,6 +235,6 @@ final case class ConsumerSettings(
 object ConsumerSettings {
   val defaultCommitTimeout: Duration = 15.seconds
 
-  def apply(bootstrapServers: List[String]) =
+  def apply(bootstrapServers: List[String]): ConsumerSettings =
     new ConsumerSettings().withBootstrapServers(bootstrapServers)
 }
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -123,9 +123,9 @@ private[consumer] final class Runloop private (
           ZIO.logDebug(s"Async commit of ${offsets.size} offsets for ${commits.size} commits") *>
             ZIO.attempt(consumer.commitAsync(offsets, callback)).catchAll(onFailure)
         } else {
-          // Continue to drive communication with the broker...
-          // No longer needed once KIP-848 is available.
-          ZIO.attempt(consumer.commitAsync(java.util.Collections.emptyMap(), null)).ignore
+          // Continue to drive communication with the broker so that commits can complete and the streams can
+          // make progress.
+          ZIO.attempt(consumer.commitAsync(java.util.Collections.emptyMap(), null)).orDie
         }
 
       def endingStreamsCompletedAndCommitsExist(newCommits: Chunk[Commit]): Task[Boolean] =
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -82,25 +82,16 @@ private[consumer] final class Runloop private (
     // `ZStream.repeat`, `Promise.await` on non-completed promises, and any other ZIO operation that shifts the work to
     // another thread cannot be used.
 
-    /**
-     * Maximum time spent in the rebalance callback.
-     *
-     * In this time zio-kafka awaits processing of records and the completion of commits.
-     *
-     * We use 3/5 of `maxPollInterval` which by default calculates to 3 minutes.
-     */
+    // Maximum time spent in the rebalance callback.
+    // In this time zio-kafka awaits processing of records and the completion of commits.
+    // We use 3/5 of `maxPollInterval` which by default calculates to 3 minutes.
     val maxEndingStreamsInterval = (maxPollInterval.toNanos / 5L) * 3L
 
-    /**
-     * Time between polling the commit queue from the rebalance listener when `rebalanceSafeCommits` is enabled.
-     */
+    // Time between polling the commit queue from the rebalance listener when `rebalanceSafeCommits` is enabled.
     val commitQueuePollInterval = 100.millis
 
-    /**
-     * End streams from the rebalance listener.
-     *
-     * When `rebalanceSafeCommits` is enabled, wait for consumed offsets to be committed.
-     */
+    // End streams from the rebalance listener.
+    // When `rebalanceSafeCommits` is enabled, wait for consumed offsets to be committed.
     def endStreams(state: State, streamsToEnd: Chunk[PartitionStreamControl]): Task[Unit] =
       if (streamsToEnd.isEmpty) ZIO.unit
       else {
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -123,6 +123,8 @@ private[consumer] final class Runloop private (
           ZIO.logDebug(s"Async commit of ${offsets.size} offsets for ${commits.size} commits") *>
             ZIO.attempt(consumer.commitAsync(offsets, callback)).catchAll(onFailure)
         } else {
+          // Continue to drive communication with the broker...
+          // No longer needed once KIP-848 is available.
           ZIO.attempt(consumer.commitAsync(java.util.Collections.emptyMap(), null)).ignore
         }
 
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -22,12 +22,13 @@ import scala.jdk.CollectionConverters._
 
 //noinspection SimplifyWhenInspection,SimplifyUnlessInspection
 private[consumer] final class Runloop private (
-  runtime: Runtime[Any],
+  sameThreadRuntime: Runtime[Any],
   hasGroupId: Boolean,
   consumer: ConsumerAccess,
   pollTimeout: Duration,
   maxPollInterval: Duration,
   commitTimeout: Duration,
+  commitQueue: Queue[Commit],
   commandQueue: Queue[RunloopCommand],
   lastRebalanceEvent: Ref.Synchronized[Runloop.RebalanceEvent],
   partitionsHub: Hub[Take[Throwable, PartitionAssignment]],
@@ -35,6 +36,7 @@ private[consumer] final class Runloop private (
   offsetRetrieval: OffsetRetrieval,
   userRebalanceListener: RebalanceListener,
   restartStreamsOnRebalancing: Boolean,
+  rebalanceSafeCommits: Boolean,
   currentStateRef: Ref[State],
   committedOffsetsRef: Ref[CommitOffsets],
   fetchStrategy: FetchStrategy
@@ -73,6 +75,127 @@ private[consumer] final class Runloop private (
     commandQueue.offer(RunloopCommand.RemoveSubscription(subscription)).unit
 
   private val rebalanceListener: RebalanceListener = {
+    // All code in this block is called from the rebalance listener and therefore runs on the same-thread-runtime. This
+    // is because the Java kafka client requires us to invoke the consumer from the same thread that invoked the
+    // rebalance listener.
+    // Unfortunately the same-thread-runtime does not work for all ZIO operations. For example, `ZIO.timeout`,
+    // `ZStream.repeat`, `Promise.await` on non-completed promises, and any other ZIO operation that shifts the work to
+    // another thread cannot be used.
+
+    /**
+     * Maximum time spent in the rebalance callback.
+     *
+     * In this time zio-kafka awaits processing of records and the completion of commits.
+     *
+     * We use 3/5 of `maxPollInterval` which by default calculates to 3 minutes.
+     */
+    val maxEndingStreamsInterval = (maxPollInterval.toNanos / 5L) * 3L
+
+    /**
+     * Time between polling the commit queue from the rebalance listener when `rebalanceSafeCommits` is enabled.
+     */
+    val commitQueuePollInterval = 100.millis
+
+    /**
+     * End streams from the rebalance listener.
+     *
+     * When `rebalanceSafeCommits` is enabled, wait for consumed offsets to be committed.
+     */
+    def endStreams(state: State, streamsToEnd: Chunk[PartitionStreamControl]): Task[Unit] =
+      if (streamsToEnd.isEmpty) ZIO.unit
+      else {
+        for {
+          _ <- ZIO.foreachDiscard(streamsToEnd)(_.end)
+          _ <- if (rebalanceSafeCommits) consumer.rebalanceListenerAccess(doAwaitStreamCommits(_, state, streamsToEnd))
+               else ZIO.unit
+        } yield ()
+      }
+
+    def doAwaitStreamCommits(
+      consumer: ByteArrayKafkaConsumer,
+      state: State,
+      streamsToEnd: Chunk[PartitionStreamControl]
+    ): Task[Unit] = {
+      val deadline = java.lang.System.nanoTime() + maxEndingStreamsInterval - commitTimeout.toNanos
+
+      val endingTps = streamsToEnd.map(_.tp).toSet
+
+      def commitsOfEndingStreams(commits: Chunk[Runloop.Commit]): Chunk[Runloop.Commit] =
+        commits.filter(commit => (commit.offsets.keySet intersect endingTps).nonEmpty)
+
+      lazy val previousPendingCommits: Chunk[Commit] =
+        commitsOfEndingStreams(state.pendingCommits)
+
+      def commitAsync(commits: Chunk[Commit]): UIO[Unit] =
+        if (commits.nonEmpty) {
+          val (offsets, callback, onFailure) = asyncCommitParameters(commits)
+          ZIO.logDebug(s"Async commit of ${offsets.size} offsets for ${commits.size} commits") *>
+            ZIO.attempt(consumer.commitAsync(offsets, callback)).catchAll(onFailure)
+        } else {
+          ZIO.attempt(consumer.commitAsync(java.util.Collections.emptyMap(), null)).ignore
+        }
+
+      def endingStreamsCompletedAndCommitsExist(newCommits: Chunk[Commit]): Task[Boolean] =
+        for {
+          streamResults <-
+            ZIO.foreach(streamsToEnd) { stream =>
+              for {
+                isDone    <- stream.completedPromise.isDone
+                endOffset <- if (isDone) stream.completedPromise.await else ZIO.none
+              } yield (isDone, endOffset)
+            }
+          committedOffsets <- committedOffsetsRef.get
+        } yield {
+          val allStreamsCompleted = streamResults.forall(_._1)
+          allStreamsCompleted && {
+            val endOffsets: Chunk[Offset] = streamResults.flatMap(_._2)
+            val allPendingCommits         = previousPendingCommits ++ commitsOfEndingStreams(newCommits)
+            endOffsets.forall { endOffset =>
+              val tp                    = endOffset.topicPartition
+              val offset                = endOffset.offset
+              def endOffsetWasCommitted = committedOffsets.contains(tp, offset)
+              def endOffsetCommitIsPending = allPendingCommits.exists { pendingCommit =>
+                pendingCommit.offsets.get(tp).exists { pendingOffset =>
+                  pendingOffset.offset() >= offset
+                }
+              }
+              endOffsetWasCommitted || endOffsetCommitIsPending
+            }
+          }
+        }
+
+      def commitSync: Task[Unit] =
+        ZIO.attempt(consumer.commitSync(java.util.Collections.emptyMap(), commitTimeout))
+
+      // Outline:
+      // - Every `commitQueuePollInterval` until the deadline has been reached:
+      //   - Get all commits from the commit queue.
+      //   - Start an async commit for these commits.
+      // - Collect all these new (pending) commits.
+      // - repeat the above until:
+      //   - All streams that were ended have completed their work, and
+      //   - we have seen a completed or pending commit for all end-offsets.
+      //     An end-offset of a stream is the offset of the last record given to that stream.
+      // - Do a single sync commit without any offsets, this has the side-effect of blocking until all
+      //   preceding async commits are complete (this requires kafka-client 3.6.0 or later).
+      //   Because all commits created here (including those from non-ending streams) are now complete, we do not
+      //   have to add them to the pending commits of the runloop state.
+      //
+      // Note, we cannot use ZStream.fromQueue because that will emit nothing when the queue is empty.
+      // Instead, we poll the queue in a loop.
+      ZIO.logDebug(s"Waiting for ${streamsToEnd.size} streams to end") *>
+        ZStream
+          .fromZIO(blockingSleep(commitQueuePollInterval) *> commitQueue.takeAll)
+          .tap(commitAsync)
+          .forever
+          .takeWhile(_ => java.lang.System.nanoTime() <= deadline)
+          .scan(Chunk.empty[Runloop.Commit])(_ ++ _)
+          .takeUntilZIO(endingStreamsCompletedAndCommitsExist)
+          .runDrain *>
+        commitSync *>
+        ZIO.logDebug(s"Done waiting for ${streamsToEnd.size} streams to end")
+    }
+
     // During a poll, the java kafka client might call each method of the rebalance listener 0 or 1 times.
     // We do not know the order in which the call-back methods are invoked.
     //
@@ -95,7 +218,7 @@ private[consumer] final class Runloop private (
           state          <- currentStateRef.get
           streamsToEnd = if (restartStreamsOnRebalancing && !rebalanceEvent.wasInvoked) state.assignedStreams
                          else Chunk.empty
-          _ <- ZIO.foreachDiscard(streamsToEnd)(_.end)
+          _ <- endStreams(state, streamsToEnd)
           _ <- lastRebalanceEvent.set(rebalanceEvent.onAssigned(assignedTps, endedStreams = streamsToEnd))
           _ <- ZIO.logTrace("onAssigned done")
         } yield (),
@@ -106,7 +229,7 @@ private[consumer] final class Runloop private (
           state          <- currentStateRef.get
           streamsToEnd = if (restartStreamsOnRebalancing && !rebalanceEvent.wasInvoked) state.assignedStreams
                          else state.assignedStreams.filter(control => revokedTps.contains(control.tp))
-          _ <- ZIO.foreachDiscard(streamsToEnd)(_.end)
+          _ <- endStreams(state, streamsToEnd)
           _ <- lastRebalanceEvent.set(rebalanceEvent.onRevoked(revokedTps, endedStreams = streamsToEnd))
           _ <- ZIO.logTrace("onRevoked done")
         } yield (),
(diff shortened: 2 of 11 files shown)


erikvanoosten:
@svroonland @guizmaii Perhaps we should offer this as an experimental feature at first. WDYT?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -22,24 +22,35 @@ import scala.jdk.CollectionConverters._
...
   offsetRetrieval: OffsetRetrieval,
   userRebalanceListener: RebalanceListener,
   restartStreamsOnRebalancing: Boolean,
+  rebalanceSafeCommits: Boolean,
   currentStateRef: Ref[State],
   committedOffsetsRef: Ref[CommitOffsets],
   fetchStrategy: FetchStrategy
 ) {
 
+  /**
```
svroonland, line 45: Are there any other kafka consumer or kafka broker settings that limit the acceptable values here?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -128,19 +139,124 @@ private[consumer] final class Runloop private (
...
+        ZIO.logDebug(s"Async commit of ${offsets.size} offsets for ${commits.size} commits") *>
+          ZIO.attempt(consumer.commitAsync(offsets, callback)).catchAll(onFailure)
+      } else {
+        ZIO.unit
+      }
+
+    // Note, we cannot use ZStream.fromQueue because that will emit nothing when the queue is empty.
+    // Instead, we poll the queue in a loop.
+    ZStream
+      .fromZIO(blockingSleep(100.millis) *> commitQueue.takeAll)
```
svroonland, line 204: Move it to some Runloop-private named constant or the ConsumerSettings

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -174,7 +291,7 @@ private[consumer] final class Runloop private (
       new OffsetCommitCallback {
         override def onComplete(offsets: util.Map[TopicPartition, OffsetAndMetadata], exception: Exception): Unit =
           Unsafe.unsafe { implicit u =>
-            runtime.unsafe.run {
+            sameThreadRuntime.unsafe.run {
```
svroonland, line 294: See comment below about user-land effects (which I wrote before noticing this bit): here as well a failed commit can result in executing a whole chain of ZIO effects while blocking the poll loop.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -517,12 +638,12 @@ private[consumer] final class Runloop private (
...
           val rc = RebalanceConsumer.Live(c)
           ZIO
-            .attempt(c.subscribe(pattern.pattern, rebalanceListener.toKafka(runtime, rc)))
+            .attempt(c.subscribe(pattern.pattern, rebalanceListener.toKafka(sameThreadRuntime, rc)))
             .as(Chunk.empty)
         case SubscriptionState.Subscribed(_, Subscription.Topics(topics)) =>
           val rc = RebalanceConsumer.Live(c)
           ZIO
-            .attempt(c.subscribe(topics.asJava, rebalanceListener.toKafka(runtime, rc)))
+            .attempt(c.subscribe(topics.asJava, rebalanceListener.toKafka(sameThreadRuntime, rc)))
```
svroonland, line 646: We should avoid running potential 'user-land' code on the same thread runtime, such as emitting diagnostics, which can result in a whole chain of ZIO effects being executed, while blocking the poll loop. We could explicitly fork the effects in the `emitDiagnostics` RebalanceListener or come up with some mechanism to make it explicit that we _don't_ want to fork the effects.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -128,19 +139,124 @@ private[consumer] final class Runloop private (
...
+    //   - Get all commits from the commit queue.
+    //   - Start an async commit for these commits.
+    // - Collect all these new (pending) commits.
+    // - repeat the above until:
+    //   - All streams that were ended have completed their work.
+    //   - We have seen a completed or pending commit for all endOffsets.
+    //     The end-offsets are those offsets for which commits need to be completed for a stream to be done.
+    // - Do a single sync commit without any offsets, this has the side-effect of blocking until all
+    //   preceding async commits are complete (this requires kafka-client 3.6.0 or later).
+    // - Return the new commits so that they can be added to the runloop state.
```
svroonland, line 183: Why should we do that, if they're no longer pending but completed (by the sync commit?) ?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -128,19 +139,124 @@ private[consumer] final class Runloop private (
...
+    // Note, we cannot use ZStream.fromQueue because that will emit nothing when the queue is empty.
+    // Instead, we poll the queue in a loop.
+    ZStream
+      .fromZIO(blockingSleep(100.millis) *> commitQueue.takeAll)
+      .tap(doAsyncCommits)
+      .forever
+      .takeWhile(_ => java.lang.System.nanoTime() <= deadline)
+      .scan(Chunk.empty[Runloop.Commit])(_ ++ _)
+      .takeUntilZIO { newCommits =>
+        for {
```
svroonland, line 210: I would extract a method for the for comprehension

svroonland (commented):
Nice! I don't see major issues besides that we need to carefully consider the same-thread behavior changes, a few comments/questions for now.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Because we're collecting commits that come in on the commit queue, including commits for partitions that are not ended.

svroonland: But after the sync commit they are no longer pending, right? so why do we put them in the pendingCommits in the State..?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: I don't know. I am just guessing here. I will go through the Kafka configurations and see if this is configurable in the first place. It might be a server configuration.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Yes, that is good thinking.
Re. user written rebalance listener, I suppose we should remove that feature. Unfortunately it is needed for transaction at the moment.
Re. diagnostics, we could use forking indeed.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Ah, mmm. Yeah, that is a good point. 😄

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Found it. [scheduled.rebalance.max.delay.ms](https://kafka.apache.org/documentation/#connectconfigs_scheduled.rebalance.max.delay.ms) 

> The maximum delay that is scheduled in order to wait for the return of one or more departed workers before rebalancing and reassigning their connectors and tasks to the group. During this period the connectors and tasks of the departed workers remain unassigned
> Default: 300000 (5 minutes)

erikvanoosten: Hmm maybe not. That is a Kafka connect config 😞

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Fixed in https://github.com/zio/zio-kafka/pull/1098/commits/a4fd6b5d42ad175754604d0e4642815de2ffc81c

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Fixed in https://github.com/zio/zio-kafka/pull/1098/commits/a4fd6b5d42ad175754604d0e4642815de2ffc81c

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Fixed in https://github.com/zio/zio-kafka/pull/1098/commits/a4fd6b5d42ad175754604d0e4642815de2ffc81c

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: I cannot find a setting that could configure the timeout.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Addressed in https://github.com/zio/zio-kafka/pull/1102

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -154,6 +155,32 @@ final case class ConsumerSettings(
   def withRestartStreamOnRebalancing(value: Boolean): ConsumerSettings =
     copy(restartStreamOnRebalancing = value)
 
+  /**
+   * @param value
+   *   Whether to hold up a rebalance until all offsets of consumed messages have been committed. The default is
+   *   `false`, but the recommended value is `true` as it prevents duplicate messages.
```
guizmaii, line 161: Why not `true` as the default then? 
Most people don't read documentation until something doesn't work as expected or until they experience an issue. 
I'd probably prefer to use the "safe" behaviour as the default one. 
People don't expect duplicate messages.

erikvanoosten: Unfortunately `true` is not a safe default.
If you're not committing offsets, or not committing all offsets (e.g. you're using `take(100)` on the stream), enabling this feature will make rebalances very slow (they'll be blocked until the time out of 3 minutes).

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -128,19 +144,123 @@ private[consumer] final class Runloop private (
...
+      .forever
+      .takeWhile(_ => java.lang.System.nanoTime() <= deadline)
+      .scan(Chunk.empty[Runloop.Commit])(_ ++ _)
+      .takeUntilZIO(endingStreamsCompletedAndCommitsExist)
+      .tap(_ => commitSync)
+      .runDrain
+      .ensuring {
+        ZIO.logDebug(s"Done waiting for ${streamsToEnd.size} streams to end")
+      }
+  }
```
guizmaii, line 248: This code is only used in the rebalance mechanism so IMO it should go inside the `rebalanceListener` code block:
```scala
private val rebalanceListener: RebalanceListener = {
  
  ...
   
  // somewhere here
  
  ...
}
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopCommand.scala
```diff
@@ -17,15 +16,12 @@ object RunloopCommand {
   /** Used as a signal that another poll is needed. */
   case object Poll extends Control
 
+  /** Used as a signal to the poll-loop that commits are available in the commit-queue. */
+  case object CommitAvailable extends Control
```
guizmaii, line 20: 🔴 This command is never handled in the code

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -108,7 +124,7 @@ private[consumer] final class Runloop private (
           state          <- currentStateRef.get
           streamsToEnd = if (restartStreamsOnRebalancing && !rebalanceEvent.wasInvoked) state.assignedStreams
                          else state.assignedStreams.filter(control => revokedTps.contains(control.tp))
-          _ <- ZIO.foreachDiscard(streamsToEnd)(_.end)
+          _ <- endStreams(state, streamsToEnd, awaitStreamCommits = rebalanceSafeCommits)
```
guizmaii, line 127: Why do we need to pass the `rebalanceSafeCommits` into the parameters of this function? 
This `endStreams` function already has access to the `rebalanceSafeCommits` variable

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopCommand.scala
erikvanoosten: The same is true for `Poll`. The only purpose of `Poll` and `CommitAvailable` is to kick start the runloop when it has halted.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Hmm, that may indeed no longer be needed, thanks. In the previous branch (the first working prototype) there was only a single method to end streams. In this branch I didn't try to converge them so now there are 2 (the other one being `endRevokedPartitions`). Anyways, this is where this parameter is coming from.

I'd like to investigate if we can converge the 2 stream ending methods in a future PR. For now we can remove the `awaitStreamCommits` parameter and move all this code inside the rebalance listener.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -128,19 +144,123 @@ private[consumer] final class Runloop private (
...
+        ZIO.unit
+      }
+
+    def endingStreamsCompletedAndCommitsExist(newCommits: Chunk[Commit]): Task[Boolean] =
+      for {
+        streamResults <-
+          ZIO.foreach(streamsToEnd) { stream =>
+            for {
+              isDone    <- stream.completedPromise.isDone
+              endOffset <- if (isDone) stream.completedPromise.await else ZIO.none
```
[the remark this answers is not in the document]
erikvanoosten: `PartitionStreamControl.end` puts an end token in the queue. The stream itself only ends once it has seen that end token. Perhaps we should rename the `end` method to something else to make this clearer. Suggestions welcome!

As for why this is all written this way: we need to be very careful because not all ZIO operations are allowed on the same-thread-runtime. If we abstract parts of this code, this constraint may no longer be visible. We could solve this with adding documentation (e.g. 'safe to use from same-thread-runtime'), but keeping stuff close together feels safer to me.

BTW, the same-thread-runtime really is quite annoying. For example, we cannot use `await` on a non-completed promise.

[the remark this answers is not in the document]
guizmaii: > As for why this is all written this way: we need to be very careful because not all ZIO operations are allowed on the same-thread-runtime. If we abstract parts of this code, this constraint may no longer be visible.

Ok I get it. Indeed it's better this way! Thanks

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -128,19 +144,123 @@ private[consumer] final class Runloop private (
...
+    //   preceding async commits are complete (this requires kafka-client 3.6.0 or later).
+    //   Because all commits created here (including those from non-ending streams) are now complete, we do not
+    //   have to add them to the pending commits of the runloop state.
+
+    val deadline = java.lang.System.nanoTime() + maxEndingStreamsInterval - commitTimeout.toNanos
+
+    val endingTps = streamsToEnd.map(_.tp).toSet
+    def commitsOfEndingStreams(commits: Chunk[Runloop.Commit]): Chunk[Runloop.Commit] =
+      commits.filter(commit => (commit.offsets.keySet intersect endingTps).nonEmpty)
+    val previousPendingCommits = commitsOfEndingStreams(state.pendingCommits)
```
guizmaii, line 192: This `previousPendingCommits` value is always computed while it's not always used. It could be only computed when needed

guizmaii, line 192: This `previousPendingCommits` value is always computed while it can be unused.
Can't we only compute it when we need it?

erikvanoosten: Sure! I'll make it a `def`.

erikvanoosten: It's a `lazy` val in https://github.com/zio/zio-kafka/pull/1098/commits/94375bc97ea51ec305978a28c10dda48b9143aeb

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -128,19 +144,123 @@ private[consumer] final class Runloop private (
...
+          val allPendingCommits         = previousPendingCommits ++ commitsOfEndingStreams(newCommits)
+          endOffsets.forall { endOffset =>
+            val tp     = endOffset.topicPartition
+            val offset = endOffset.offset
+            committedOffsets.contains(tp, offset) ||
+            allPendingCommits.exists { pendingCommit =>
+              pendingCommit.offsets.get(tp).exists { pendingOffset =>
+                pendingOffset.offset() >= offset
+              }
+            }
```
guizmaii: Could we make this piece of boolean logic easier to read with nicely named variables maybe?
Something like:
```scala
def offsetCommited = committedOffsets.contains(tp, offset)
// I didn't find a good name for this one
def `<aGoodName>` = allPendingCommits.exists { pendingCommit =>
              pendingCommit.offsets.get(tp).exists { pendingOffset =>
                pendingOffset.offset() >= offset
              }
            }

offsetCommited || `<aGoodName>`
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -22,24 +22,40 @@ import scala.jdk.CollectionConverters._
...
+   * In this time zio-kafka awaits processing of records and the completion of commits.
+   *
+   * We use 3/5 of `maxPollInterval` which by default calculates to 3 minutes.
+   */
+  private val maxEndingStreamsInterval = (maxPollInterval.toNanos / 5L) * 3L
+
+  /**
+   * Time between polling the commit queue from the rebalance listener when `rebalanceSafeCommits` is enabled.
+   */
+  private val commitQueuePollInterval = 100.millis
```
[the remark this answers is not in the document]
erikvanoosten: Addressed in https://github.com/zio/zio-kafka/pull/1098/commits/94375bc97ea51ec305978a28c10dda48b9143aeb.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Addressed in https://github.com/zio/zio-kafka/pull/1098/commits/94375bc97ea51ec305978a28c10dda48b9143aeb.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: If you want readability its going to be a long name 😄  I'll think about it.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
guizmaii: Why not moving it in the block of code using it?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -73,6 +75,127 @@ private[consumer] final class Runloop private (
...
+      // - Every `commitQueuePollInterval` until the deadline has been reached:
+      //   - Get all commits from the commit queue.
+      //   - Start an async commit for these commits.
+      // - Collect all these new (pending) commits.
+      // - repeat the above until:
+      //   - All streams that were ended have completed their work, and
+      //   - we have seen a completed or pending commit for all end-offsets.
+      //     An end-offset of a stream is the offset of the last record given to that stream.
+      // - Do a single sync commit without any offsets, this has the side-effect of blocking until all
+      //   preceding async commits are complete (this requires kafka-client 3.6.0 or later).
```
guizmaii, line 129: To be sure to correctly understand this comment and the code, it seems to me (I could be wrong, ofc) that the `commitSync` will be repeated until one of the conditions is satisfied but it seems to me that, according to this comment, you're not expecting this `commitSync` call to be done more than once 🤔

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -73,6 +75,127 @@ private[consumer] final class Runloop private (
...
+        .tap(commitAsync)
+        .forever
+        .takeWhile(_ => java.lang.System.nanoTime() <= deadline)
+        .scan(Chunk.empty[Runloop.Commit])(_ ++ _)
+        .takeUntilZIO(endingStreamsCompletedAndCommitsExist)
+        .tap(_ => commitSync)
+        .runDrain
+        .ensuring {
+          ZIO.logDebug(s"Done waiting for ${streamsToEnd.size} streams to end")
+        }
```
guizmaii, line 196: Related to this remark: https://github.com/zio/zio-kafka/pull/1098/files#r1386974834
If you're expecting only 1 `commitSync` call, I'd write this code this way:
```scala
    ZStream
      .fromZIO(blockingSleep(commitQueuePollInterval) *> commitQueue.takeAll)
      .tap(commitAsync)
      .scan(Chunk.empty[Runloop.Commit])(_ ++ _)
      .takeUntilZIO(commits =>
        ZIO.succeed(java.lang.System.nanoTime() > deadline) || endingStreamsCompletedAndCommitsExist(commits)
      )
      .forever
      .runDrain *>
      commitSync *>
      ZIO.logDebug(s"Done waiting for ${streamsToEnd.size} streams to end")
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Because then it is recalculated every 100ms.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Addressed in https://github.com/zio/zio-kafka/pull/1098/commits/d1a961f4bf55832f56de178e6ca8b9b9104b7495:

```scala
val allStreamsCompleted = streamResults.forall(_._1)
allStreamsCompleted && {
  val endOffsets: Chunk[Offset] = streamResults.flatMap(_._2)
  val allPendingCommits         = previousPendingCommits ++ commitsOfEndingStreams(newCommits)
  endOffsets.forall { endOffset =>
    val tp                    = endOffset.topicPartition
    val offset                = endOffset.offset
    def endOffsetWasCommitted = committedOffsets.contains(tp, offset)
    def endOffsetCommitIsPending = allPendingCommits.exists { pendingCommit =>
      pendingCommit.offsets.get(tp).exists { pendingOffset =>
        pendingOffset.offset() >= offset
      }
    }
    endOffsetWasCommitted || endOffsetCommitIsPending
  }
}

```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Your code is correct! It is like that now in:
https://github.com/zio/zio-kafka/pull/1098/commits/36225bca90ff5ffcea4a84e8c2cb2268bc6284c6

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: The parameter is gone in https://github.com/zio/zio-kafka/pull/1098/commits/d1a961f4bf55832f56de178e6ca8b9b9104b7495.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Fixed now.

erikvanoosten:
> One thing I want to carefully review is off-by-one errors in the offsets.

This is now done and indeed there was a bug which is now fixed.

I have marked the new mode as experimental.

@guizmaii @svroonland When the test complete (in ~20 minutes), this PR is ready for merging.

erikvanoosten:
Haha, I cheered too early. Working on it... 😄

erikvanoosten:
> It seems we have a new flaky test:
> 
> ```
>     - Consumer Streaming - does not process messages twice for transactional producer, even when rebalancing - org.apache.kafka.clients.consumer.RangeAssignor
> Timeout of 2 m exceeded.
> ```

This has been fixed.

erikvanoosten:
Okay, _now_ this PR is ready for merging. The tests have run several time without issues.

There is one outstanding comment (about configuration parameters) but I cannot answer that question (I did try). By marking this feature as experimental we can collect feedback about this (and other things) from the adventurous users and not disappoint the users that need more stability.

@svroonland @guizmaii Please review again.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -73,6 +75,120 @@ private[consumer] final class Runloop private (
     commandQueue.offer(RunloopCommand.RemoveSubscription(subscription)).unit
 
   private val rebalanceListener: RebalanceListener = {
```
svroonland, line 77: Maybe we should extract a class to a separate file here

svroonland: Yeah, separate PR is fine

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Yes, you are entirely correct; #1102 solves the problem for diagnostics.
Besides diagnostics, do you see another way in which a failed commit results in a chain of ZIO effects?

svroonland: With code like `offset.commit.catchAll(e => callSomeSlowExternalApi)`

svroonland: It depends on what fiber the awaiting of a Promise (which is what `offset.commit` does internally) continues. I'm not sure whether it is the fiber that completes the promise or the original awaiting fiber..

erikvanoosten: My guess would be that even though the continuation might be executed on a different fiber, it would not be executed on a different executor.

erikvanoosten: I said that wrong. I think the continuation will continue on the same fiber and on the same executor.

Anyways, to be sure I asked [the question on Discord](https://discord.com/channels/629491597070827530/630498701860929559/1176863600040738896).

erikvanoosten: It appears I was correct. So this is not a problem.

svroonland: Cool! Then I have no more objections :)

svroonland (approved):

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: @svroonland I tried to move the rebalance listener to a separate file. Unfortunately they are so integrated that it would require a lot of refactoring. I don't think it is feasible.

PR #1102, opened 2023-11-08
Protect agains user diagnostics, better rebalance events

erikvanoosten:
Diagnostics is a feature of zio-kafka that allows users to listen to key events. Since zio-kafka calls out to the user's implementation of the `Diagnostics` trait, there are no guarantees on how well it behaves.

This is even more important inside the rebalance listener where we (soon, with #1098) run on the same-thread-runtime and can not afford to be switched to another thread by ZIO operations that are normally safe to use.

To protect against these issues the user's diagnostics implementation is run on a separate fiber, feeding from a queue of events.

In addition, the rebalance events are replaced by a single event which is emitted from outside the rebalance listener. The new event gives the full picture of a rebalance, including which streams were ended. Previously it was not clear which rebalance events belonged to the same rebalance.

**Breaking change**

Since the rebalance events are changed, this is a breaking change.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/diagnostics/Diagnostics.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/diagnostics/Diagnostics.scala
@@ -2,6 +2,7 @@ package zio.kafka.consumer.diagnostics
 
 import zio.stream.ZStream
 import zio._
+import zio.kafka.consumer.diagnostics.DiagnosticEvent.Finalization.ConsumerFinalized
 
 trait Diagnostics {
   def emit(event: => DiagnosticEvent): UIO[Unit]
@@ -30,16 +31,13 @@ object Diagnostics {
     def make(wrapped: Diagnostics): ZIO[Scope, Nothing, Diagnostics] =
       if (wrapped == Diagnostics.NoOp) ZIO.succeed(Diagnostics.NoOp)
       else {
-        // Run the diagnostics from a separate fiber. The fiber is interrupted when it tries to poll from the queue
-        // while that queue is shut down. To give the fiber a chance to get the last item in the queue, we delay the
-        // queue shut down by 10ms.
         for {
-          queue <- ZIO.acquireRelease(Queue.unbounded[DiagnosticEvent])(_.shutdown.delay(10.millis))
-          _     <- ZStream.fromQueue(queue).tap(wrapped.emit(_)).runDrain.forkDaemon
+          queue <- ZIO.acquireRelease(Queue.unbounded[DiagnosticEvent])(_.shutdown)
+          fib   <- ZStream.fromQueue(queue).tap(wrapped.emit(_)).takeUntil(_ == ConsumerFinalized).runDrain.forkScoped
+          _     <- ZIO.addFinalizer(queue.offer(ConsumerFinalized) *> fib.await)
         } yield new Diagnostics {
           override def emit(event: => DiagnosticEvent): UIO[Unit] = queue.offer(event).unit
         }
       }
   }
-
 }
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/diagnostics/Diagnostics.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/diagnostics/Diagnostics.scala
@@ -1,5 +1,6 @@
 package zio.kafka.consumer.diagnostics
 
+import zio.stream.ZStream
 import zio.{ Queue, Scope, UIO, ZIO }
 
 trait Diagnostics {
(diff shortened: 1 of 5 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -134,7 +131,7 @@ private[consumer] final class Runloop private (
       for {
         p <- Promise.make[Throwable, Unit]
         _ <- commandQueue.offer(RunloopCommand.Commit(offsets, p)).unit
-        _ <- diagnostics.emit(DiagnosticEvent.Commit.Started(offsets))
+        _ <- diagnostics.emit(DiagnosticEvent.Commit.Started(offsets)).forkDaemon
```
svroonland, line 134: In my experience, with daemon fibers it's easy to have errors go unnoticed. What if instead we put the on a queue and have a forked fiber in the Runloop that reads from the queue and emits them.

erikvanoosten: After creating this PR I realized that forking has the problem that diagnostics might be received in the wrong order. An intermediate queue would solve that as well.

But in your proposal there is still a forked fiber, why would error be noticed there?

svroonland: Come to think of it, Diagnostics.emit is `UIO[Unit]` so it  cannot fail. But you are right about the ordering, which speaks for having a queue.

erikvanoosten:
> > there are no guarantees on how well it behaves.
> 
> What about hiding the diagnostics from the users? What's the purpose of exposing it? Shouldn't it be an internal thing?

Well, we have had some feature requests around observability that are covered by diagnostics. But I don't know the original rationale for introducing this.

erikvanoosten:
I just pushed a new implementation that first emits to a queue and then runs the user's Diagnostics on a separate thread.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Addressed in https://github.com/zio/zio-kafka/pull/1102/commits/8a274146cc452565a49f01acb2c97ec784d769e4

svroonland:
> Well, we have had some feature requests around observability that are covered by diagnostics. But I don't know the original rationale for introducing this.

Observability indeed. Though it could probably use some more attention at some point.

zio-kafka/src/main/scala/zio/kafka/consumer/diagnostics/Diagnostics.scala
```diff
@@ -10,11 +11,29 @@ object Diagnostics {
...
   }
+
+  object QueuedDiagnostics {
+
+    /**
+     * @return
+     *   a new `Diagnostics` that emits the events to a queue first and runs the wrapped `Diagnostics` in a separate
+     *   fiber
+     */
+    def make(wrapped: Diagnostics, queueSize: Int = 16): ZIO[Scope, Nothing, Diagnostics] =
```
svroonland, line 29: Nice. I wonder about the queue size. We want to avoid blocking the runloop as far as is reasonably possible. Perhaps it should just be unbounded..

guizmaii:
> I just pushed a new implementation that first emits to a queue and then runs the user's Diagnostics on a separate thread.

If we use the approach, do we still need to change the emitted events?

erikvanoosten:
> > I just pushed a new implementation that first emits to a queue and then runs the user's Diagnostics on a separate thread.
> 
> If we use the approach, do we still need to modify the emitted events?

No we don't.

However, IMHO the new event is more useful then the old events. They contain more data (endedStreams) but also there is only event per rebalance. Before you could get multiple events without knowledge which happened in the same rebalance.

erikvanoosten:
And another flaky test:
```
      - Consumer Streaming / issue #856 / Ordering of finalizers matters. If subscriptions are finalized after the runloop, it creates a deadlock / When not consuming, the Runloop is not started so only the Consumer is finalized
        ✗ Chunk() did not have the same elements as  Chunk(ConsumerFinalized())
        finalizationEvents did not satisfy hasSameElements(Chunk(ConsumerFinalized))
        finalizationEvents = Chunk()
        at /home/runner/work/zio-kafka/zio-kafka/zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala:1199 
```

erikvanoosten:
Some tests depend on diagnostics. They will have to adapted...

zio-kafka/src/main/scala/zio/kafka/consumer/diagnostics/Diagnostics.scala
erikvanoosten: Done in https://github.com/zio/zio-kafka/pull/1102/commits/3659b85bdfb7253d97b05577af6bdcb7a0429ba7.

erikvanoosten:
> Some tests depend on diagnostics. They will have to adapted...

Done.

zio-kafka/src/main/scala/zio/kafka/consumer/diagnostics/Diagnostics.scala
```diff
@@ -10,11 +11,35 @@ object Diagnostics {
...
+     */
+    def make(wrapped: Diagnostics): ZIO[Scope, Nothing, Diagnostics] =
+      if (wrapped == Diagnostics.NoOp) ZIO.succeed(Diagnostics.NoOp)
+      else {
+        // Run the diagnostics from a separate fiber. The fiber is interrupted when it tries to poll from the queue
+        // while that queue is shut down. To give the fiber a chance to get the last item in the queue, we delay the
+        // queue shut down by 10ms.
+        for {
+          queue <- ZIO.acquireRelease(Queue.unbounded[DiagnosticEvent])(_.shutdown.delay(10.millis))
+          _     <- ZStream.fromQueue(queue).tap(wrapped.emit(_)).runDrain.forkDaemon
```
svroonland, line 38: Why not forkScoped here so this fiber gets shut down before the closing of the queue?

erikvanoosten: When you do that, the tests start to fail because you will always miss the last event.

svroonland (approved):

PR #1108, opened 2023-11-15
add consumeVia method

mberndt123:
`consumeWith` _almost_ does what I need it to, but is unfortunately too restrictive. I'd therefore like to add a slightly generalized variant that works with a `ZPipeline`.
Specifically, in my use case I need to do some amount of work sequentially, while other parts are required to run concurrently for performance, and this new method allows me to do this kind of processing.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
@@ -137,6 +137,18 @@ trait Consumer {
   ): ZIO[R & R1, Throwable, Unit]
 
   /**
+   * See [[Consumer.consumeVia]].
+   */
+  def consumeVia[R: EnvironmentTag, R1: EnvironmentTag, K, V](
+    subscription: Subscription,
+    keyDeserializer: Deserializer[R, K],
+    valueDeserializer: Deserializer[R, V],
+    commitRetryPolicy: Schedule[Any, Any, Any] = Schedule.exponential(1.second) && Schedule.recurs(3)
+  )(
+    f: ZPipeline[R1, Nothing, CommittableRecord[K, V], Offset]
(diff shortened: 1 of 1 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -517,12 +545,24 @@ private[consumer] final class ConsumerLive private[consumer] (
     commitRetryPolicy: Schedule[Any, Any, Any] = Schedule.exponential(1.second) && Schedule.recurs(3)
   )(
     f: ConsumerRecord[K, V] => URIO[R1, Unit]
+  ): ZIO[R & R1, Throwable, Unit] =
+    consumeVia(subscription, keyDeserializer, valueDeserializer, commitRetryPolicy)(
+      ZPipeline.mapZIO(r => f(r.record).as(r.offset))
```
paulpdaniels, line 550: I think you want to preserve the `.mapChunkZIO` usage from the variant you removed. `mapZIO` will destroy the chunking and tank throughput.

mberndt123: Good point, thanks.

erikvanoosten:
From Discord:

> Hey there, 
> 
> Sorry for the late reply.
> 
> I kind of agree with @vroste. I think we should offer a minimalistic API otherwise we'll end up having to provide these 10+ variants.
> Also, seeing your code comments, it seems that your function can be easily misused which isn't something we should offer IMO.
> Another issue I see is that ZPipeline is a low-level ZIO API. I'm not sure we should encourage people using it. 
> Staying at the ZStream level is enough for 95% of users.
> 
> Sorry @Matthias Berndt but I don't think that your function should be part of zio-kafka API. 
> You should probably have this in your code as an extension function.

PR #1109, opened 2023-11-16
Make access to the java consumer fair

### LINKED ISSUE #1107: PartitionnedStream hanging after stream consumption started
Hi, I notice a bug in `v2.6.0` that is not in `v2.3.4`.

I am using a `Consumer.partitionedStream` and for each partition I need to get the committed and endOffset with the `Consumer.endOffsets` / `Consumer.committed` method: 

```
Consumer.partitionedStream(???).mapZIOParUnordered(12)(
        (e: (TopicPartition, stream.Stream[Throwable, CommittableRecord[Bytes, Either[UnknownEvent, Event]]])) => {
          val (topicPartition, streamIn) = e
          Consumer.endOffsets(Set(topicPartition)).flatMap {
            ZIO.debug("some more exec here") *> streamIn.runDrain
          }
        }
      )
```
I have 12 partitions so I use `mapZIOParUnordered` with a parallelism of 12 to process all my partitions.

I notice that as soon as the first stream starts for one partition, then the Consumer.endOffsets and Consumer.committed are really long for others partitions and too long. It looks like the polling of the consumer blocks other threads to achieve Consumer.endOffsets and Consumer.committed tasks. The result after few minutes is that all the partitions are not assigned anymore forever but the process is not kill.

flavienbert:
Replace the semaphore that is used to control access to the java consumer with a reentrant lock that has fairness enabled. This prevents liveness problems for consumers that need to use the consumer, e.g. to fetch committed offsets. Fixes #1107.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -100,7 +100,8 @@ lazy val zioKafka =
     .settings(
       libraryDependencies ++= Seq(
         kafkaClients,
-        scalaCollectionCompat
+        scalaCollectionCompat,
+        "dev.zio" %% "zio-concurrent" % zioVersion.value
       )
     )
 
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/ConsumerAccess.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/ConsumerAccess.scala
@@ -19,7 +19,7 @@ private[consumer] final class ConsumerAccess(
     withConsumerZIO[Any, A](c => ZIO.attempt(f(c)))
 
   def withConsumerZIO[R, A](f: ByteArrayKafkaConsumer => RIO[R, A]): RIO[R, A] =
-    access.lock *> withConsumerNoPermit(f) <* access.unlock
+    access.lock.zipRight(withConsumerNoPermit(f)).ensuring(access.unlock)
 
   private[consumer] def withConsumerNoPermit[R, A](
     f: ByteArrayKafkaConsumer => RIO[R, A]
@@ -36,7 +36,7 @@ private[consumer] final class ConsumerAccess(
    * Do not use this method outside of the Runloop
    */
   private[internal] def runloopAccess[R, E, A](f: ByteArrayKafkaConsumer => ZIO[R, E, A]): ZIO[R, E, A] =
-    access.lock *> f(consumer) <* access.unlock
+    access.lock.zipRight(f(consumer)).ensuring(access.unlock)
 }
 
 private[consumer] object ConsumerAccess {
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/ConsumerAccess.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/ConsumerAccess.scala
@@ -8,16 +8,18 @@ import zio.kafka.consumer.ConsumerSettings
 import zio.kafka.consumer.internal.ConsumerAccess.ByteArrayKafkaConsumer
 
 import scala.jdk.CollectionConverters._
+import zio.concurrent.ReentrantLock
 
 private[consumer] final class ConsumerAccess(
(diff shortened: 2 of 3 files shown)


flavienbert:
Fixes #1107

svroonland:
Have you verified that this fixes the issue? A unit test would of course be very welcome

zio-kafka/src/main/scala/zio/kafka/consumer/internal/ConsumerAccess.scala
```diff
@@ -8,16 +8,18 @@ import zio.kafka.consumer.ConsumerSettings
...
-  access: Semaphore
+  access: ReentrantLock
 ) {
+
   def withConsumer[A](f: ByteArrayKafkaConsumer => A): Task[A] =
     withConsumerZIO[Any, A](c => ZIO.attempt(f(c)))
 
   def withConsumerZIO[R, A](f: ByteArrayKafkaConsumer => RIO[R, A]): RIO[R, A] =
-    access.withPermit(withConsumerNoPermit(f))
+    access.lock *> withConsumerNoPermit(f) <* access.unlock
```
erikvanoosten, line 22: Shouldn't that be:
```suggestion
    access.lock.zipRight.withConsumerNoPermit(f).ensuring(access.unlock)
```

erikvanoosten:
I like it!

So now the question is, how to unit test this? Perhaps by calling to the consumer from the stream (e.g. get offsets), and then verifying that it makes progress?

erikvanoosten:
> Probably the java consumer is not used in time during shutdow process?

Likely.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/ConsumerAccess.scala
erikvanoosten: Or perhaps this is better:
```suggestion
    (access.lock *> withConsumerNoPermit(f)).ensuring(access.unlock)
```

flavienbert:
I tried to run the test I wrote with the semaphore version, and the test fail.

erikvanoosten (approved):

erikvanoosten:
Thanks @flavienbert !

flavienbert:
@erikvanoosten you're welcome. Do you plan to publish a new version with this fix soon?

erikvanoosten:
Yes definitely! Probably the coming weekend.

build.sbt
```diff
@@ -100,7 +100,8 @@ lazy val zioKafka =
     .settings(
       libraryDependencies ++= Seq(
         kafkaClients,
-        scalaCollectionCompat
+        scalaCollectionCompat,
+        "dev.zio" %% "zio-concurrent" % zioVersion.value
```
guizmaii, line 104: Why do we need to add this dependency? zio-concurrent's repo has been archived: https://github.com/zio/zio-concurrent. I'm not sure it's a good idea to be dependent on it. 🤔

erikvanoosten:
> @flavienbert @erikvanoosten Seeing this PR from ZIO (merged but not yet released): [zio/zio#8554](https://github.com/zio/zio/pull/8554) and seeing that zio-concurrent repo is archived, do we really want/need to use zio-concurrent `ReentrantLock`?
> 
> Edit: I checked and `ReentrantLock` is directly present in ZIO. We don't need the zio-concurrent dependency

Zio-concurrent is now a module in the main zio repository.
Eh, I do not see it in ZIO itself... did I miss something?

guizmaii:
@erikvanoosten My bad. I though it was coming from the zio-concurrent repo but comes indeed from ZIO's main repo

erikvanoosten:
Now that the standard semaphore in zio is fair (since 2.0.20, see https://github.com/zio/zio/pull/8554), we can revert this change, allowing us to remove an external dependency.

PR #1118, opened 2023-11-27
Make maxRebalanceDuration configurable

erikvanoosten:


zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -29,6 +29,7 @@ final case class ConsumerSettings(
   rebalanceListener: RebalanceListener = RebalanceListener.noop,
   restartStreamOnRebalancing: Boolean = false,
   rebalanceSafeCommits: Boolean = false,
+  maxRebalanceDuration: Option[Duration] = None,
```
svroonland, line 32: Why not have the default duration here?

erikvanoosten: Because the default depends on `maxPollInterval` for which the default is fetched in `RunloopAccess`.

There are other options:
- store a function in ConsumerSettings (maxPollInterval => maxBalanceDuration)
- ~fetch the default `maxPollInterval` in ConsumerSettings~ doesn't work, it depends on the properties

svroonland (approved):

PR #1139, opened 2023-12-17
Make tests faster

erikvanoosten:
Make tests faster

Make tests faster by
 - running them in parallel
 - reduce number of runs with `nonFlaky`

Results for `sbt test`:
 - on GHA: from 17m 34s to 7m
 - on my laptop: from 231s to 87s

Also: upgrade logback from 1.3.14 to 1.4.14.

### DIFF
--- a/zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
+++ b/zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
@@ -34,19 +34,31 @@ object KafkaTestUtils {
     )
 
   /**
-   * Default transactional Producer settings you can use in your tests.
+   * Default transactional producer settings you can use in your tests.
+   *
+   * Note: to run multiple tests in parallel, you need to use different transactional ids via
+   * `transactionalProducerSettings(transactionalId)`.
    */
   val transactionalProducerSettings: ZIO[Kafka, Nothing, TransactionalProducerSettings] =
+    transactionalProducerSettings("test-transaction")
+
+  def transactionalProducerSettings(transactionalId: String): ZIO[Kafka, Nothing, TransactionalProducerSettings] =
     ZIO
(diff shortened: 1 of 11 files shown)


zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
```diff
@@ -36,17 +36,17 @@ object KafkaTestUtils {
   /**
    * Default transactional Producer settings you can use in your tests.
    */
-  val transactionalProducerSettings: ZIO[Kafka, Nothing, TransactionalProducerSettings] =
+  def transactionalProducerSettings(transactionalId: String): ZIO[Kafka, Nothing, TransactionalProducerSettings] =
```
svroonland, line 39: Is this backwards-incompatible? Should we add a default?

erikvanoosten: No, its not backwards compatible. Adding a default value is a good idea, at least it would make this source compatible (which I guess is already great for a test-kit).

erikvanoosten: Actually, we add additional methods and keep the `val`s and stay completely backward compatible. Commit coming soon!

erikvanoosten: Fixed in https://github.com/zio/zio-kafka/pull/1139/commits/ef19ec29328d0f99b1c8417cadd14c5c10d9e022.

erikvanoosten:
Sorry for the force push. I squashed all the commits to make rebasing easier.

erikvanoosten:
Instead of removing the `nonFlaky` annotations, in the last commit only the number of retries is reduced.

svroonland (approved):

Issue #1126, opened 2023-11-29
Should commitOrRetry also retry commit timeouts?


fmeeuw:
Hi, 

The Offset.commitOrRetry method currently is implemented to only retry `RetriableCommitFailedException` errors, like this:
```
    commit.retry(
      Schedule.recurWhile[Throwable] {
        case _: RetriableCommitFailedException => true
        case _                                 => false
      } && policy
 ```

I was expecting a `zio.kafka.consumer.CommitTimeout` to be also retried, but that currently isn't the case. 
There is of course a simple workaround, to add another .retry after the .commitOrRetry. 
However I think it would be nice if commitOrRetry also retries commit timeouts, what do you think?

erikvanoosten:
Solved by #1140.

PR #1140, opened 2023-12-19
retry Consumer.CommitTimeout exception

myazinn:
Though this is a debatable change, I think it's not obious that `Consumer.CommitTimeout` exception is not retrieable when using `Offset#commitOrRetry` or `OffsetBatch#commitOrRetry`.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/Offset.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/Offset.scala
@@ -35,6 +35,7 @@ object Offset {
     commit.retry(
       Schedule.recurWhile[Throwable] {
         case _: RetriableCommitFailedException => true
+        case Consumer.CommitTimeout            => true
         case _                                 => false
       } && policy
     )


erikvanoosten:
The purpose of a time out is to abort an operation that is taking too long and is therefor unlikely to succeed. Sure, you can retry this, but it should never be tried indefinitely, `RetriableCommitFailedException` exceptions can be retried indefinitely.

In short, zio-kafka cannot decide how often a commit timeout should be retried. Therefore, we have to delegate this to the user. The user can either increase the timeout, or do a retry in their code.

erikvanoosten:
Okay, I am re-reading the code. And now I see that there is also the user provided policy... Have to think more about this 🙂

erikvanoosten:
@myazinn Can you tell a bit about what motivated you to write this PR?

svroonland:
See also #1126

myazinn:
> I am afraid we cannot accept it.

If that's the case, no worries, we'll add it on our side :) I just thought it was overlooked

> Can you tell a bit about what motivated you to write this PR?

One of our services has been failing from time to time (once or twice in a month) with timeouts after we updated `zio-kafka` several months ago. It's unclear why it fails since a lot of others work fine, but usually it timeouts a few times and then works fine. I was surprised too see that even though we use `commitOrRetry`, it doesn't retry anything and just restarts

erikvanoosten (approved):

erikvanoosten:
My earlier reasoning didn't make sense. Thanks for your PR @myazinn.

PR #1141, opened 2023-12-22
Delete `RunloopTimeout` and small improvements

erikvanoosten:
* Delete `RunloopTimeout` as it is no longer used.
* Do not suppress error in `commitTransactionWithOffsets` (even though the only invoker still dies upon any error).
* Better debug logging in `Runloop`.
* Shave tens of seconds of test by never taking more messages than can be sent.
* Run test less often, 3 times is enough (`nonFlaky(3)`` runs the test 4 times!).


svroonland (approved):
Nice!

Issue #941, opened 2023-06-21
[Idea] pause/resume/stop Stream and remove `stopConsumption`


guizmaii:
We had some discussion yesterday on Discord about removing `stopConsumption`
See:
- Starting here: https://discord.com/channels/629491597070827530/629497941719121960/1120614848015900744
- "interruptor" idea: https://discord.com/channels/629491597070827530/629497941719121960/1120649014350708748

Found a new idea this morning while taking my shower.
I think we could "generalize" the "interruptor" idea.
We could return a "control" interface that'd have 3 methods:
- pause: would pause the stream
- resume: would resume the paused stream
- stop: would stop the stream and the subscription

```scala
// pseudo-code just to give show the idea
(
  val (control, stream0) = 
    Consumer
      .plainStream(Sub.topic("topic-a"), ...)
      .mapZIO(processRecord0(_))
      .runDrain

  control.pause() // pauses stream0
  control.resume() // resume stream0
  control.stop() // stop/finalize stream0
).provide(ZLayer.scoped(Consumer.make(settings)))
```

The "Control" interface would be something like:
```scala
trait Control {
  def pause: IO[AlreadyStopped, Boolean] // Boolean as we probably want to give a feedback to the user
  def resume: IO[AlreadyStopped, Boolean] 
  def stop: UIO[AlreadyStopped, Boolean]
}
```

erikvanoosten:
I have never used it, but I see that github offers [Discussions](https://github.com/zio/zio-kafka/discussions). Perhaps we can move this there?

erikvanoosten:
I guess it follows logically from our discussions because I had the almost the exact same idea. At least you put it in writing!
Important is that when the stream is paused, that commits are still processed. We need similar capacities (per partition instead of per subscription) to control rebalances.

guizmaii:
> when the stream is paused, that commits are still processed

Yes, that's the idea.
Also, the `stop` would probably need to finish committing everything scheduled (in the commit queue) before actually stopping, no?

erikvanoosten:
> > We need similar capacities (per partition instead of per subscription) to control rebalances.
> 
> Not sure to see what you mean. Can you give an example maybe, please?

This is the reason I started contributing to zio-kafka; during a rebalance we need to stop polling for a stream but let its commits complete (and wait for commit completion). That sounds pretty similar, right?
Right now, during a rebalance we stop polling, but we don't wait for the commits to complete.

guizmaii:
[...]

PR #1201, opened 2024-03-24
Graceful shutdown of a stream for a single subscription

svroonland:
Implements functionality for gracefully stopping a stream for a single subscription: stop fetching records for the assigned topic-partitions but keep being subscribed so that offsets can still be committed. Intended to replace `stopConsumption`, which did not support multiple-subscription use cases. 

A new command `EndStreamsBySubscription` is introduced, which calls the `end` method on the `PartitionStreamControl` of streams matching a subscription. In the method `Consumer#runWithGracefulShutdown` we then wait for the user's stream to complete, before removing the subscription.

This is experimental functionality, intended to replace `stopConsumption` at some point. Methods with this new functionality are offered besides existing methods to maintain compatibility.

~All the fiber and scope trickery proved to be very hard to get right (the lifetime of this PR is a testimony to that), and there may still be subtle issues here.~ This is now traced back to issue https://github.com/zio/zio/issues/9288

Implements some of #941.


docs/consuming-kafka-topics-using-zio-streams.md
```diff
@@ -62,3 +62,52 @@ Consumer.partitionedStream(Subscription.topics("topic150"), Serde.string, Serde.
   .mapZIO(_.commit)
   .runDrain
 ```
+
+## Controlled shutdown
+
+The examples above will keep processing records forever, or until the fiber is interrupted, typically at application shutdown. When interrupted, in-flight records will not be processed fully through all stream stages and offsets may not be committed. For fast shutdown in an at-least-once processing scenario this is fine. zio-kafka also supports a _graceful shutdown_, where the fetching of records for the subscribed topics/partitions is stopped, the streams are ended and all downstream stages are completed, allowing all in-flight messages to be fully processed.
```
erikvanoosten, line 68: > allowing all in-flight messages to be fully processed.

I suggest we change these words to:

> allowing commits for in-flight records to complete.

1. Let's continue to use the word 'records'.
2. Removed the word 'all' because when an interrupt is received, internally queued records are dropped and _not_ passed to the consumer.
3. The remaining in-flight records are already passed to the consumer so processing already commenced. The only relevant operation from zio-kafka point of view is that the records can be committed.

docs/consuming-kafka-topics-using-zio-streams.md
```diff
@@ -62,3 +62,52 @@ Consumer.partitionedStream(Subscription.topics("topic150"), Serde.string, Serde.
...
+
+Use the `*withControl` variants of `plainStream`, `partitionedStream` and `partitionedAssignmentStream` for this purpose. These return a ZIO producing a `SubscriptionStreamControl` and requiring a `Scope`. When provided with a `Scope`, eg using `ZIO.scoped { .. }`, the consumer remains subscribed to the given topics for the lifetime of the `Scope`. The graceful shutdown may be initiated by calling `SubscriptionStreamControl#stop`.
```
erikvanoosten, line 70:
```suggestion
Use the `*withControl` variants of `plainStream`, `partitionedStream` and `partitionedAssignmentStream` for this purpose. These return a ZIO producing a `SubscriptionStreamControl` and requiring a `Scope`. When provided with a `Scope`, e.g. using `ZIO.scoped { .. }`, the consumer remains subscribed to the given topics for the lifetime of the `Scope`. Commits are possible for as long as the consumer is subscribed. The graceful shutdown may be initiated by calling `SubscriptionStreamControl#stop`.
```

It is not clear here what `stop` does. I guess it will stop consuming new records. Is it save to close the scope when `stop` completes?

docs/consuming-kafka-topics-using-zio-streams.md
```diff
@@ -62,3 +62,52 @@ Consumer.partitionedStream(Subscription.topics("topic150"), Serde.string, Serde.
...
+
+```scala
+import zio.Console.printLine
+import zio.kafka.consumer._
+
+ZIO.scoped {
+  for {
+    streamControl <- Consumer.partitionedStreamWithControl(Subscription.topics("topic150"), Serde.string, Serde.string)
```
erikvanoosten, line 78: Could this be layed out for smaller screens please?

docs/consuming-kafka-topics-using-zio-streams.md
```diff
@@ -62,3 +62,52 @@ Consumer.partitionedStream(Subscription.topics("topic150"), Serde.string, Serde.
...
+}
+```
+
+In most use cases you don't need to explicitly control when to shutdown the stream, but you simply want a graceful shutdown on interruption when the application terminates. The `Consumer#runWithGracefulShutdown` method implements this pattern and can be used as follows:
+
+```scala
+import zio.Console.printLine
+import zio.kafka.consumer._
+
+ZIO.scoped {
```
erikvanoosten, line 101: Would it be possible/desirable to create the scope inside `runWithGracefulShutdown` so that the user doesn't need to define it?

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
```diff
@@ -355,6 +355,226 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
...
       } @@ TestAspect.nonFlaky(2),
+      suite("streamWithControl")(
+        test("stop must end streams while still processing commits") {
+          for {
+            topic  <- randomTopic
+            group  <- randomGroup
+            client <- randomClient
+
+            keepProducing <- Ref.make(true)
+            _             <- produceOne(topic, "key", "value").repeatWhileZIO(_ => keepProducing.get).fork
```
erikvanoosten, line 366: You can also use test helper method `scheduledProduce`.

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
```diff
@@ -355,6 +355,226 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
...
+                            }
+                            .runDrain
+                            .tap(_ => ZIO.logDebug("Stream completed"))
+                   } yield ()
+                 }
+                   .provideSomeLayer[Kafka](
+                     consumer(client, Some(group))
+                   )
+            _ <- keepProducing.set(false)
+          } yield assertCompletes
```
erikvanoosten, line 387: Shouldn't we assert that all commits complete?

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
```diff
@@ -355,6 +355,226 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
...
+                            Consumer
+                              .runWithGracefulShutdown(
+                                Consumer.plainStreamWithControl(Subscription.topics(topic), Serde.string, Serde.string)
+                              ) { stream =>
+                                stream.mapConcatZIO { record =>
+                                  for {
+                                    nr <- messagesReceived.updateAndGet(_ + 1)
+                                    _  <- stop.succeed(()).when(nr == 10)
+                                  } yield if (nr < 10) Seq(record.offset) else Seq.empty
+                                }
```
erikvanoosten, line 410: Why not `.take(10)`?

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
```diff
@@ -355,6 +355,226 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
...
+                                  .transduce(Consumer.offsetBatches)
+                                  .mapZIO(_.commit)
+                                  .runDrain
+                              }
+                              .forkScoped
+                          _      <- stop.await *> fib.interrupt
+                          offset <- Consumer.committed(Set(new TopicPartition(topic, 0))).map(_.values.head)
+                        } yield offset
+                      }.provideSomeLayer[Kafka](consumer(client, Some(group)))
+          } yield assert(offset.map(_.offset))(isSome(equalTo(9L)))
```
erikvanoosten, line 420: 👍

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
```diff
@@ -355,6 +355,226 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
...
+                            .tap(_ => ZIO.logDebug("Stream completed"))
+                   } yield ()
+                 }
+                   .provideSomeLayer[Kafka](
+                     consumer(client, Some(group))
+                   )
+            _ <- keepProducing.set(false)
+          } yield assertCompletes
+        },
+        test("process outstanding commits after a stopping the subscription") {
```
erikvanoosten, line 452:
```suggestion
        test("process outstanding commits after stopping the subscription") {
```

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
erikvanoosten: Never mind, this is tested below.

erikvanoosten (commented):
I didn't look at the implementation yet, only docs and tests.

docs/consuming-kafka-topics-using-zio-streams.md
erikvanoosten: I see now that `runWithGracefulShutdown` already does this. So we don't need the `ZIO.scoped` here.

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -288,6 +376,46 @@ object Consumer {
...
+
+  /**
+   * Takes a SubscriptionStreamControl for some stream and runs the given ZIO workflow on that stream such that, when
+   * interrupted, stops fetching records and gracefully waits for the ZIO workflow to complete.
+   *
+   * @param streamControl
+   *   Result of `plainStreamWithControl`, `partitionedStreamWithControl` or `partitionedAssignmentStreamWithControl`
+   * @param shutdownTimeout
+   *   Timeout for the workflow to complete after initiating the graceful shutdown
+   * @param f
```
erikvanoosten, line 400: Is this a pattern? `f` is pretty minimal name 😉

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -288,6 +376,46 @@ object Consumer {
...
+   *   Takes the stream as input and returns a ZIO workflow that processes the stream. The workflow cannot have a
+   *   meaningful result value (`Any` type), because it is expected to run forever until external interruption. `f` is
```
erikvanoosten, line 402: > because it is expected to run forever until external interruption.

Is this a hard requirement? It it is not a requirement, we can reword this a bit. For example:

> In most programs the given workflow runs until an external interruption, in these cases the result value (`Any` type) is meaningless.

However, I suspect it only adds confusion. Perhaps better remove this aspect completely.

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -288,6 +376,46 @@ object Consumer {
...
+   *   typically something like `stream => stream.mapZIO(record => ZIO.debug(record)).mapZIO(_.offset.commit)`
+   */
+  def runWithGracefulShutdown[StreamType, R, E](
+    streamControl: ZIO[Scope with Consumer, E, SubscriptionStreamControl[StreamType]],
+    shutdownTimeout: Duration = 60.seconds
```
erikvanoosten, line 407: In my experience cloud runtime environments give an application less time to shutdown. Some give 10, some give 30 seconds. So 60 seconds is quite generous. By default fs2-kafka gives 15 seconds to complete commits. Shall we also use 15s by default?

zio-kafka/src/main/scala/zio/kafka/consumer/SubscriptionStreamControl.scala
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Allows graceful shutdown of a stream, where no more records are being fetched but the in-flight records can continue
+ * to be processed and their offsets committed.
+ *
+ * As long as this object is in scope, the Kafka consumer remains subscribed.
+ *
+ * @tparam StreamType
+ *   Type of the stream returned from [[stream]]
+ */
+trait SubscriptionStreamControl[StreamType] {
```
erikvanoosten, line 13: `StreamType`, somehow I find it ugly to have the word `Type` in there. Scala style guide recommends we use a single letter. E.g. `S`.

Do we need to / can we restrict the type?

```suggestion
trait SubscriptionStreamControl[S <: X] {
```

for some type `X`

zio-kafka/src/main/scala/zio/kafka/consumer/SubscriptionStreamControl.scala
```diff
@@ -0,0 +1,32 @@
...
+   * proceed (consumer remains subscribed)
+   */
+  def stop: UIO[Unit]
+}
+
+object SubscriptionStreamControl {
+  def apply[T](stream0: T, stop0: UIO[Unit]): SubscriptionStreamControl[T] = new SubscriptionStreamControl[T] {
+    override def stream: T       = stream0
+    override def stop: UIO[Unit] = stop0
+  }
```
erikvanoosten, line 31: `T` is java style. The scala style guide follows Haskell and recommends to use single letters starting with `A`, then `B`, etc.

However, here we can use `S` because the type is supposed to be a stream.


```suggestion
  def apply[S](stream0: S, stop0: UIO[Unit]): SubscriptionStreamControl[S] = new SubscriptionStreamControl[S] {
    override def stream: S       = stream0
    override def stop: UIO[Unit] = stop0
  }
```

erikvanoosten (commented):
Still need more time to digest this.

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
svroonland: I copy-pasted the test

svroonland:
Hmm, should we instead of this:

```scala
Consumer.runWithGracefulShutdown(Consumer.partitionedStreamWithControl(Subscription.topics("topic150"), Serde.string, Serde.string)) { stream => ... }
```

offer this:

```scala
Consumer.partitionedStreamWithGracefulShutdown(Subscription.topics("topic150"), Serde.string, Serde.string) { (stream, _) => stream.flatMapPar(...) }
```

The second parameter would be the `SubscriptionStreamControl`, which you could always manually call `stop` on. Or would that prevent certain use cases.. 🤔

erikvanoosten:
> Hmm, should we instead of this:

If I understand it correctly, the proposal allows for _more_ use cases; with it you can also call `stop` for any condition you want. Is it true that after stopping, you can start consuming again?

svroonland:
Well, I mean compared to just the `partitionedStreamWithControl` method. In both cases you would need to do something with the stream that ultimately reduces to a ZIO of Any, so I don't think the `partitionedStreamWithGracefulShutdown` is limiting in that regard.

`stop` currently doesn't support that, since the stream would then be finished. We could probably build `pause` and `resume` like in #941.

erikvanoosten:
If resume after `stop` is not supported (and never will be), then I like the first proposal better where you don't need to call `stop`. What would you do after calling `stop`?

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
erikvanoosten: I like `withStream`

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -488,27 +598,135 @@ private[consumer] final class ConsumerLive private[consumer] (
...
+      stream =>
+        withStream(stream.flattenChunks)
+    }
 
   override def plainStream[R, K, V](
     subscription: Subscription,
     keyDeserializer: Deserializer[R, K],
     valueDeserializer: Deserializer[R, V],
-    bufferSize: Int
+    bufferSize: RuntimeFlags
```
guizmaii, line 681: This is probably an IDE mistake
```suggestion
    bufferSize: Int
```

guizmaii:
Hey :)

Thanks for the great work!

Here's some initial feedback:

I'm not a big fan of the `SubscriptionStreamControl` implementation.

To me, functions/methods returning it should return a Tuple `(stream, control)`. 
It avoids adding one more concept for our users to understand and learn (Kafka already has a lot of concepts)
It also simplifies the interface of the `control` type, the current one with the `[S <: ZStream[_, _, _]]` being complex
It also simplifies the returns type of our functions/methods, avoiding this kind of type: 
```scala
SubscriptionStreamControl[Stream[Throwable, Chunk[(TopicPartition, ZStream[R, Throwable, CommittableRecord[K, V]])]]]
````
in favor of:
```scala
(Stream[Throwable, Chunk[(TopicPartition, ZStream[R, Throwable, CommittableRecord[K, V]])], SubscriptionStreamControl)
```

Made the change in the PR to show how, to me, it simplifies things: https://github.com/zio/zio-kafka/pull/1207/files

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
```diff
@@ -55,16 +55,23 @@ private[consumer] final class RunloopAccess private (
...
       // starts the Runloop if not already started
       _ <- withRunloopZIO(requireRunning = true)(_.addSubscription(subscription))
       _ <- ZIO.addFinalizer {
              withRunloopZIO(requireRunning = false)(_.removeSubscription(subscription)) <*
                diagnostics.emit(Finalization.SubscriptionFinalized)
            }
-    } yield stream
+    } yield SubscriptionStreamControl(
+      ZStream.fromQueue(partitionAssignmentQueue),
+      withRunloopZIO(requireRunning = true)(_.stopSubscribedTopicPartitions(subscription)) *> partitionAssignmentQueue
```
guizmaii, line 71: Not sure you want to use `true` here. `true` means: if the runloop is not running, start it and apply the `stopSubscribedTopicPartitions` function. 
In your case, IMO, if the Runloop isn't running, calling the `control.stop` should be a no-op.

guizmaii, line 71:
```suggestion
      withRunloopZIO(requireRunning = false)(_.stopSubscribedTopicPartitions(subscription)) *> partitionAssignmentQueue
```

Not sure you want to use `true` here. `true` means: if the runloop is not running, start it and apply the `stopSubscribedTopicPartitions` function. 
In your case, IMO, if the Runloop isn't running, calling the `control.stop` should be a no-op, which will be the case if you use `false` instead

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
```diff
@@ -55,16 +55,23 @@ private[consumer] final class RunloopAccess private (
    */
   def subscribe(
     subscription: Subscription
-  ): ZIO[Scope, InvalidSubscriptionUnion, UStream[Take[Throwable, PartitionAssignment]]] =
+  ): ZIO[Scope, InvalidSubscriptionUnion, SubscriptionStreamControl[UStream[Take[Throwable, PartitionAssignment]]]] =
     for {
-      stream <- ZStream.fromHubScoped(partitionHub)
+      partitionAssignmentQueue <- Queue.unbounded[Take[Throwable, PartitionAssignment]]
```
guizmaii, line 60: Not sure to understand why you need this additional queue 🤔

guizmaii:
Didn't finish my review yet. I still have some parts of the code to explore/understand, but I have to go. I'll finish it later 🙂

svroonland:
Thanks for the feedback Jules. Agreed about the extra concept that would be unwanted. Check out my latest interface proposal where there is only a `plainStreamWithGracefulShutdown` method and `SubscriptionStreamControl` remains hidden.

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -70,6 +70,22 @@ trait Consumer {
     valueDeserializer: Deserializer[R, V]
   ): Stream[Throwable, Chunk[(TopicPartition, ZStream[R, Throwable, CommittableRecord[K, V]])]]
 
+  /**
+   * Like [[partitionedAssignmentStream]] but wraps the stream in a construct that ensures graceful shutdown
```
erikvanoosten, line 74:
```suggestion
   * Like [[partitionedAssignmentStream]] but wraps the stream in a construct that ensures graceful shutdown.
```

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -93,6 +109,22 @@ trait Consumer {
     valueDeserializer: Deserializer[R, V]
   ): Stream[Throwable, (TopicPartition, ZStream[R, Throwable, CommittableRecord[K, V]])]
 
+  /**
+   * Like [[partitionedStream]] but wraps the stream in a construct that ensures graceful shutdown
```
erikvanoosten, line 113:
```suggestion
   * Like [[partitionedStream]] but wraps the stream in a construct that ensures graceful shutdown.
```

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -111,13 +143,27 @@ trait Consumer {
...
   def plainStream[R, K, V](
     subscription: Subscription,
     keyDeserializer: Deserializer[R, K],
     valueDeserializer: Deserializer[R, V],
     bufferSize: Int = 4
   ): ZStream[R, Throwable, CommittableRecord[K, V]]
 
+  /**
+   * Like [[plainStream]] but wraps the stream in a construct that ensures graceful shutdown
+   */
```
erikvanoosten, line 156: Since this method might be the most attractive way to use zio-kafka, lets extend the documentation a bit.

e.g.
```suggestion
  /**
   * Like [[plainStream]] but wraps the stream in a construct that ensures graceful shutdown. During a graceful shutdown the consumer is stopped but the stream can complete processing and commit already fetched records.
   *
   * Example usage:
   * {{{
   *   ... todo ...
   * }}}
   */
```

We could also include a reference to the documentation (though I am always extremely happy when the scaladocs are all you need, e.g. scalatest documentation is my benchmark).

erikvanoosten (commented):
Still reading the code...

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
erikvanoosten: > In your case, IMO, if the Runloop isn't running, calling the `control.stop` should be a no-op

We need to stop the subscription also when runloop isn't running. Right?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -75,6 +75,13 @@ private[consumer] final class Runloop private (
   private[internal] def removeSubscription(subscription: Subscription): UIO[Unit] =
     commandQueue.offer(RunloopCommand.RemoveSubscription(subscription)).unit
 
+  private[internal] def stopSubscribedTopicPartitions(subscription: Subscription): UIO[Unit] =
```
erikvanoosten, line 78: This method is stopping streams, not subscriptions. WDYT of:

```suggestion
  private[internal] def stopStreams(subscription: Subscription): UIO[Unit] =
```

erikvanoosten:
I understand now that when graceful shutdown starts we're ending the subscribed streams. That should work nicely. Lets work out what will happen next to the runloop. The runloop would still be happily fetching records for that stream. When those are offered to the stream, `PartitionStreamControl.offerRecords` will probably append those records to the queue (even though it now also contains an 'end' token). Because of the 'end' token that is already in that queue, these new records will never be taken out. Back pressure will kick in (depending on the fetch strategy) and the partitions will be paused. Once we're unsubscribed, 15 seconds later, the queue will be garbage collected. So far so good.

We can do slightly better though. We're storing all these records in the queue for nothing, even potentially causing an OOM for systems that are tuned for the case where processing happens almost immediately.

My proposal is to:
1. stop accepting more records in `PartitionStreamControl.offerRecords` when the queue was ended
2. in `Runloop.handlePoll` only pass running streams to `fetchStrategy.selectPartitionsToFetch` so that partitions for ended streams are immediately paused

svroonland:
@erikvanoosten If you have some time to implement those two things, by all means.

erikvanoosten:
> @erikvanoosten If you have some time to implement those two things, by all means.

Done in commit https://github.com/zio/zio-kafka/pull/1201/commits/1218204f9924281f15da9ccec06438975d03fdb4.

Now I am wondering, how can test this?

svroonland:
Change look good. Totally forgot to implement this part.

docs/consuming-kafka-topics-using-zio-streams.md
svroonland: No longer relevant in latest version.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
svroonland: This is to inject a Take.end in the stream returned from `ZStream.fromHubScoped(partitionHub)`, at the moment when we want to gracefuly terminate the stream

erikvanoosten: This might be done more elegantly by merging the `stream <- ZStream.fromHubScoped(partitionHub)` with another stream in which we inject the Take.end. WDYT?

svroonland: Good idea. Like so?

erikvanoosten: Yeah, `stream.merge(otherStream, ZStream.HaltStrategy.Either)` should work. The `otherStream` could be build with ZStream.fromQueue, or with something like ZStream.fromZIO(promise.get).

Another option that looks even nicer is to use `stream.haltWhen(unsubscribePromise)`.

erikvanoosten: > Good idea. Like so?

Only seeing this now. Which commit are you referring to?

svroonland: That doesn't work unfortunately, since it's a stream of `Take`s (explicitly)

docs/consuming-kafka-topics-using-zio-streams.md
erikvanoosten: @svroonland This comment is still relevant. Documentation needs to be readable even with all the headers, borders, menus etc. around it.

erikvanoosten (commented):
See notes below. I feel that this is going in the right direction. Will look again when these notes are addressed. For now 🤯 
Next time I'll try to trace the shutdown sequence a bit better.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -910,6 +947,7 @@ object Runloop {
       _ <- ZIO.addFinalizer(
              ZIO.logDebug("Shutting down Runloop") *>
                runloop.shutdown *>
+               commandQueue.offer(RunloopCommand.StopRunloop) *>
```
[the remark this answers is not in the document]
svroonland: Because `shutdown` is called from `checkStreamPollInterval`, where I don't want it to stop the runloop, because then finalization will fail here (line 951)

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -661,8 +684,22 @@ private[consumer] final class Runloop private (
                 cmd.succeed *> doChangeSubscription(newSubState)
             }
         }
-      case RunloopCommand.RemoveSubscription(subscription) =>
-        state.subscriptionState match {
+      case RunloopCommand.EndStreamsBySubscription(subscription, cont) =>
+        ZIO.foreachDiscard(
+          state.assignedStreams.filter(stream => Subscription.subscriptionMatches(subscription, stream.tp))
+        )(_.end) *> cont
+          .succeed(())
```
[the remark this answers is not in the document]
svroonland: Yes, the `ZIO[Any, Nothing, Unit]` cannot fail

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -679,7 +716,7 @@ private[consumer] final class Runloop private (
                 ZIO.logDebug(s"Unsubscribing kafka consumer") *>
                   doChangeSubscription(SubscriptionState.NotSubscribed)
             }
-        }
+        }).tapBoth(cont.fail, _ => cont.succeed(()).unit)
       case RunloopCommand.RemoveAllSubscriptions => doChangeSubscription(SubscriptionState.NotSubscribed)
       case RunloopCommand.StopAllStreams =>
         for {
```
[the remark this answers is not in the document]
svroonland: The bug being that if you restart a subscription after calling `stopConsumption`, there'd be old `assignedStreams`?

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
```diff
@@ -355,6 +355,180 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
...
+                                  .transduce(Consumer.offsetBatches)
+                                  .mapZIO(batch =>
+                                    ZIO.logInfo("Starting batch commit") *> batch.commit.tapErrorCause(
+                                      ZIO.logErrorCause(_)
+                                    ) *> ZIO.logInfo("Commit done")
+                                  )
+                                  .runDrain
+                              }
+                              .forkScoped
+                          _      <- produceMany(topic, kvs)
```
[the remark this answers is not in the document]
svroonland: Thanks, will implement

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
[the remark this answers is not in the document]
erikvanoosten: Yes, indeed.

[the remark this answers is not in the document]
svroonland: Indeed ,it should not empty `pendingCommits`. The semantics would be to stop fetching new records, flush any buffered records, end the partition streams and end all top level streams.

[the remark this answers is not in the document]
svroonland: I believe that is why one of the graceful shutdown tests fails after I changed that (it's now reverted).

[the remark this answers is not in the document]
svroonland: Actually, the emptying of `assignedStreams` caused the "process outstanding commits after a graceful shutdown with aggregateAsync using `maxRebalanceDuration`" test to fail..

Will have to think a bit more about what is going on here.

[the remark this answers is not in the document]
erikvanoosten: I was already very surprised I would have missed this during my earlier refactorings. We really need to improve the documentation of this, if it becomes too complicated for ourselves...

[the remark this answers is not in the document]
svroonland: Added some docs and an assertion. When I also clear assignedStreams, I see the warning being logged and the test fails.

svroonland:
With the stronger test I found some new issues with unexpected interruptions, which by a lucky shot I was able to fix with a `forkDaemon.flatMap(_.join)`. I can't explain exactly why this happens, but with the new mechanism the test works well with a `nonFlaky(100)`, where it previously failed with just 10 repetitions.

erikvanoosten:
> which by a lucky shot I was able to fix with a `forkDaemon.flatMap(_.join)`.

Ouch. That is not a fix, that is a kludge 😞 I really wonder what is going on here

BTW, it would be nice if we can make more test stronger with `scheduleProduce` (eventually).

svroonland:
Was able to create a minimized reproducer of the issue: https://github.com/zio/zio/issues/9288

svroonland:
The abovementioned issue has been fixed and will probably be in the next ZIO release. For compatibility with older versions let's keep the `.fork.flatMap(_.join)` workaround.

erikvanoosten:
> The abovementioned issue has been fixed and will probably be in the next ZIO release. For compatibility with older versions let's keep the `.fork.flatMap(_.join)` workaround.

Great work!

Shall we slap a comment on it? E.g. something like:
```
// Workaround for a bug in ZIO up to v2.1.12. See https://github.com/zio/zio/issues/9288
```

zio-kafka/src/main/scala/zio/kafka/consumer/SubscriptionStreamControl.scala
```diff
@@ -0,0 +1,15 @@
...
+ * Allows graceful shutdown of a stream, where no more records are being fetched but the in-flight records can continue
+ * to be processed and their offsets committed.
+ *
+ * @param stream
+ *   The stream of partitions / records for this subscription
+ * @param stop
+ *   Stop fetching records for the subscribed topic-partitions and end the associated streams, while allowing commits to
+ *   proceed (consumer remains subscribed)
+ */
+final private[consumer] case class SubscriptionStreamControl[S <: ZStream[_, _, _]](stream: S, stop: UIO[Unit])
```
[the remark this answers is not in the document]
erikvanoosten: There is an opportunity to clean up the code that is using this with methods like `map`:

```scala
case class StreamControl[R, E, A]](stream: ZStream[R, E, A], stop: UIO[Unit]) {
  def map[R1 <: R, E1 >: E, B](f: ZStream[R, E, A] => ZStream[R1, E1, B]): StreamControl[R1, E1, B] =
    StreamControl(f(stream), stop)
}
```

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -404,24 +469,163 @@ private[consumer] final class ConsumerLive private[consumer] (
...
+            .tapErrorCause(cause => ZIO.logErrorCause("Error in withStream fiber in runWithGracefulShutdown", cause))
+            .forkScoped
+        result <-
+          fib.join.onInterrupt(
+            control.stop *>
+              fib.join
+                .timeout(shutdownTimeout)
+                .someOrElseZIO(
+                  ZIO.logError(
+                    "Timeout joining withStream fiber in runWithGracefulShutdown. Not all pending commits may have been processed."
```
erikvanoosten, line 612: Here is my attempt at rewriting the message more towards the user's view. Not sure if this is entirely correct though.
```suggestion
                    "Timeout waiting for `withStream` to gracefully shut down. Not all in-flight records may have been processed."
```

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -404,24 +469,163 @@ private[consumer] final class ConsumerLive private[consumer] (
...
+                  )
+                )
+                .tapErrorCause(cause =>
+                  ZIO.logErrorCause("Error joining withStream fiber in runWithGracefulShutdown", cause)
```
erikvanoosten, line 616: If I understand this correctly, the problem is not that the join fails, but more likely the stream itself.
```suggestion
                  ZIO.logErrorCause("Stream failed while awaiting its graceful shutdown", cause)
```

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -404,24 +469,163 @@ private[consumer] final class ConsumerLive private[consumer] (
...
+                )
+                // The fork and join is a workaround for https://github.com/zio/zio/issues/9288 for ZIO <= 2.1.12
+                .forkDaemon
+                .flatMap(_.join)
+                .tapErrorCause(cause =>
```
erikvanoosten, line 621: Question: when we remove this workaround (I am actually in favor of that, we can document that we require zio 2.1.13+), do we then still need this second `tapErrorCause`?

erikvanoosten (commented):
So many little comments. I'll push this now so you have something to work with. Meanwhile I'll continue reviewing.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -203,17 +213,20 @@ private[consumer] final class Runloop private (
...
                s" ${pendingCommitCount} pending commits," +
-               s" resuming $partitionsToFetch partitions"
+               s" resuming ${partitionsToFetch.size} out of ${state.assignedStreams.size} partitions"
            )
       _ <- currentStateRef.set(state)
       pollResult <-
         consumer.runloopAccess { c =>
           for {
+            assignment <- ZIO.attempt(c.assignment())
+            _          <- verifyAssignedStreamsMatchesAssignment(state.assignedStreams, assignment.asScala.toSet)
```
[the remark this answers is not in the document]
svroonland: It's a helpful assertion to verify that we implemented stream assignment & cleanup properly. During development of this PR it pointed out some errors.

It's correct because it's executed before polling, so at any time the streams and the consumer's assignment should match.

zio-kafka/src/main/scala/zio/kafka/consumer/SubscriptionStreamControl.scala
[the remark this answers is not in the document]
svroonland: Nice, I like it

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
```diff
@@ -52,19 +52,26 @@ private[consumer] final class RunloopAccess private (
...
       // starts the Runloop if not already started
       _ <- withRunloopZIO(requireRunning = true)(_.addSubscription(subscription))
       _ <- ZIO.addFinalizer {
-             withRunloopZIO(requireRunning = false)(_.removeSubscription(subscription)) <*
+             withRunloopZIO(requireRunning = false)(_.removeSubscription(subscription).orDie) <*
                diagnostics.emit(Finalization.SubscriptionFinalized)
            }
-    } yield stream
+    } yield SubscriptionStreamControl(
+      stream = stream.merge(ZStream.fromZIO(end.await).as(Take.end)),
```
erikvanoosten, line 72: When I read (line 72) this some alarm bells went of in my head.

`PartitionStreamControl` registers which records (offsets) were pulled. This information is later used by the rebalance listener when `rebalanceSafeCommit` is enabled.

When the 'end' gets merged into the stream provided by `PartitionStreamControl`, it could be that the latter just pulled a chunk of records, but before the chunk was given downstream, the end is merged in. This causes downstream to never see those records, even though `PartitionStreamControl` thinks those have been pulled. When this happens, the ending rebalance will wait for maxRebalanceDuration.

erikvanoosten: A solution could be to stop offering data from the runloop (see also https://github.com/zio/zio-kafka/pull/1201/files#r1986290475).

svroonland: The `stream` here is not a single partition stream, but the stream of partition streams (`ZStream[Any, Nothing, Take[Throwable, (TopicPartition, Stream[Throwable, ByteArrayCommittableRecord])]]`). So the `end` just ends the top-level stream. 

Does that change your concern?

erikvanoosten: > Does that change your concern?

It does! 🤦

Perhaps we should add a comment. Something like "Ends the stream of partition streams." WDYT?

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -69,6 +69,28 @@ trait Consumer {
     valueDeserializer: Deserializer[R, V]
   ): Stream[Throwable, Chunk[(TopicPartition, ZStream[R, Throwable, CommittableRecord[K, V]])]]
 
+  /**
+   * Like [[partitionedAssignmentStream]] but wraps the stream in a construct that ensures graceful shutdown.
+   *
+   * When this effect is interrupted, all partition streams are closed upstream, allowing the stream created by
```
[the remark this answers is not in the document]
svroonland: I can see why that is unclear, 'upstream' and 'downstream' are not quite official ZStream concepts. I mean to say here that the source of the stream is ended so that subsequent stream stages can complete.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
svroonland: Done

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -404,24 +469,163 @@ private[consumer] final class ConsumerLive private[consumer] (
...
+    withPartitionedAssignmentStream(subscription, keyDeserializer, valueDeserializer, shutdownTimeout) { stream =>
+      withStream(stream.flattenChunks)
+    }
+
   override def plainStream[R, K, V](
     subscription: Subscription,
     keyDeserializer: Deserializer[R, K],
     valueDeserializer: Deserializer[R, V],
-    bufferSize: Int
+    bufferSize: Int = 4
```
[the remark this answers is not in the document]
svroonland: Not sure why I made this change, reverted it.

svroonland:
I was preparing some work for removing `stopConsumption` but I'm now beginning to wonder if stopping the stream via interruption is the best API. With `stopConsumption`, we could stop the streams and access the result of the `runCollect` or something like that. With this interruption-based stopping, we can no longer return a value from `with*Stream`. Not sure how useful that is in practice, but it is used in our unit tests. 

We've discussed it earlier (about a year ago), but perhaps this alternative API is more powerful for our users, for usage scenario's we haven't thought of.. https://github.com/zio/zio-kafka/pull/1501

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -203,17 +213,20 @@ private[consumer] final class Runloop private (
 
   private def handlePoll(state: State): Task[State] = {
     for {
-      partitionsToFetch  <- settings.fetchStrategy.selectPartitionsToFetch(state.assignedStreams)
-      pendingCommitCount <- committer.pendingCommitCount
+      runningStreamsBeforePoll <- ZIO.filter(state.assignedStreams)(_.isRunning)
```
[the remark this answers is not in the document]
svroonland: Can you elaborate on that?

docs/consuming-kafka-topics-using-zio-streams.md
```diff
@@ -75,3 +75,33 @@ Each time a new partition is assigned to the consumer, a new partition stream is
...
 
 In this example, each partition is processed in parallel, on separate fibers, while a single fiber is doing commits.
+
+## Controlled shutdown (experimental)
+
+The examples above will keep processing records forever, or until the fiber is interrupted, typically at application shutdown. When interrupted, some records may be 'in-flight', e.g. being processed by one of the stages of your consumer stream user code. Those records will be processed partly and their offsets may not be committed. For fast shutdown in an at-least-once processing scenario this is fine.
+
+zio-kafka also supports a _graceful shutdown_, where the fetching of records for the subscribed topics/partitions is stopped, the streams are ended and all downstream stages are completed, allowing in-flight records to be fully processed.
+
+Use the `with*Stream` variants of `plainStream`, `partitionedStream` and `partitionedAssignmentStream` for this purpose. These methods accept a parameter that describes the execution of a stream, which is gracefully ended when the method is interrupted.
```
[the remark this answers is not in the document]
svroonland: Updated it

docs/consuming-kafka-topics-using-zio-streams.md
```diff
@@ -82,26 +82,29 @@ The examples above will keep processing records forever, or until the fiber is i
...
 import zio.Console.printLine
 import zio.kafka.consumer._
 
-consumer.withPartitionedStream(
-  Subscription.topics("topic150"), 
-  Serde.string, 
-  Serde.string
-) { stream =>
-    stream.flatMapPar(Int.MaxValue) { case (topicPartition, partitionStream) =>
+for {
```
erikvanoosten, line 95: Should we wrap all of this in a `ZIO.scoped` block?

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -164,27 +120,62 @@ trait Consumer {
...
+  ): ZIO[Scope with R, Throwable, StreamControl[R, Throwable, CommittableRecord[K, V]]]
+
+  /**
+   * Like [[partitionedStream]] but returns an object to control the stream.
+   *
+   * The Scope provided to the returned effect controls the lifetime of the subscription. The subscription is
+   * unsubscribed when the scope is ended. Calling [[StreamControl.end]] stops fetching data for the subscription
+   * partitions but will not unsubscribe until the Scope is ended.
+   *
+   * EXPERIMENTAL API
```
erikvanoosten, line 149:
```suggestion
   * WARNING: this is an EXPERIMENTAL API and may disappear or change in an incompatible way without notice in any zio-kafka version.
```

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -164,27 +120,62 @@ trait Consumer {
...
+  ]
+
+  /**
+   * Like [[partitionedAssignmentStream]] but returns an object to control the stream.
+   *
+   * The Scope provided to the returned effect controls the lifetime of the subscription. The subscription is
+   * unsubscribed when the scope is ended. Calling [[StreamControl.end]] stops fetching data for the subscription
+   * partitions but will not unsubscribe until the Scope is ended.
+   *
+   * EXPERIMENTAL API
```
erikvanoosten, line 168: Same

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -405,6 +396,54 @@ object Consumer {
...
     case object None     extends AutoOffsetStrategy
   }
+
+  /**
+   * Takes a StreamControl for some stream and runs the given ZIO workflow on that stream such that, when interrupted,
+   * stops fetching records and gracefully waits for the ZIO workflow to complete.
+   *
+   * This is useful for running streams from within your application's Main class, such that streams are cleanly stopped
+   * when the application is shutdown (for example by your container runtime).
+   *
```
erikvanoosten, line 406: Add an experimental tag here as well?

erikvanoosten (commented):
> The Scope provided to the returned effect controls the lifetime of the subscription. The subscription is unsubscribed when the scope is ended. Calling [[StreamControl.end]] stops fetching data for the subscription partitions but will not unsubscribe until the Scope is ended.

How do you coordinate calling `end` and closing the scope? Does `end` return immediately, or does it block until its safe to close the scope? Or is there another way to know the scope may be closed?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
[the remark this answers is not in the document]
erikvanoosten: `streamControl.isRunning` only returns `true` when the stream is already ended.

We should stop fetching data for the stream _before_ that. We can stop fetching as soon as `streamcontrol.end` is called.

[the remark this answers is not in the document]
svroonland: Absolutely. Made the change

svroonland:
> How do you coordinate calling end and closing the scope? Does end return immediately, or does it block until its safe to close the scope? Or is there another way to know the scope may be closed?

`end` returns when the command has been processed by the Runloop, so after all relevant PartitionStreamControls have put a `Take.end` in their `dataQueue`. It's okay to call end more than once (see unit test). 

When the scope is closed without `end` having been called, there is no graceful shutdown.

Does that make sense?

erikvanoosten:
Yes, that makes sense. However, could we also wait until the stream provided by the user ends?

svroonland:
The intended usage pattern of this construct is something like:
```
streamControl <- Consumer.plainStreamWithControl
fib <- streamControl.stream.tap(..).runDrain
...
// At some point
_ <- streamControl.end
_ <- fib.join
```

So no additional need to wait for the stream to end when closing the scope. In fact, if we do that we might actually limit useability in some use cases we haven't thought of yet. If you use `Consumer.runWithGracefulShutdown` you can just interupt the fiber and it will `end` and wait for the stream to complete.

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
```diff
@@ -327,91 +327,427 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
...
+            producer <- KafkaTestUtils.makeProducer
+            _ <-
+              KafkaTestUtils.scheduledProduce(producer, topic, Schedule.fixed(50.millis).jittered).runDrain.forkScoped
+
+            consumer <- KafkaTestUtils.makeConsumer(
+                          client,
+                          Some(group),
+                          commitTimeout = 4.seconds,
+                          rebalanceSafeCommits = true,
+                          maxRebalanceDuration = 6.seconds
```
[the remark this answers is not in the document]
erikvanoosten: Note to self done in https://github.com/zio/zio-kafka/pull/1505 🙂

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -115,13 +115,71 @@ trait Consumer {
    * On completion of the stream, the consumer is unsubscribed. In case of multiple subscriptions, the total consumer
    * subscription is changed to exclude this subscription.
    */
+
```
erikvanoosten, line 118: this newline should go away

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -410,11 +520,44 @@ private[consumer] final class ConsumerLive private[consumer] (
...
+      }
+    )
+  }
+
   override def partitionedStream[R, K, V](
     subscription: Subscription,
     keyDeserializer: Deserializer[R, K],
     valueDeserializer: Deserializer[R, V]
-  ): ZStream[Any, Throwable, (TopicPartition, ZStream[R, Throwable, CommittableRecord[K, V]])] =
+  ): Stream[
```
erikvanoosten, line 554: IMHO we should not use `Stream`, only `ZStream`. WDYT?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
```diff
@@ -52,19 +52,28 @@ private[consumer] final class RunloopAccess private (
...
       // starts the Runloop if not already started
       _ <- withRunloopZIO(requireRunning = true)(_.addSubscription(subscription))
       _ <- ZIO.addFinalizer {
-             withRunloopZIO(requireRunning = false)(_.removeSubscription(subscription)) <*
+             withRunloopZIO(requireRunning = false)(_.removeSubscription(subscription).orDie) <*
                diagnostics.emit(Finalization.SubscriptionFinalized)
            }
-    } yield stream
+    } yield new StreamControl[Any, Nothing, Take[Throwable, PartitionAssignment]] {
+      override def stream = partitionAssignmentStream.merge(ZStream.fromZIO(ended.await).as(Take.end))
```
erikvanoosten, line 72:
```suggestion
      override def stream = partitionAssignmentStream.haltWhen(ended)
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
```diff
@@ -52,19 +52,28 @@ private[consumer] final class RunloopAccess private (
...
+  ): ZIO[Scope, InvalidSubscriptionUnion, StreamControl[Any, Nothing, Take[Throwable, PartitionAssignment]]] =
     for {
-      stream <- ZStream.fromHubScoped(partitionHub)
+      ended                     <- Promise.make[Nothing, Unit] // For ending the stream of partition streams
+      partitionAssignmentStream <- ZStream.fromHubScoped(partitionHub)
       // starts the Runloop if not already started
       _ <- withRunloopZIO(requireRunning = true)(_.addSubscription(subscription))
       _ <- ZIO.addFinalizer {
-             withRunloopZIO(requireRunning = false)(_.removeSubscription(subscription)) <*
+             withRunloopZIO(requireRunning = false)(_.removeSubscription(subscription).orDie) <*
```
erikvanoosten, line 68: The `orDie` should not be needed after proposed changes in runloop.

erikvanoosten (commented):
Unfortunately, I thought of a new problem. When there is a rebalance, new partitions might be assigned to this consumer, even though the consumer is trying to do a graceful shutdown! We need to make sure NOT to start new streams for subscriptions for which we received a EndStreamsBySubscription.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -421,8 +443,19 @@ private[consumer] final class Runloop private (
                 cmd.succeed *> doChangeSubscription(newSubState)
             }
         }
-      case RunloopCommand.RemoveSubscription(subscription) =>
-        state.subscriptionState match {
+      case RunloopCommand.EndStreamsBySubscription(subscription, cont) =>
+        for {
+          _ <- ZIO.foreachDiscard(
+                 state.assignedStreams.filter(stream => Subscription.subscriptionMatches(subscription, stream.tp))
+               )(_.end)
```
erikvanoosten, line 450:
```suggestion
          _ <- ZIO.foreachDiscard {
            state.assignedStreams.filter(stream => Subscription.subscriptionMatches(subscription, stream.tp))
          }(_.end)
```

docs/consuming-kafka-topics-using-zio-streams.md
```diff
@@ -75,3 +75,38 @@ Each time a new partition is assigned to the consumer, a new partition stream is
...
+            partitionStream
+              .tap(record => printLine(s"key: ${record.key}, value: ${record.value}"))
+              .map(_.offset)
+          }
+          .aggregateAsync(Consumer.offsetBatches)
+          .mapZIO(_.commit)
+          .runDrain
+    } yield ()
+}
+```
```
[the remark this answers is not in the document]
erikvanoosten: I meant this text:

The intended usage pattern of this construct is something like:

```scala
streamControl <- Consumer.plainStreamWithControl
fib <- streamControl.stream.tap(..).runDrain
...
// At some point
_ <- streamControl.end
_ <- fib.join
```

erikvanoosten:
Okay, my review is complete. Finally 🙂
Only a few things left to address and then I think its good to go!

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
svroonland: Unfortunately it still is. Previously we just fire-and-forget with this command, now we await its result. It could always fail, because of calling `unsubscribe()` on the `KafkaConsumer`.

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
svroonland: Regardless of my opinion, there is no need for this change in this PR

svroonland:
> Unfortunately, I thought of a new problem. When there is a rebalance, new partitions might be assigned to this consumer, even though the consumer is trying to do a graceful shutdown! We need to make sure NOT to start new streams for subscriptions for which we received a EndStreamsBySubscription.

Ouch, we'll need to think about how to implement and test this..

docs/consuming-kafka-topics-using-zio-streams.md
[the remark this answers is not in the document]
svroonland: Added to docs

docs/consuming-kafka-topics-using-zio-streams.md
```diff
@@ -75,3 +75,38 @@ Each time a new partition is assigned to the consumer, a new partition stream is
...
+
+## Controlled shutdown (experimental)
+
+The examples above will keep processing records forever, or until the fiber is interrupted, typically at application shutdown. When interrupted, some records may be 'in-flight', e.g. being processed by one of the stages of your consumer stream user code. Those records will be processed partly and their offsets may not be committed. For fast shutdown in an at-least-once processing scenario this is fine.
+
+zio-kafka also supports a _graceful shutdown_, where the fetching of records for the subscribed topics/partitions is stopped, the streams are ended and all downstream stages are completed, allowing in-flight records to be fully processed.
+
+Use the `*StreamWithControl` variants of `plainStream`, `partitionedStream` and `partitionedAssignmentStream` for this purpose. These methods return a `StreamControl` object allowing you to stop fetching records and terminate the execution of the stream gracefully. 
+
+There is also the `Consumer.runWithGracefulShutdown` method which can gracefully terminate the stream upon fiber interruption, useful for a controlled shutdown when your application is terminated.
```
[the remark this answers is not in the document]
svroonland: Agreed, added to docs

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
erikvanoosten: Sure, but that doesn't matter. My proposal was to make `Runloop.removeSubscription` return UIO again because any problem there is a defect and not a failure.

docs/consuming-kafka-topics-using-zio-streams.md
```diff
@@ -82,11 +82,37 @@ The examples above will keep processing records forever, or until the fiber is i
...
+import zio._
+
+ZIO.scoped {
+  for {
+    control <- consumer.partitionedStreamWithControl(
+      Subscription.topics("topic150"),
+      Serde.string,
+      Serde.string
+    )
+    _ <- Consumer.runWithGracefulShutdown(control, shutdownTimeout = 30.seconds) { stream =>
```
erikvanoosten, line 101: Accessor method have been deprecated 🙂 

```suggestion
    _ <- consumer.runWithGracefulShutdown(control, shutdownTimeout = 30.seconds) { stream =>
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
erikvanoosten: Ah, I am mixing up things. Sorry, you're right.

erikvanoosten:
> we'll need to think about how to implement and test this..

I can't think of anything else than adding terminating subscriptions to the runloop state. They can be removed as soon as the subscription is actually terminated.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -439,9 +472,11 @@ private[consumer] final class Runloop private (
                 ZIO.logDebug(s"Unsubscribing kafka consumer") *>
                   doChangeSubscription(SubscriptionState.NotSubscribed)
             }
-        }
+        }).tapBoth(cont.fail, _ => cont.succeed(()).unit)
```
erikvanoosten, line 475:
```suggestion
        }).intoPromise(cont)
```

erikvanoosten: No never mind. That doesn't work because `intoPromise` changes the result type.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopAccess.scala
svroonland: `haltWhen` did not work, `interruptWhen` did. I think `haltWhen` only ends upon the next pull, while `interruptWhen` is immediate.

erikvanoosten: That is odd. I do not see how this merge construct is different from haltWhen. Both should end the stream upon the next pull.

svroonland:
> Unfortunately, I thought of a new problem. When there is a rebalance, new partitions might be assigned to this consumer, even though the consumer is trying to do a graceful shutdown! We need to make sure NOT to start new streams for subscriptions for which we received a EndStreamsBySubscription.

Addressed in latest commit. The [`interruptWhen(ended)` ](https://github.com/zio/zio-kafka/pull/1201/files#diff-689216f569dbe08f28a5245c0b6d35721937d800ccd3743164e2c42a4ffdb4c6R73)will prevent partitions assigned during rebalance after initiating graceful shutdown to be consumed.

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
```diff
@@ -323,91 +323,588 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
...
+                        }
+                        .aggregateAsync(Consumer.offsetBatches)
+                        .mapZIO(_.commit)
+                        .runDrain *>
+                        consumer.committed(Set(new TopicPartition(topic, 0))).map(_.values.head)
+
+            lastOffset <- lastProcessedOffset.get
+          } yield assertTrue(offset.map(_.offset).contains(lastOffset))
+        } @@ TestAspect.nonFlaky(2),
+        test("handles RebalanceInProgressExceptions transparently") {
```
erikvanoosten, line 413: This test `handles RebalanceInProgressExceptions transparently` was removed in https://github.com/zio/zio-kafka/pull/1494 as it is now better tested in a lower level test. Please remove it from the PR.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -52,17 +53,20 @@ final class PartitionStreamControl private (
...
-      for {
-        now <- Clock.nanoTime
-        newPullDeadline = now + maxStreamPullIntervalNanos
-        _ <- queueInfoRef.update(_.withOffer(newPullDeadline, data.size))
-        _ <- dataQueue.offer(Take.chunk(data))
-      } yield ()
-    }
+  private[internal] def offerRecords(data: Chunk[ByteArrayCommittableRecord]): UIO[Unit] =
+    ZIO.ifZIO(hasEnded)(
+      onTrue = ZIO.dieMessage("Partition stream has already ended, cannot offer more record"),
```
erikvanoosten, line 58:
```suggestion
      onTrue = ZIO.dieMessage("Partition stream has already ended, cannot offer more records"),
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -108,17 +112,26 @@ final class PartitionStreamControl private (
   private[internal] def end: ZIO[Any, Nothing, Unit] =
     logAnnotate {
       ZIO.logDebug(s"Partition ${tp.toString} ending") *>
-        dataQueue.offer(Take.end).unit
+        dataQueue.offer(Take.end).unit *> hasEndedRef.set(true)
```
erikvanoosten, line 115: Shouldn't this be the other way around? With the original order method `offerRecords` could insert new records after `offer(Take.end)` and before `hasEndedRef.set(true)`.

```suggestion
        hasEndedRef.set(true) *>
        dataQueue.offer(Take.end).unit
```

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -123,6 +123,63 @@ trait Consumer {
     bufferSize: Int = 4
   ): ZStream[R, Throwable, CommittableRecord[K, V]]
 
+  /**
+   * Like [[plainStream]] but returns an object to control the stream.
+   *
+   * The Scope provided to the returned effect controls the lifetime of the subscription. The subscription is
+   * unsubscribed when the scope is ended. Calling [[StreamControl.end]] stops fetching data for the subscription
+   * partitions but will not unsubscribe until the Scope is ended.
```
erikvanoosten, line 131: This API is more complex then the other things we have. IMHO we should point to the documentation (for example to `docs/consuming-kafka-topics-using-zio-streams.md`) .

Don't forget to add that this when the consumer has multiple subscriptions, they all need to be stopped at the same time, keeping some active but not others it nos supported.

(Same comment for the other new methods.)

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
```diff
@@ -323,91 +323,588 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
...
+                                            nr <- processedMessageOffsets
+                                                    .updateAndGet(
+                                                      _ :+ (tp -> record.offset.offset)
+                                                    )
+                                                    .map(_.size)
+                                            _ <- stop.succeed(()).when(nr == 10)
+                                          } yield Seq(record.offset)
+                                        }
+                                      }
+                                      .transduce(Consumer.offsetBatches)
```
erikvanoosten, line 540: I don't think it matters but I think (by using `transduce`) this doesn't do any grouping so each record will be committed separately.

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
```diff
@@ -323,91 +323,588 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
...
-                        } yield record.offset
+            producer <- KafkaTestUtils.makeProducer
+            consumer <- KafkaTestUtils.makeConsumer(client, Some(group))
+            _        <- KafkaTestUtils.produceMany(producer, topic1, kvs)
+            _        <- KafkaTestUtils.produceMany(producer, topic2, kvs)
+            _ <- ZIO.scoped {
+                   for {
+                     stream1Started <- Promise.make[Nothing, Unit]
+                     stream1Done    <- Promise.make[Nothing, Unit]
+//                     stream1Interrupted <- Promise.make[Nothing, Unit]
```
erikvanoosten, line 782: Can this line be removed?

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
```diff
@@ -323,91 +323,588 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
...
-                      consumer.committed(Set(new TopicPartition(topic, 0))).map(_.values.head)
-
-          lastOffset <- lastProcessedOffset.get
-        } yield assertTrue(offset.map(_.offset).contains(lastOffset))
-      } @@ TestAspect.nonFlaky(2),
+                      .mapZIO(offsetBatch =>
+                        stream1Started.succeed(()) *>
+                          offsetBatch.commit.repeatUntilZIO(_ =>
+                            stream2Started.isDone
+                          ) // Somewhat artifical condition, to simulate that consumer1 is still processing records. We need to keep committing to drive polling
```
erikvanoosten, line 873:
```suggestion
                          ) // Somewhat artificial condition: to simulate that consumer1 is still processing records we need to keep committing to drive polling
```

svroonland: No, they are both separate statements :)

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
```diff
@@ -962,18 +1461,19 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
...
-            } yield assertTrue((c1Keys intersect c2Keys).isEmpty)
+            } yield assertTrue((c1Keys.toSet intersect c2Keys).isEmpty)
           }
 
         // Test for both default partition assignment strategies
         Seq(
-          testForPartitionAssignmentStrategy[RangeAssignor],
-          testForPartitionAssignmentStrategy[CooperativeStickyAssignor]
+          testForPartitionAssignmentStrategy[RangeAssignor]
+//          testForPartitionAssignmentStrategy[CooperativeStickyAssignor]
```
erikvanoosten, line 1387: Why is this assigned commented out?

svroonland: No idea, not intentional

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
```diff
@@ -1551,7 +2047,7 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
     )
       .provideSomeShared[Scope](
         Kafka.embedded
-      ) @@ withLiveClock @@ timeout(2.minutes) @@ TestAspect.timed
+      ) @@ withLiveClock @@ timeout(10.minutes) @@ TestAspect.timed @@ TestAspect.sequential
```
erikvanoosten, line 1961: Are these changes needed?

The new tests have `@@ nonFlaky(10)`, that does increase total test time quite a bit. Can we reduce that?

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
erikvanoosten: The second part of this comment is not yet addressed.

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
erikvanoosten: Okay. But then at least fix the spelling typo :)

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
erikvanoosten: _All_ the new tests have `@@ nonFlaky(10)`. If its needed fine, but please remove it it not.

erikvanoosten: Why was `@@ TestAspect.sequential` added?

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
erikvanoosten: The new tests use `transduce` in more places. Better update all of them.

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
erikvanoosten: No worries, we can keep this for another PR.

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
erikvanoosten: > Don't forget to add that this when the consumer has multiple subscriptions, they all need to be stopped at the same time, keeping some active but not others it nos supported.

Please ignore this part, it was based on a misunderstanding.

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
erikvanoosten: See https://github.com/zio/zio-kafka/pull/1521.

PR #1205, opened 2024-03-30
Do not run user rebalance listener on same thread runtime

svroonland:
Since #1098 the rebalance listener runs on a special same thread runtime. This is also true for the user provided rebalance listener. The same thread runtime is not convenient to use at all; a lot of care must be taken to keep the zio fiber on the same thread. Note, the only reason for using the same thread runtime is to enforce the same-thread access that is required by the kafka consumer (see #1098 for more details). However, the kafka consumer that is passed to the rebalance listener should be used via the internal `ConsumerAccess`.

Since #1098 it is no longer necessary to use the rebalance listener to commit offsets, this is done with the `rebalanceSafeCommits` feature. We do not see other use cases for the kafka consumer from inside the rebalance listener.

For these reasons, this change runs user provided rebalance listener on the regular runtime again. In addition, access to the kafka consumer from within the user rebalance listeners is removed.


erikvanoosten:
Can we say this is an 'addition' iso of a 'correction'? 😉

erikvanoosten (approved):

erikvanoosten:
We need to change some documentation as well.

erikvanoosten:
For example we can remove this remark in `RebalanceListener`:

> Note that the given ZIO effects are executed directly on the Kafka poll thread. Fork and shift to another executor when this is not desired.

erikvanoosten (approved):

erikvanoosten:
I remember now why I didn't do this in the context of #1098. This is because running on another executor means you can't use the kafka consumer, at least not the one that is passed in. You have to use `ConsumerAccess` which is an internal API.

I see 3 options:
* We keep as is, and properly document that you have to be careful because you're running on a restricted executor.
* We go back to pre-1098 state, where users can't use the consumer that is given to the rebalance listener.
* We update this PR such that we don't even pass in the consumer into the rebalance listener.

svroonland:
Ah that is good to know. I would be in favour of #4. Any downsides to that, use cases where users explicitly want to commit offsets? I suppose that is now supported by `rebalanceSafeCommits`

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -239,7 +241,8 @@ private[consumer] final class Runloop private (
         } yield ()
     )
 
-    recordRebalanceRebalancingListener ++ settings.rebalanceListener
+    (recordRebalanceRebalancingListener ++ settings.rebalanceListener.runOnExecutor(topLevelExecutor))
+      .toKafka(sameThreadRuntime, rc)
```
guizmaii, line 245: What do you think of something like this:

```scala
    // Here we just want to avoid any executor shift if the user provided listener is the noop listener.
    val userRebalanceListener =
      settings.rebalanceListener match {
        case RebalanceListener.noop => RebalanceListener.noop
        case _                      => settings.rebalanceListener.runOnExecutor(topLevelExecutor)
      }

    (recordRebalanceRebalancingListener ++ userRebalanceListener).toKafka(sameThreadRuntime, rc)
```

guizmaii, line 245: What do you think of something like this:

```suggestion
    // Here we just want to avoid any executor shift if the user provided listener is the noop listener.
    val userRebalanceListener =
      settings.rebalanceListener match {
        case RebalanceListener.noop => RebalanceListener.noop
        case _                      => settings.rebalanceListener.runOnExecutor(topLevelExecutor)
      }

    (recordRebalanceRebalancingListener ++ userRebalanceListener).toKafka(sameThreadRuntime, rc)
```

guizmaii, line 245: Sorry if it's a stupid question: Do we still need to expose publicly the `toKafka` method? 🤔

erikvanoosten:
> Ah that is good to know. I would be in favour of #4.

Me as well.

> Any downsides to that, use cases where users explicitly want to commit offsets? I suppose that is now supported by `rebalanceSafeCommits`

Exactly. What else could someone be doing in the rebalance listener that requires the consumer?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Most people will not install a user rebalance listener so I like this idea.

erikvanoosten: Not a stupid question. I think it was never useful for our users.

erikvanoosten: This is implemented now.

erikvanoosten: toKafka is no longer exposed.

erikvanoosten (approved):

guizmaii (approved):
LGTM 🙂

PR #1207, opened 2024-04-05
Proposal: Prefer Tuples over complex `SubscriptionStreamControl` interface

guizmaii:


erikvanoosten:
`SubscriptionStreamControl` is not more or less complex than a tuple. The problem with tuples is that you can not decompose them in a for comprehension like `(a, b) <- monadicTuple` unless you have the better-monads compiler plugin. Without it you have to do a `t <- monadicTuple; (a, b) = t`. With better-monads you can still write `SubscriptionStreamControl(a, b) <- monadicCaseClass`.

So in short: without better-monads a case class gives better DX, with better-monads the options are equivalent.

PR #1214, opened 2024-04-13
Add MIMA checks to ensure binary compatibility

erikvanoosten:
Add MIMA checks to ensure binary compatibility across different zio-kafka versions.

Also: use scala 3 syntax in sbt files.

Based on #1018 by @myazinn.

Mima is disabled in this PR because master is currently _not_ binary compatible with the latest released version.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,16 @@
 import sbt.Def
 import MimaSettings.mimaSettings
 
+/**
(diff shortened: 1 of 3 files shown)


erikvanoosten:
Note: I cancelled the bench and profile actions. Only the "lint" action has to pass.

svroonland (approved):

PR #1221, opened 2024-04-16
Remove no longer needed collection-compat

erikvanoosten:


svroonland (approved):
What was it needed for, scala 2.12?

erikvanoosten:
> What was it needed for, scala 2.12?

Probably.

PR #1230, opened 2024-04-23
Initialize Java Producer in blocking threadpool

kyri-petrou:
As far as I understand, the initialization of the Java producer is side-effecting and blocking. It seems that while the Consumer initialization is wrapped in ZIO.blocking, the initialization of the Producer is not


erikvanoosten (approved):

erikvanoosten:
Thanks @kyri-petrou . I am not sure if IO happens in the kafka producer constructor, but it for sure does a lot!

Issue #1250, opened 2024-06-07
Partition lost not recovering, possible issue with `RunLoop`


josdirksen:
This might be related to the https://github.com/zio/zio-kafka/issues/1233 issue, but the last couple of weeks / months we see issues where after a partition is lost, it isn't recovering correctly. We've tried to analyze or debug it, but this occurs so infrequently that we haven't been able to isolate it.

By analyzing the code we might have identified the reason, but there is so much async stuff happening there, that we might be interpreting stuff wrongly.

What happens in our case is the following:

1. Services are consuming from n partitions. Note that these are partitions for a topic that has intervals with little to no traffic.
2. At a certain point, usually in the middle of the night, there is some network issue where kafka-client identifies the partitions as lost.
3. We see these lost partitions in the logging, but partitions aren't being recovered, and we also see no errors in our own client code that the streams have failed.

For partitions that are revoked everything seems to be working correctly though.

What we see as possible cause for this is this. In the `Runloop` this happens for lost partitions:

```scala
      onLost = lostTps =>
        for {
          _              <- ZIO.logDebug(s"${lostTps.size} partitions are lost")
          rebalanceEvent <- lastRebalanceEvent.get
          state          <- currentStateRef.get
          lostStreams = state.assignedStreams.filter(control => lostTps.contains(control.tp))
          _ <- ZIO.foreachDiscard(lostStreams)(_.lost)
          _ <- lastRebalanceEvent.set(rebalanceEvent.onLost(lostTps))
          _ <- ZIO.logTrace(s"onLost done")
        } yield ()
```

Resulting in this call in the `PartitionStreamControl`:

```scala
  private[internal] def lost: UIO[Boolean] = {
    val lostException = new RuntimeException(s"Partition ${tp.toString} was lost") with NoStackTrace
    interruptionPromise.fail(lostException)
  }
```

Looking at the way the `interruptionPromise` is handled this doesn't seem to work correctly when there are no records to be processed. In `PartitionStreamControl` we've got this repeating effect:

```scala
                 ZStream.repeatZIOChunk {
                   // First try to take all records that are available right now.
                   // When no data is available, request more data and await its arrival.
                   dataQueue.takeAll.flatMap(data => if (data.isEmpty) requestAndAwaitData else ZIO.succeed(data))
                 }.flattenTake.chunksWith { s =>
                   s.tap(records => registerPull(queueInfo, records))
[...]

PR #1251, opened 2024-06-07
Allow stream to be interrupted when there is no traffic

erikvanoosten:
When a partition is lost while there is no traffic, `PartitionStreamControl` is blocked waiting for data. We fix this by racing with the `interruptPromise`. (Thanks @josdirksen for the analysis! See #1250.)

Note: this situation can only occur with lost partitions. Timeouts (the other reason for interrupts) do not occur when there is no traffic. Currently, we have no way to test lost partitions. Therefore, there are no added tests.

This PR does _not_ change how lost partitions are handled. That is, the stream for the partition that is lost is interrupted, the other streams are closed gracefully, the consumer aborts with an error.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
@@ -142,9 +142,11 @@ object PartitionStreamControl {
       queueInfo           <- Ref.make(QueueInfo(now, 0, None, 0))
       requestAndAwaitData =
         for {
-          _     <- commandQueue.offer(RunloopCommand.Request(tp))
-          _     <- diagnostics.emit(DiagnosticEvent.Request(tp))
-          taken <- dataQueue.takeBetween(1, Int.MaxValue)
+          _ <- commandQueue.offer(RunloopCommand.Request(tp))
+          _ <- diagnostics.emit(DiagnosticEvent.Request(tp))
+          taken <- dataQueue
+                     .takeBetween(1, Int.MaxValue)
+                     .race(interruptionPromise.await.as(Chunk.empty))
         } yield taken
 
       stream = ZStream.logAnnotate(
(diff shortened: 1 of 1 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -135,16 +135,18 @@ object PartitionStreamControl {
...
       now                 <- Clock.nanoTime
       queueInfo           <- Ref.make(QueueInfo(now, 0, None, 0))
       requestAndAwaitData =
         for {
-          _     <- commandQueue.offer(RunloopCommand.Request(tp))
-          _     <- diagnostics.emit(DiagnosticEvent.Request(tp))
-          taken <- dataQueue.takeBetween(1, Int.MaxValue)
+          _ <- commandQueue.offer(RunloopCommand.Request(tp))
+          _ <- diagnostics.emit(DiagnosticEvent.Request(tp))
+          taken <- dataQueue
```
svroonland, line 147: I'm surprised this compiles, a race between a (ZIO of) Chunk and Nothing. Does it work though? Maybe explicitly map it to a Chunk.empty

svroonland (approved):
This would indeed fix the situation as described in #1250.

So now the only remaining distinction between handling revoked and lost is that for lost we terminate immediately. Like you mentioned before, I think that lost partitions will not have received any data for a long time anyway. But since we can't test that, this is probably a safe implementation.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
erikvanoosten: Yes! It was `Unit` before, but `Nothing` is much better. There can be no value that is of type `Nothing`, that is the beauty, the compile knows that this await will never return a value, it can only fail. So no, no mapping to Chunk.empty is needed.

svroonland: Ah of course, I was mixing up the success and failure channels there. Still early morning for me 😅

erikvanoosten:
Note: @svroonland found a bug in this code after adding a unit test. See #1371.

PR #1252, opened 2024-06-10
A lost partition is no longer fatal

### LINKED ISSUE #1233: Consumer not reconnecting on lost session on proxy connectivity
This behaviour doesn't seem to apply when not using a kafka proxy. Connecting directly we observe the stream always recomposes itself.

Example code:

```scala
import org.apache.kafka.clients.consumer.ConsumerConfig
import zio._
import zio.kafka.consumer.{Consumer, ConsumerSettings, RebalanceListener, Subscription}
import zio.kafka.serde.{Deserializer, Serializer}

object ZioKafkaProxyTest extends ZIOAppDefault {

  private val Props = Map(
    ConsumerConfig.GROUP_ID_CONFIG                 -> "zio-kafka-test",
    ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG   -> "org.apache.kafka.common.serialization.StringDeserializer",
    ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG -> "org.apache.kafka.common.serialization.StringDeserializer",
    ConsumerConfig.AUTO_OFFSET_RESET_CONFIG        -> "earliest"
  )

    for {
      consumer <- Consumer.make(
        ConsumerSettings(List("kafka-proxy:9192"))
          .withProperties(Props)
          .withPollTimeout(1.second)
          .withRebalanceListener(RebalanceListener(
            (assigned, _) => ZIO.logInfo(s"assigned: ${assigned.map(_.partition())}"),
            (revoked, _) => ZIO.logInfo(s"revoked: ${revoked.map(_.partition())}"),
            (lost, _) => ZIO.logInfo(s"lost: ${lost.map(_.partition())}"),
          ))
      )
      _ <-
        consumer.plainStream(Subscription.topics("topic"), Deserializer.string, Serializer.string, 2)
          .mapZIO(m => ZIO.logInfo(m.value))
          .runDrain
    } yield ExitCode.success
  }

}
```
In version 2.3.2 it goes something like this:

```
timestamp=2024-04-30T16:58:09.273167Z level=INFO thread=#zio-fiber-19 message="assigned: Set(0)" 
proxy stopped
wait for session timeout
proxy start
timestamp=2024-04-30T16:58:40.757980Z level=INFO thread=#zio-fiber-90 message="lost: Set(0)" 
timestamp=2024-04-30T16:58:43.769893Z level=INFO thread=#zio-fiber-97 message="assigned: Set(0)"
stream resumes
```

In version 2.7.4 it goes:

```
timestamp=2024-04-30T15:26:46.500919Z level=INFO thread=#zio-fiber-18 message="assigned: Set(0)"
stopped proxy
wait for session timeout
started proxy
timestamp=2024-04-30T15:29:32.603663Z level=INFO thread=#zio-fiber-22 message="lost: Set(0)" 
hangs indefinitely
when stopping:
2024-04-30 16:31:57 ERROR ConsumerCoordinator:1201 - [Consumer clientId=consumer-zio-kafka-test-1, groupId=zio-kafka-test] LeaveGroup request with Generation{generationId=-1, memberId='consumer-zio-kafka-test-1-e2669255-c590-4ac7-9457-a585dfbf3969', protocol='null'} failed with error: The coordinator is not aware of this member.
```

erikvanoosten:
Before 2.7.0 a lost partition was treated as a revoked partition. Since the partition is already assigned to another node, this potentially leads to duplicate processing of records.

Zio-kafka 2.7.0 assumes that a lost partition is a fatal event. It leads to an interrupt in the stream that handles the partition. The other streams are ended, and the consumer closes with an error. Usually, a full program restart is needed to resume consuming.

It should be noted that stream processing is not interrupted immediately. Only when the stream requests new records, the interrupt is observed. Unfortunately, we have not found a clean way to interrupt the stream consumer directly.

Meanwhile, from bug reports (#1233, #1250), we understand that partitions are usually lost when no records have been received for a long time.

In conclusion, 1) it is not possible to immediately interrupt user stream processing, and 2) it is most likely not needed anyway because the stream is already done processing and awaiting more records.

With this change, a lost partition no longer leads to an interrupt. Instead, we first drain the stream's internal queue (just to be sure, it is probably already empty), and then we end the stream gracefully (that is, without error, like we do with revoked partitions). Other streams are not affected, the consumer will continue to work.

Lost partitions do not affect the features `rebalanceSafeCommits` and `restartStreamsOnRebalancing`; they do _not_ hold up a rebalance waiting for commits to complete, and they do _not_ lead to restarts of other streams.

Since we currently have no way to test lost partitions, there is no change to the tests.

Fixes #1233 and #1250.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
@@ -82,12 +82,6 @@ final class PartitionStreamControl private (
   private[internal] def maxPollIntervalExceeded(now: NanoTime): UIO[Boolean] =
     queueInfoRef.get.map(_.deadlineExceeded(now))
 
-  /** To be invoked when the partition was lost. */
-  private[internal] def lost: UIO[Boolean] = {
-    val lostException = new RuntimeException(s"Partition ${tp.toString} was lost") with NoStackTrace
-    interruptionPromise.fail(lostException)
-  }
-
   /** To be invoked when the stream is no longer processing. */
   private[internal] def halt: UIO[Boolean] = {
     val timeOutMessage = s"No records were polled for more than $maxPollInterval for topic partition $tp. " +
@@ -96,6 +90,14 @@ final class PartitionStreamControl private (
     val consumeTimeout = new TimeoutException(timeOutMessage) with NoStackTrace
     interruptionPromise.fail(consumeTimeout)
   }
+
+  /** To be invoked when the partition was lost. It clears the queue end ends the stream. */
+  private[internal] def lost: UIO[Unit] =
+    logAnnotate {
+      ZIO.logDebug(s"Partition ${tp.toString} lost") *>
+        dataQueue.takeAll *>
+        dataQueue.offer(Take.end).unit
+    }
 
   /** To be invoked when the partition was revoked or otherwise needs to be ended. */
   private[internal] def end: ZIO[Any, Nothing, Unit] =
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
@@ -91,7 +91,7 @@ final class PartitionStreamControl private (
     interruptionPromise.fail(consumeTimeout)
   }
 
-  /** To be invoked when the partition was lost. It clears the queue end ends the stream. */
+  /** To be invoked when the partition was lost. It clears the queue and ends the stream. */
   private[internal] def lost: UIO[Unit] =
     logAnnotate {
       ZIO.logDebug(s"Partition ${tp.toString} lost") *>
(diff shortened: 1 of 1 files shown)


erikvanoosten:
A note for reviewers: we assume that when a partition is lost, the underlying Java consumer no longer returns records for that partition.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -97,6 +91,14 @@ final class PartitionStreamControl private (
     interruptionPromise.fail(consumeTimeout)
   }
 
+  /** To be invoked when the partition was lost. It clears the queue and ends the stream. */
+  private[internal] def lost: UIO[Unit] =
+    logAnnotate {
+      ZIO.logDebug(s"Partition ${tp.toString} lost") *>
```
svroonland, line 97: Can we log how many records were taken and discarded?

erikvanoosten: Fixed in https://github.com/zio/zio-kafka/pull/1252/commits/ff8e92d2e99b120b1df8975dbaee8abca9ed9b93.

svroonland (approved):
I think this is a usability improvement in the situations that we have identified.

We should mention in the release notes that clients that want to handle the partition lost event somehow, instead of handling the failed stream they need to create their own RebalanceListener and handle the `onLost` call.

Issue #1239, opened 2024-05-09
Partition handling with plainStream and partitionedStream


petern-sc:
Hi, I was tracking down a performance issue with our Kafka consumer and I noticed the difference came from the usage of `plainStream` an `partitionedStream`.

We originally had something along the lines of:
```scala
val maxParallelism = 5
Consumer
    .partitionedStream(Subscription.topics(topicName), Serde.byteArray.asOption, Serde.byteArray.asOption)
    .flatMapPar(maxParallelism) { case (_, stream) => stream.mapZIO(handleRecord) }
    .aggregateAsync(Consumer.offsetBatches)
    .schedule(Schedule.fixed(2.seconds))
    .run(ZSink.foreach(_.commit))
```
Which resulted in very low throughput and the consumer stalling.

When using plainStream, with all else being equal, results in going back to "normal" performance.
```scala
val maxParallelism = 5
Consumer
    .plainStream(Subscription.topics(topicName), Serde.byteArray.asOption, Serde.byteArray.asOption)
    .mapZIOPar(maxParallelism)(handleRecord)
    .aggregateAsync(Consumer.offsetBatches)
    .schedule(Schedule.fixed(2.seconds))
    .run(ZSink.foreach(_.commit))
```

Looking at the implementations, the only difference I can see is that `plainStream` has infinite parallelism when pulling from `partitionedStream`. However I was expected that as a pull stream, when I call `.mapZIOPar(maxParallelism)` this would still bottleneck the records that we pull from `plainStream` so that the performance would be similar.

Is there something I'm misunderstanding in how the 2 methods differ?

Minimal repo to reproduce this: https://github.com/petern-sc/kafka-perf

---
Update:

I realised this wasn't a performance issue but a logic issue. 

With a parallelism of 5, only 5 partitions were being consumed from. The consumer never yields and begins consuming from the other partitions causing Kafka to eventually evict the consumer. Is this behaviour expected from `partitionedStream` ie. partitionedStreams always requires parallelism >= number of partitions for the consumer.

```
kafka-consumer-groups --group test-consumer-1715230505326 --describe --bootstrap-server localhost:9091

GROUP                       TOPIC           PARTITION  CURRENT-OFFSET  LOG-END-OFFSET  LAG             CONSUMER-ID                               HOST            CLIENT-ID
test-consumer-1715230505326 test-topic-1    38         272832          272832          0               test-e0c3645d-2794-4d6f-9246-37e7f3782735 /192.168.65.1   test
[...]

PR #1269, opened 2024-07-02
Update consumer docs

petern-sc:
- Mention the importance of using Int.MaxValue or similar when using partitionedStream with flatMapPar, otherwise it will lead to a hung consumer that gets evicted by Kafka.

Addresses https://github.com/zio/zio-kafka/issues/1239


erikvanoosten (approved):

erikvanoosten:
Thanks @petern-sc !

PR #1270, opened 2024-07-04
 Allow for some auth exceptions during poll

erikvanoosten:
Some brokers are sometimes too slow to authorize or authenticate a poll. This results in spurious exceptions. With this change we continue polling unless the error happens too often.

We test the change with a new type of unit test for Runloop.

### DIFF
--- a/zio-kafka-test/src/test/scala/zio/kafka/consumer/internal/RunloopSpec.scala
+++ b/zio-kafka-test/src/test/scala/zio/kafka/consumer/internal/RunloopSpec.scala
@@ -2,12 +2,12 @@ package zio.kafka.consumer.internal
 
 import org.apache.kafka.clients.consumer.{ ConsumerRecord, MockConsumer, OffsetResetStrategy }
 import org.apache.kafka.common.TopicPartition
-import org.apache.kafka.common.errors.AuthorizationException
+import org.apache.kafka.common.errors.{ AuthenticationException, AuthorizationException }
 import zio._
 import zio.kafka.consumer.{ ConsumerSettings, Subscription }
 import zio.kafka.consumer.diagnostics.Diagnostics
 import zio.kafka.consumer.internal.RunloopAccess.PartitionAssignment
-import zio.metrics.MetricState
+import zio.metrics.{ MetricState, Metrics }
 import zio.stream.{ Take, ZStream }
 import zio.test.TestAspect.withLiveClock
 import zio.test._
@@ -35,7 +35,7 @@ object RunloopSpec extends ZIOSpecDefault {
           }
           for {
             streamStream <- ZStream.fromHubScoped(partitionsHub)
-            _            <- runloop.addSubscription(Subscription.Topics(Set("t1")))
+            _            <- runloop.addSubscription(Subscription.Topics(Set(tp10.topic())))
             record <- streamStream
                         .map(_.exit)
                         .flattenExitOption
@@ -52,7 +52,7 @@ object RunloopSpec extends ZIOSpecDefault {
           )
         }
       },
-      test("runloop retries poll upon AuthorizationException") {
+      test("runloop retries poll upon AuthorizationException and AuthenticationException") {
         withRunloop { (mockConsumer, partitionsHub, runloop) =>
           mockConsumer.schedulePollTask { () =>
             mockConsumer.updateEndOffsets(Map(tp10 -> Long.box(0L)).asJava)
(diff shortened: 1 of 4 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -33,7 +30,8 @@ final case class ConsumerSettings(
   maxRebalanceDuration: Option[Duration] = None,
   fetchStrategy: FetchStrategy = QueueSizeBasedFetchStrategy(),
   metricLabels: Set[MetricLabel] = Set.empty,
-  runloopMetricsSchedule: Schedule[Any, Unit, Long] = Schedule.fixed(500.millis)
+  runloopMetricsSchedule: Schedule[Any, Unit, Long] = Schedule.fixed(500.millis),
+  pollAuthErrorRetries: Int = 5
```
svroonland, line 34: Do these errors, when occurring intermittent, occur after the poll timeout or immediately? If the former, a number of retries schedule would be sufficient, without the delay.

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -33,7 +30,8 @@ final case class ConsumerSettings(
   maxRebalanceDuration: Option[Duration] = None,
   fetchStrategy: FetchStrategy = QueueSizeBasedFetchStrategy(),
   metricLabels: Set[MetricLabel] = Set.empty,
-  runloopMetricsSchedule: Schedule[Any, Unit, Long] = Schedule.fixed(500.millis)
+  runloopMetricsSchedule: Schedule[Any, Unit, Long] = Schedule.fixed(500.millis),
```
svroonland, line 33: We could also have user pass a custom Schedule object.

zio-kafka-test/src/test/scala/zio/kafka/consumer/internal/RunloopSpec.scala
```diff
@@ -0,0 +1,127 @@
...
+import zio.kafka.consumer.diagnostics.Diagnostics
+import zio.kafka.consumer.internal.RunloopAccess.PartitionAssignment
+import zio.metrics.{ MetricState, Metrics }
+import zio.stream.{ Take, ZStream }
+import zio.test.TestAspect.withLiveClock
+import zio.test._
+
+import scala.jdk.CollectionConverters._
+
+object RunloopSpec extends ZIOSpecDefault {
```
svroonland, line 17: Nice!

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
erikvanoosten: Given that:
1. authentication/authorization should be the first thing the broker does,
2. the error happens because the request from the broker to the auth-provider times-out,
the error happens the broker's time out setting after the poll. What this setting would be, 🤷

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
erikvanoosten: Ah, you mean just like `runloopMetricsSchedule`?

erikvanoosten: I like this, more flexible. I'll make the chance.

erikvanoosten: Done in https://github.com/zio/zio-kafka/pull/1270/commits/0dbe30a03df9a07dcfd62796ed7964adf00bcda4

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -300,6 +298,22 @@ final case class ConsumerSettings(
...
+   *   [[org.apache.kafka.common.errors.AuthenticationException]].
+   *
+   * This setting helps with failed polls due to too slow authorization or authentication in the broker. You may also
+   * consider increasing `pollTimeout` to reduce auth-work on the broker.
+   *
+   * Set to `Schedule.stop` to fail the consumer on the first auth error.
+   *
+   * The default is {{{Schedule.recurs(5) && Schedule.spaced(500.millis)}}} which is, to retry 5 times, spaced by 500ms.
+   */
+  def withPollAuthErrorRetrySchedule(pollAuthErrorRetrySchedule: Schedule[Any, Throwable, Any]): ConsumerSettings =
```
svroonland, line 314: Nitpicking, but do we need the 'poll' prefix here?

svroonland (approved):
One tiny thing, but LGTM!

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
erikvanoosten: Good question. My initial answer was yes, because we'll do the same for the producer. But that already has its own settings object... I will remove the prefix 😄

erikvanoosten: Fixed in https://github.com/zio/zio-kafka/pull/1270/commits/e7711b531b4500adddf52fb9f0dd9f059c360e0d

PR #1272, opened 2024-07-10
Small improvements to the Producer

erikvanoosten:
By using ZIO.async, we no longer need a reference to the zio runtime, nor do we need the `exec` trickery anymore.


svroonland (approved):
Much better this way. I think this code stems from very early ZIO days, possibly there was some reason back then not to use ZIO.async, but I can't think of any.

PR #1281, opened 2024-07-14
Add alternative fetch strategy for many partitions

erikvanoosten:
When many hundreds of partitions need to be consumed, an excessive amount of heap can be used for pre-fetching. The  `ManyPartitionsQueueSizeBasedFetchStrategy` works similarly as the default `QueueSizeBasedFetchStrategy` but limits total memory usage.


erikvanoosten:
This `ManyPartitionsQueueSizeBasedFetchStrategy` has been run in production for a half a year now and works as expected. It was split of from #970.

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/ManyPartitionsQueueSizeBasedFetchStrategy.scala
```diff
@@ -0,0 +1,51 @@
...
+ */
+final case class ManyPartitionsQueueSizeBasedFetchStrategy(
+  maxPartitionQueueSize: Int = 1024,
+  maxTotalQueueSize: Int = 20480
+) extends FetchStrategy {
+  override def selectPartitionsToFetch(
+    streams: Chunk[PartitionStream]
+  ): ZIO[Any, Nothing, Set[TopicPartition]] = {
+    // By shuffling the streams we prevent read-starvation for streams at the end of the list.
+    val shuffledStreams = scala.util.Random.shuffle(streams)
```
svroonland, line 39: Would it be better for testing purposes to use ZIO's random functionality?

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/ManyPartitionsQueueSizeBasedFetchStrategy.scala
```diff
@@ -0,0 +1,51 @@
...
+ * consumed.
+ *
+ * @param maxPartitionQueueSize
+ *   Maximum number of records to be buffered per partition. This buffer improves throughput and supports varying
+ *   downstream message processing time, while maintaining some backpressure. Large values effectively disable
+ *   backpressure at the cost of high memory usage, low values will effectively disable prefetching in favour of low
+ *   memory consumption. The number of records that is fetched on every poll is controlled by the `max.poll.records`
+ *   setting, the number of records fetched for every partition is somewhere between 0 and `max.poll.records`.
+ *
+ * The default value for this parameter is 2 * the default `max.poll.records` of 500, rounded to the nearest power of 2.
```
svroonland, line 22: Can it be set to a very high value to effectively only limit by total queue size?

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/ManyPartitionsQueueSizeBasedFetchStrategy.scala
```diff
@@ -0,0 +1,51 @@
...
+ * pre-fetching in each poll.
+ */
+final case class ManyPartitionsQueueSizeBasedFetchStrategy(
+  maxPartitionQueueSize: Int = 1024,
+  maxTotalQueueSize: Int = 20480
+) extends FetchStrategy {
+  override def selectPartitionsToFetch(
+    streams: Chunk[PartitionStream]
+  ): ZIO[Any, Nothing, Set[TopicPartition]] = {
+    // By shuffling the streams we prevent read-starvation for streams at the end of the list.
```
svroonland, line 38: Is shuffling more fair than for example sorting by queue size, so that those with no buffered records can fetch first?

svroonland (commented):
Nice offering besides the existing fetch strategy.

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/ManyPartitionsQueueSizeBasedFetchStrategy.scala
erikvanoosten: That is pretty smart. The only problem I can think of is that it will go wrong when many partitions have the same queue size. This is somewhat likely for queue size 0 (but then its not an issue because the sum is still 0), and luckily a lot less likely for larger queue sizes. Shuffling does not have this issue. After the things we had to roll back last year, I don't feel comfortable with experimenting with this.

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/ManyPartitionsQueueSizeBasedFetchStrategy.scala
erikvanoosten: Perhaps. With sufficient property based tests that is not a hard necessity though. Having said that, there is currently no test that checks that the order is different every time.
A bigger problem is that I could not find a way to efficiently suffle an array with zio's random. Do you know of one?

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/ManyPartitionsQueueSizeBasedFetchStrategy.scala
erikvanoosten: Yes, that should be possible.

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/ManyPartitionsQueueSizeBasedFetchStrategy.scala
erikvanoosten: Just thought of something much worse. Ordering by queue size will go horribly wrong in the following case. Say there are 1000 partitions, half of those receive just 10 messages/s, the other half receives 200 messages/s. In this situation the lower throughput partitions will always take all the available slots and the higher throughput partitions will experience read starvation.

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/ManyPartitionsQueueSizeBasedFetchStrategy.scala
erikvanoosten: Added something to that extend in the scaladoc.

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/ManyPartitionsQueueSizeBasedFetchStrategy.scala
erikvanoosten: Okay, that was actually pretty easy. It is now using ZIO.random.

svroonland (approved):

zio-kafka/src/main/scala/zio/kafka/consumer/fetch/ManyPartitionsQueueSizeBasedFetchStrategy.scala
```diff
@@ -0,0 +1,59 @@
...
+ * @param maxTotalQueueSize
+ *   Maximum number of records to be buffered over all partitions together. This can be used to limit memory usage when
+ *   consuming a large number of partitions.
+ *
+ * When multiple streams are eligible for pre-fetching (because their queue size is below `maxPartitionQueueSize`), but
+ * together they exceed `maxTotalQueueSize`, then every call a random set of eligible streams is selected that stays
+ * below `maxTotalQueueSize`. The randomization ensures fairness and prevents read-starvation for streams at the end of
+ * the list.
+ *
+ * The default value is 20 * the default for `maxTotalQueueSize`, allowing approximately 20 partitions to do
```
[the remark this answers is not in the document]
erikvanoosten: 👍  Fixed in https://github.com/zio/zio-kafka/pull/1281/commits/bdbffd57498deafc81564bc57ae32f4dc1746280

PR #1283, opened 2024-07-16
producer: allow to retry on auth failures

domdorn:


erikvanoosten:
This won't work. This only looks at the first record, if that one fails all records are retried.
A full solution will support retry of some of the records, potentially having retry sending a smaller set of records every retry. There is no quick fix, unfortunately solving this properly will require some more smarts.

erikvanoosten:
Closed in favor of #1284.

erikvanoosten:
Zio-kafka supports this now. For more details see #1437.

PR #1285, opened 2024-07-17
Alternative producer implementation

erikvanoosten:
Refactoring of the producer so that it handles errors per record.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
@@ -1,6 +1,8 @@
 package zio.kafka.producer
 
+import org.apache.kafka.clients.producer
 import org.apache.kafka.clients.producer.{ KafkaProducer, Producer => JProducer, ProducerRecord, RecordMetadata }
+import org.apache.kafka.common.errors.{ AuthenticationException, AuthorizationException }
 import org.apache.kafka.common.serialization.ByteArraySerializer
 import org.apache.kafka.common.{ Metric, MetricName, PartitionInfo }
 import zio._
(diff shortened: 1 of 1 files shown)


zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -457,53 +460,63 @@ private[producer] final class ProducerLive(
...
+    for {
+      done <- Promise.make[Throwable, RecordMetadata]
+      _ <- ZIO
+             .attempt[Any] {
+               p.send(
+                 record,
+                 (metadata: RecordMetadata, err: Exception) =>
+                   unsafeRun {
+                     if (err == null) done.succeed(metadata)
+                     else done.fail(err)
```
domdorn, line 513: if we change this to `done.fail((record,err))` we could implement a retry strategy in 494

erikvanoosten: I like the idea. But are you sure you want to do the retries record by record?

svroonland (approved):

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
domdorn: I think that is the only way to not deliver messages that have been previously accepted, e.g. prevent double delivery?

erikvanoosten: The other option is to do something after line 489. There you can `records.zip(results)` and filter those that failed. The failed records can then be retried together. The challenge is mostly to write elegant code that merges the original results with the retry results.

PR #1286, opened 2024-07-18
Retry send records after auth exception

erikvanoosten:
Some brokers are sometimes too slow to authorize or authenticate a send. This results in spurious exceptions. With this change we retry sends unless the error happens too often.

Still todo:
- [ ] make it a bit more efficient for the success case (e.g. only create schedule driver when necessary)
- [ ] make retry setup more efficient (loop through only once to collect firstAuthError, authErrorCount and errorCount
- [ ] add tests
- [ ] fix comment typo `firs auth error` => `first auth error`


erikvanoosten:
Replaced by #1437 .

PR #1287, opened 2024-07-18
Alive checks

domdorn:
hooks from outside to check if a producer or consumer is "still alive", e.g. a consumers runloop is still active or a producers sendQueue hasn't finished. 

I'm going to use these in my applications to provide health-checks exposed to kubernetes, which will in turn restart the app as a whole when these checks fail (or another health-check of another component in my app)

### DIFF
--- a/zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerAliveSpec.scala
+++ b/zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerAliveSpec.scala
@@ -1,0 +1,129 @@
+package zio.kafka.consumer
+
+//import io.github.embeddedkafka.EmbeddedKafka
+//import org.apache.kafka.clients.consumer.ConsumerConfig
+//import org.apache.kafka.clients.consumer.ConsumerPartitionAssignor
+//import org.apache.kafka.clients.consumer.CooperativeStickyAssignor
+//import org.apache.kafka.clients.consumer.MockConsumer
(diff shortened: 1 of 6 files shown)


jivanic-demystdata:
The PR needs some tests

domdorn:
Hi @svroonland,

Thanks for replying to this PR, which I will close with this comment.

Unfortunately, my team lost trust in zio-kafka and I got outvoted, so we are now wrapping spring-kafka. 

The issue was that zio-kafka does not handle all the exceptions documented in Kafkas producer/consumer. E.g. a exception in Kafkas NetworkClient could bubble up and kill the runloop, leaving the app in a state where the stream never finishes but also not fails. 

Together with Consumer/Producers that are created in Layers (with the provided helpers of zio-kafka) this left the application in a state where no events where processed anymore but we could also not detect/recover from this state as no errors where emitted. 
The health checks where an attempt to detect these errors, e.g. additional to the code in this PR, we created a component + topic where we regularly were posting to and reading from and when the time between message creation and consumption differed too much, we'd know that the runloop died again and restart the app.

Unfortunately, this was also kind of flaky, product owners wanted assurances that this "kafka stuff now finally works" and other teams had solutions with java+spring-kafka that ran 24/7 without problems.

Thanks for spending time reviewing this PR and contributing to open source!

erikvanoosten:
Thanks for your message @domdorn. Very valuable feedback. It seems we need to study what spring-kafka is doing more closely. BTW, for your next project, please know that v2.8.3 fixed a bug that could cause the consumer to halt without error after a rebalance.

PR #1298, opened 2024-08-14
Do not start streams for immediately revoked/lost partitions

erikvanoosten:
As discovered by @ytalashko (see https://github.com/zio/zio-kafka/pull/1294), it is possible that a partition is assigned and immediately revoked in the same poll. No stream should be started for these partitions.

Unlike #1294, this PR does _not_ hide the assigned+revoked partitions from the diagnostic events and rebalance metrics. In addition, this PR also supports the unlikely event of a assigned+lost partition.


svroonland:
I thought about solving it this way as well, but I figured the semantics of RebalanceEvent made more sense in the other way, where in a single RebalanceEvent partitions can either be assigned or revoked, not both. Which semantics are desirable for the metrics, I don't know. 

Since I have no strong preference for either direction, this PR is fine for me as well.

erikvanoosten:
As @ytalashko points out in https://github.com/zio/zio-kafka/pull/1294#issuecomment-2289793056, this PR has a bug when a partition is first revoked and then assigned (in the same poll). Closing!

PR #1304, opened 2024-08-21
Revert back to 2.8.0 sendFromQueue implementation in producer

ytalashko:
During running services with the new version of library 2.8.1, I noticed huge increase in messages production time to kafka.
Some quick rough tests shoving me around 40x-100x times increase in amount of time taken to `produceChunk` on even 1-10 records chunks comparing to version 2.8.0 (or also to version 2.8.1 with the changes in this MR). Please, note, it is not a proper benchmarks. Also, to note, Im using Scala version 3.3.1.

This MR just reverts two MRs updating `ProducersendFromQueue` implementation:
 - https://github.com/zio/zio-kafka/pull/1272
 - https://github.com/zio/zio-kafka/pull/1285
 
 Not sure if it's possible to revert two MRs at a time (with a single one for revert), so created this one.
 
 I haven't researched yet which exact change/changes are causing such performance degradation.
 I would suggest the next steps:
  - confirm the problem exists
  - reverting to the previous implementation (the one from 2.8.0/this MR)
  - release fixed version (to allow users have a nicely working version)
  - investigate & fix problem from the https://github.com/zio/zio-kafka/pull/1272 and/or https://github.com/zio/zio-kafka/pull/1285

It is only suggestions on the approach, feel free to ignore them.
Also, feel free to modify/ignore this MR and treat it as an issue.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
@@ -8,7 +8,9 @@ import zio.kafka.serde.Serializer
 import zio.kafka.utils.SslHelper
 import zio.stream.{ ZPipeline, ZStream }
 
+import java.util.concurrent.atomic.AtomicLong
 import scala.jdk.CollectionConverters._
+import scala.util.control.NonFatal
 
 trait Producer {
 
@@ -218,12 +220,13 @@ object Producer {
     settings: ProducerSettings
   ): ZIO[Scope, Throwable, Producer] =
     for {
+      runtime <- ZIO.runtime[Any]
       sendQueue <-
         Queue.bounded[(Chunk[ByteRecord], Promise[Nothing, Chunk[Either[Throwable, RecordMetadata]]])](
           settings.sendBufferSize
         )
-      producer = new ProducerLive(javaProducer, sendQueue)
-      _ <- producer.sendFromQueue.forkScoped
+      producer = new ProducerLive(javaProducer, runtime, sendQueue)
+      _ <- ZIO.blocking(producer.sendFromQueue).forkScoped
     } yield producer
 
   /**
@@ -359,6 +362,7 @@ object Producer {
 
 private[producer] final class ProducerLive(
   private[producer] val p: JProducer[Array[Byte], Array[Byte]],
+  runtime: Runtime[Any],
   sendQueue: Queue[(Chunk[ByteRecord], Promise[Nothing, Chunk[Either[Throwable, RecordMetadata]]])]
 ) extends Producer {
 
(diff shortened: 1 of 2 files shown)


erikvanoosten (approved):

svroonland:
Results on my macbook, with the benchmark from #1310.

Implementation before producer changes:
```
Benchmark                               Mode  Cnt     Score    Error  Units
[info] ProducerBenchmark.produceSingleRecord  thrpt   25    45,083 ±  1,819  ops/s
[info] ProducerBenchmark.produceChunk          avgt   25  2150,752 ± 85,044  ms/op
```

After producer changes (which were reverted in this PR):
```
[info] Benchmark                               Mode  Cnt     Score    Error  Units
[info] ProducerBenchmark.produceSingleRecord  thrpt   25    46,079 ±  1,268  ops/s
[info] ProducerBenchmark.produceChunk          avgt   25  2178,640 ± 73,707  ms/op
```

erikvanoosten:
@svroonland is a bit limited as it only produces 1 chunk per benchmark.

@ytalashko can you share a bit about the messages you produce? E.g. size per message, number of messages per batch,  throughput, number of topics, does a batch have a single topic or multiple, do you have compression enabled, does the data compress well, etc. No need to be precise, ballpark numbers are fine.

ytalashko:
> Results on my macbook, with the benchmark from https://github.com/zio/zio-kafka/pull/1310.

Interesting, the results are close.

> can you share a bit about the messages you produce?

Sure. Those are approximations from the service experienced the issue, the first and only one I updated to the version 2.8.1:
 - Messages are vary in size, and most should be from 2 to 20 KBytes uncompressed.
 - We do use compression (snappy), which decreases the size of message at least 3-4 times.
 - Batch size is also vary in size (from like 5 to 200 approximately), aggregated within 100 ms
 - Throughput is from like 50/100 to 1000+ messages per second per partition.
 - Each service instance typically handles around 5-7 partitions (from two topics) in parallel, e.g. stream per partition.
 - There is one producer per service instance (having `.withSendBufferSize(16384)`), and per each partition there can be up to two batches in progress (due to the `produceChunkAsync` + `.mapZIOPar(1)`), so typically like up to 14 batches at a time.

erikvanoosten:
Thanks @ytalashko.

Perhaps due to a tiny optimization, #1311 is faster (on github's machines) than 2.8.2. (See the last graph on https://zio.github.io/zio-kafka/dev/bench/.) @svroonland would you like to compare #1311 against master and 2.8.1 as well?

erikvanoosten:
Nope, scratch that, to much variance to say that #1311 is faster. A re-run shows that #1311 and master are comparable with a bit more variation (so far) for #1311.

svroonland:
It could be changes in ZIO (streams) that affect performance here. @ytalashko Could that be the case in your situation and could you try to revert to an previous ZIO version and see what performance is like?

ytalashko:
> It could be changes in ZIO (streams) that affect performance here. @ytalashko Could that be the case in your situation and could you try to revert to an previous ZIO version and see what performance is like?

Upd: checked with previous ZIO version (`2.1.6`), the results remain the same

PR #1310, opened 2024-08-24
Simple producer benchmark

svroonland:
It's not much, but it gives us something to compare against.

Relates to #1304

### DIFF
--- a/zio-kafka-bench/src/main/scala/zio/kafka/bench/ProducerBenchmark.scala
+++ b/zio-kafka-bench/src/main/scala/zio/kafka/bench/ProducerBenchmark.scala
@@ -16,8 +16,9 @@ import java.util.concurrent.TimeUnit
 class ProducerBenchmark extends ZioBenchmark[Kafka with Producer] {
   val topic1                      = "topic1"
   val nrPartitions                = 6
-  val nrMessages                  = 50000
+  val nrMessages                  = 5000
   val kvs: List[(String, String)] = List.tabulate(nrMessages)(i => (s"key$i", s"msg$i"))
+  val records                     = Chunk.fromIterable(kvs.map { case (k, v) => new ProducerRecord(topic1, k, v) })
 
   override protected def bootstrap: ZLayer[Any, Nothing, Kafka with Producer] =
     ZLayer.make[Kafka with Producer](Kafka.embedded, producer).orDie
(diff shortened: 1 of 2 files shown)


zio-kafka-bench/src/main/scala/zio/kafka/bench/ProducerBenchmark.scala
```diff
@@ -0,0 +1,51 @@
...
+
+import java.util.concurrent.TimeUnit
+
+@State(Scope.Benchmark)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+class ProducerBenchmark extends ZioBenchmark[Kafka with Producer] {
+  val topic1                      = "topic1"
+  val nrPartitions                = 6
+  val nrMessages                  = 50000
+  val kvs: List[(String, String)] = List.tabulate(nrMessages)(i => (s"key$i", s"msg$i"))
```
erikvanoosten, line 20: Nice, small messages makes us able to test worst case. Still, 50k small messages is not a lot of traffic yet. I propose we increase that to at least 500k.

zio-kafka-bench/src/main/scala/zio/kafka/bench/ProducerBenchmark.scala
```diff
@@ -0,0 +1,51 @@
...
+  } yield ()
+
+  @Benchmark
+  @BenchmarkMode(Array(Mode.AverageTime))
+  def produceChunk(): Any = runZIO {
+    for {
+      _ <- Producer
+             .produceChunk[Any, String, String](
+               Chunk.fromIterable(kvs.map { case (k, v) =>
+                 new ProducerRecord(topic1, k, v)
```
erikvanoosten, line 37: Can we move the construction of `ProducerRecord`s outside the benchmark?

erikvanoosten (commented):
Great start. I propose we:
* move construction of ProducerRecord to outside the benchmark
* in the benchmark we publish 100 chunks of 5000 small messages

WDYT?

erikvanoosten (approved):
I just realized that there is one super-tiny thing we can improve: use a binary key/value. Because we now using string, there is the tiny overhead of converting string to byte array. I am not sure if you want to put in the effort, and this is most probably already good.

erikvanoosten:
FYI, I have removed the first run of this new bench from the charts at https://zio.github.io/zio-kafka/dev/bench/ because it was producing a lot less data and is therefore not comparable with later runs.

svroonland:
> I just realized that there is one super-tiny thing we can improve: use a binary key/value. Because we now using string, there is the tiny overhead of converting string to byte array. I am not sure if you want to put in the effort, and this is most probably already good.


I think it doesn't really matter since this isn't currently a benchmark about how much performance we can squeeze out of the Producer. It's just to have a reference for future changes.

erikvanoosten:
Your right, good point 👍

svroonland:
Would be nice to have a "max performance" benchmark though, using the *async methods, batching and clever pipelining.

PR #1315, opened 2024-08-25
Use AtomicInteger instead of AtomicLong inside Producer.sendFromQueue implementation

ytalashko:
Minor change (alignment), in case we gonna keep current `Producer.sendFromQueue` implementation (https://github.com/zio/zio-kafka/pull/1311#discussion_r1730449835), otherwise, this PR can be closed.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
@@ -8,7 +8,7 @@ import zio.kafka.serde.Serializer
 import zio.kafka.utils.SslHelper
 import zio.stream.{ ZPipeline, ZStream }
 
-import java.util.concurrent.atomic.AtomicLong
+import java.util.concurrent.atomic.AtomicInteger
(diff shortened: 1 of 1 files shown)


zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -471,8 +471,8 @@ private[producer] final class ProducerLive(
             val it: Iterator[(ByteRecord, Int)] = serializedRecords.iterator.zipWithIndex
             val res: Array[Either[Throwable, RecordMetadata]] =
               new Array[Either[Throwable, RecordMetadata]](serializedRecords.length)
-            val count: AtomicLong = new AtomicLong
-            val length            = serializedRecords.length
+            val count: AtomicInteger = new AtomicInteger
+            val length               = serializedRecords.length
```
ytalashko, line 475: `AtomicInteger` is sufficient, since `serializedRecords.length` is `Int`

svroonland (approved):

PR #1321, opened 2024-08-31
More precise batch publish errors

ytalashko:
This increases the precision of the result of producer method `produceChunkAsyncWithFailures` for cases where something went wrong. There are two changes compared to before:

1. Each entry in the result of`produceChunkAsyncWithFailures` now accurately corresponds to each record in the input chunk. Previously, if sending fails directly (*) on any of the given records, the error would be used for _all_ records in the batch, ignoring the send-outcome of the other records. An advantage of this change is that if sending some records failed, but some other records were actually sent, you can now correctly see all of that in the method's response.
2. In addition, if sending fails directly for a record (*), we no longer attempt to send subsequent records from the input. The result contains a `PublishOmittedException` for each record that is not sent. When implementing retries, this change makes it easier to publish records in the original order.

In addition, we introduce unit-level tests for the producer.

(*) By 'sending' we mean offering a record to the underlying java Kafka producer. Sending can fail directly (when we call the method), or later on (from a callback).

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
@@ -8,9 +8,8 @@ import zio.kafka.serde.Serializer
 import zio.kafka.utils.SslHelper
 import zio.stream.{ ZPipeline, ZStream }
 
-import java.util.concurrent.atomic.AtomicInteger
 import scala.jdk.CollectionConverters._
-import scala.util.control.NonFatal
+import scala.util.control.{ NoStackTrace, NonFatal }
 
 trait Producer {
 
@@ -187,6 +186,10 @@ trait Producer {
 }
 
 object Producer {
+  case object SendOmittedDueToPreviousRecordSendCallFailureError
+      extends RuntimeException("Send omitted due to the previous record send call failure")
+      with NoStackTrace
+
   val live: RLayer[ProducerSettings, Producer] =
     ZLayer.scoped {
       for {
@@ -466,41 +469,72 @@ private[producer] final class ProducerLive(
     ZStream
       .fromQueueWithShutdown(sendQueue)
       .mapZIO { case (serializedRecords, done) =>
-        ZIO.succeed {
-          try {
-            val it: Iterator[(ByteRecord, Int)] = serializedRecords.iterator.zipWithIndex
-            val res: Array[Either[Throwable, RecordMetadata]] =
-              new Array[Either[Throwable, RecordMetadata]](serializedRecords.length)
-            val count: AtomicInteger = new AtomicInteger
-            val length               = serializedRecords.length
+        ZIO.suspendSucceed {
+          val recordsLength                                = serializedRecords.length
+          val recordsIterator: Iterator[(ByteRecord, Int)] = serializedRecords.iterator.zipWithIndex
+          val sentResults: Array[Either[Throwable, RecordMetadata]] =
+            new Array[Either[Throwable, RecordMetadata]](recordsLength)
 
-            while (it.hasNext) {
-              val (rec, idx): (ByteRecord, Int) = it.next()
-
-              val _ = p.send(
-                rec,
-                (metadata: RecordMetadata, err: Exception) =>
-                  Unsafe.unsafe { implicit u =>
-                    exec {
-                      if (err != null) res(idx) = Left(err)
-                      else res(idx) = Right(metadata)
-
-                      if (count.incrementAndGet == length) {
-                        exec {
-                          runtime.unsafe.run(done.succeed(Chunk.fromArray(res))).getOrThrowFiberFailure()
-                        }
-                      }
-                    }
-                  }
-              )
-            }
-          } catch {
-            case NonFatal(e) =>
+          Ref.make(0).map { sentRecordsCountRef =>
+            @inline def safelyInsertSentResult(resultIndex: Int, sentResult: Either[Throwable, RecordMetadata]): Unit =
               Unsafe.unsafe { implicit u =>
                 exec {
-                  runtime.unsafe.run(done.succeed(Chunk.fill(serializedRecords.size)(Left(e)))).getOrThrowFiberFailure()
+                  runtime.unsafe.run(
+                    sentRecordsCountRef.update { sentRecordsCount =>
+                      // Updating sentResults[resultIndex] here is safe:
+                      //  - Ref.update guarantees sentResults.update executed atomically
+                      //  - Ref.update starts with volatile variable read and ends with volatile variable write,
+                      //    which guarantees sentResults.update executed on the latest updated version of sentResults
+                      //    and currently updated version of sentResults
+                      //    will be visible to the next sentResults read or update called within Ref.update
+                      sentResults.update(resultIndex, sentResult)
+
+                      val newSentRecordsCount = sentRecordsCount + 1
+                      if (newSentRecordsCount == recordsLength) {
+                        val sentResultsChunk = Chunk.fromArray(sentResults)
+
+                        exec {
+                          runtime.unsafe.run(done.succeed(sentResultsChunk))
+                        }
+                      }
+
+                      newSentRecordsCount
+                    }
+                  )
                 }
               }
+
+            var previousSendCallSucceed = true
+
+            while (recordsIterator.hasNext) {
+              val (record: ByteRecord, recordIndex: Int) = recordsIterator.next()
+
+              if (previousSendCallSucceed) {
+                try {
+                  val _ = p.send(
+                    record,
+                    (metadata: RecordMetadata, err: Exception) =>
+                      safelyInsertSentResult(
+                        recordIndex,
+                        if (err eq null) Right(metadata) else Left(err)
+                      )
+                  )
+                } catch {
+                  case NonFatal(err) =>
+                    previousSendCallSucceed = false
+
+                    safelyInsertSentResult(
+                      recordIndex,
+                      Left(err)
+                    )
+                }
+              } else {
+                safelyInsertSentResult(
+                  recordIndex,
+                  Left(Producer.SendOmittedDueToPreviousRecordSendCallFailureError)
+                )
+              }
+            }
           }
         }
       }
--- a/zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
@@ -186,8 +186,8 @@ trait Producer {
 }
 
 object Producer {
-  case object SendOmittedException
-      extends RuntimeException("Send omitted due to a send error for a previous record in the chunk")
+  case object PublishOmittedException
+      extends RuntimeException("Publish omitted due to a publish error for a previous record in the chunk")
       with NoStackTrace
 
   val live: RLayer[ProducerSettings, Producer] =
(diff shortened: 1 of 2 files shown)


zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -187,6 +186,10 @@ trait Producer {
 }
 
 object Producer {
+  case object SendOmittedDueToPreviousRecordSendCallFailureError
```
ytalashko, line 189: Name suggestions are appreciated

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -466,41 +469,72 @@ private[producer] final class ProducerLive(
...
-                rec,
-                (metadata: RecordMetadata, err: Exception) =>
-                  Unsafe.unsafe { implicit u =>
-                    exec {
-                      if (err != null) res(idx) = Left(err)
-                      else res(idx) = Right(metadata)
-
-                      if (count.incrementAndGet == length) {
+        ZIO.suspendSucceed {
+          val recordsLength                                = serializedRecords.length
```
ytalashko, line 473: Noticed during changes, that there is no guarantee `serializedRecords` are not equal to `null`.
I think we should ensure this before writing `Chunk` into  `sendQueue`.
It is not something touching `sendFromQueue` implementation, so in case we want to ensure no `null` values allowed into `sendQueue`, I could add this as a separate PR.
Wdyt?

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -466,41 +469,72 @@ private[producer] final class ProducerLive(
...
+              Unsafe.unsafe { implicit u =>
+                exec {
+                  runtime.unsafe.run(
+                    sentRecordsCountRef.update { sentRecordsCount =>
+                      // Updating sentResults[resultIndex] here is safe:
+                      //  - Ref.update guarantees sentResults.update executed atomically
+                      //  - Ref.update starts with volatile variable read and ends with volatile variable write,
+                      //    which guarantees sentResults.update executed on the latest updated version of sentResults
+                      //    and currently updated version of sentResults
+                      //    will be visible to the next sentResults read or update called within Ref.update
```
ytalashko, line 489: This way (via `Ref`) we ensure that no race is taking place and `sentResults` array data is "visible" for the users getting this data.

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -466,41 +469,72 @@ private[producer] final class ProducerLive(
...
-                exec {
-                  runtime.unsafe.run(done.succeed(Chunk.fill(serializedRecords.size)(Left(e)))).getOrThrowFiberFailure()
+                  )
+                }
+              }
+
+            var previousSendCallSucceed = true
+
+            while (recordsIterator.hasNext) {
+              val (record: ByteRecord, recordIndex: Int) = recordsIterator.next()
```
erikvanoosten, line 510: This is equivalent to:
```suggestion
            recordsIterator.foreach { case (record: ByteRecord, recordIndex: Int) =>
```

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -466,41 +469,72 @@ private[producer] final class ProducerLive(
...
-          } catch {
-            case NonFatal(e) =>
-              Unsafe.unsafe { implicit u =>
-                exec {
-                  runtime.unsafe.run(done.succeed(Chunk.fill(serializedRecords.size)(Left(e)))).getOrThrowFiberFailure()
+                  )
+                }
+              }
+
+            var previousSendCallSucceed = true
```
erikvanoosten, line 507:
```suggestion
            var previousSendCallsSucceed = true
```

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -187,6 +186,10 @@ trait Producer {
 }
 
 object Producer {
+  case object SendOmittedDueToPreviousRecordSendCallFailureError
+      extends RuntimeException("Send omitted due to the previous record send call failure")
+      with NoStackTrace
```
erikvanoosten, line 191: In ZIO context we always talk about 'error' not 'failure' (an error is if something goes wrong,  failure is if something bad happens). As a library, we do not know if not being able to send a record is a failure or not. `FailureError` is confusing, we can stick to the java convention: Exception.

I really don't like the very long name, we could move explanation to the message. WDYT?

The exception message can be a bit more elaborate.

I propose:
```suggestion
  case object SendOmittedException
      extends RuntimeException("Send omitted due to a send error for a previous record in the chunk")
      with NoStackTrace
```

Or:
```suggestion
  case object SendOmittedException
      extends RuntimeException("Send omitted due to a send error for an earlier record in the chunk")
      with NoStackTrace
```

I am not a native speaker, a recommendation on what sounds better is appreciated.

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -466,41 +469,72 @@ private[producer] final class ProducerLive(
...
+            new Array[Either[Throwable, RecordMetadata]](recordsLength)
+
+          Ref.make(0).map { sentRecordsCountRef =>
+            @inline def safelyInsertSentResult(resultIndex: Int, sentResult: Either[Throwable, RecordMetadata]): Unit =
+              Unsafe.unsafe { implicit u =>
+                exec {
+                  runtime.unsafe.run(
+                    sentRecordsCountRef.update { sentRecordsCount =>
+                      // Updating sentResults[resultIndex] here is safe:
+                      //  - Ref.update guarantees sentResults.update executed atomically
```
erikvanoosten, line 485: IMHO, there is no reason to explain ref.update, all zio maintainers are fully aware of this. Also I am assuming that people that are new to ZIO are not going to look at this code :)

erikvanoosten (commented):
Nice! By making sure the loop uses the ref for every case we can retain the trick of counting the number of updates to the array before calling done.succeed. Much better than my rough proposal. 🚀 

The benchmark looks very good! (this pr is the last point on the graph)
Scherm­afbeelding 2024-09-01 om 09 41 03

What is missing now is a test and documentation. Let us know if you want help with either of these.

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
erikvanoosten: Most Scala code I know pretends `null` does not exist. Only when we need to interoperate with java code we'd wrap the returned value in an Option first and work from there.

Since everybody does this (no-one introduces `null` in Scala code), it works surprisingly well.

afbeelding

ytalashko: Makes perfect sense, thanks Erik, 👍

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
ytalashko: Good call, applied, 👍

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
ytalashko: Yeah, a bit more suitable name, applied

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
ytalashko: I like the shorter name, thanks!
Used `Send omitted due to a send error for a previous record in the chunk` message option, 'cause `previous` seems more accurate than `earlier` to me, but Im also not a native speaker.

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
ytalashko: Good call, left only JMM related part, explaining results are visible during access in concurrent env

ytalashko:
> What is missing now is a test and documentation. Let us know if you want help with either of these.

Sounds great, 👍, thanks for the suggestions, Erik!
Overall, if there is no hurry with this changes, I can add both myself bit by bit, if this works for you.
Already started adding unit tests. Could you, please, elaborate, which documentation we want to have and where, I believe a new `.md` file in `docs` folder or addition to some existing one (not sure which one, 'cause there is not much on the producer side)?
Let me know otherwise, so I can commit what I have started for tests, maybe it will be of any use.

erikvanoosten:
> Could you, please, elaborate, which documentation we want to have and where

I was mostly thinking of the scaladoc of the `produceChunk*` methods in `trait Producer`. Perhaps add something like:

For all produce methods except `produceChunkAsyncWithFailures`:
```
When publishing any of the records fails, the whole batch fails even though some records might have been published. Use produceChunkAsyncWithFailures to get results per record.
```

And for `produceChunkAsyncWithFailures`:
```
When publishing a record fails, the following records in the batch are not published. This is indicated with a [[SendOmittedException]].
```

About that, I now see that `SendOmittedException` should have been called `PublishOmittedException`!

> Let me know otherwise, so I can commit what I have started for tests, maybe it will be of any use.

We don't have any low level tests for the producer yet (I mean, with a mock kafka client). Anything in this area would be super helpful!

ytalashko:
> I was mostly thinking of the scaladoc of the produceChunk* methods in trait Producer. Perhaps add something like:

Oh, got it, sounds good, will do so, 👍
Thanks for the suggestions!

> About that, I now see that SendOmittedException should have been called PublishOmittedException!

Yeah, this is a great call, thanks, Erik!

 > We don't have any low level tests for the producer yet (I mean, with a mock kafka client). Anything in this area would be super helpful!
 
Nice, will update PR soon.

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -162,6 +173,9 @@ trait Producer {
    * This variant of `produceChunkAsync` more accurately reflects that individual records within the Chunk can fail to
    * publish, rather than the failure being at the level of the Chunk.
    *
+   * When attempt to send a record into buffer for publication fails, the following records in the chunk are not
+   * published. This is indicated with a [[Producer.PublishOmittedException]].
```
ytalashko, line 177: Used a bit different wording to highlight send attempt failure rather than publication failure.
Let me know, if you prefer initially suggested wording, and I'll update.

erikvanoosten: LGTM!

erikvanoosten:
I think this looks amazing. The test are pretty complicated but they look like what is needed.
Would you like me to update the description of the PR? IMHO that would be the last thing before we can merge.

@svroonland, did you already take a look as well?

ytalashko:
> The test are pretty complicated but they look like what is needed.

I think they can be revisited later and simplified updated, this is just a start so to say

> Would you like me to update the description of the PR?

Sure, would love that, thanks, Erik!

erikvanoosten:
> > Would you like me to update the description of the PR?
> 
> Sure, would love that, thanks, Erik!

Done!

erikvanoosten (approved):

erikvanoosten:
In this PR the producer no longer attempts to send subsequent records after the `send` method fails. As of zio-kafka 2.10.0 the producer also stops sending after a failure communicated via the callback. For more details see #1437.

PR #1326, opened 2024-09-14
Optimize Producer sendFromQueue implementation

ytalashko:
From java's producer send method doc: `Note that callbacks will generally execute in the I/O thread of the producer and so should be reasonably fast or they will delay the sending of messages from other threads`.
I believe we should try to keep callback implementation fast, but with recent changes we added to provide proper error responses, we make the callback slower. In general we are talking approximately about 1-3 microseconds (current master) instead of 0,1-0,2 microseconds (version 2.8.2) for single callback execution time on my laptop. I believe it was reasonable, but at the same time wanted to keep implementation fast, and provided the fastest version I come up with at that time.
Now, I got with an idea of faster implementation (with keeping the same proper error semantics). With this implementation single callback execution takes approximately 0,05-0,2 microseconds (even a bit faster than version 2.8.2).

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
@@ -493,30 +493,18 @@ private[producer] final class ProducerLive(
             new Array[Either[Throwable, RecordMetadata]](recordsLength)
 
           @inline def insertSentResult(resultIndex: Int, sentResult: Either[Throwable, RecordMetadata]): Unit = {
-            var notInserted = true
+            // Updating sentResults[resultIndex] here is safe,
+            //  because only a single update for every resultIndex of sentResults is performed
+            sentResults.update(resultIndex, sentResult)
 
-            while (notInserted) {
-              val sentRecordsCount = sentRecordsCounter.get()
+            // Reading from sentRecordsCounter guarantees fully updated version of sentResults
+            //  will be visible and used to complete done promise
+            if (sentRecordsCounter.incrementAndGet() == recordsLength) {
+              val sentResultsChunk = Chunk.fromArray(sentResults)
 
-              // Updating sentResults[resultIndex] here is safe,
-              //  because volatile variable read performed right before the update and volatile variable write after,
-              //  which guarantees sentResults.update executed on the latest updated version of sentResults
-              //  and currently updated version of sentResults
-              //  will be visible to the next sentResults read or update called after volatile variable read
-              sentResults.update(resultIndex, sentResult)
-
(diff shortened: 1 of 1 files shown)


zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -484,68 +485,68 @@ private[producer] final class ProducerLive(
...
+
+            while (notInserted) {
+              val sentRecordsCount = sentRecordsCounter.get()
+
+              // Updating sentResults[resultIndex] here is safe,
+              //  because volatile variable read performed right before the update and volatile variable write after,
+              //  which guarantees sentResults.update executed on the latest updated version of sentResults
+              //  and currently updated version of sentResults
+              //  will be visible to the next sentResults read or update called after volatile variable read
+              sentResults.update(resultIndex, sentResult)
```
erikvanoosten, line 506: Since `sentResults` is an Array, we can assign to any index, as long as we know there is a single writer for each index (which is the case). The only thing that needs to be multi-thread protected is `sentRecordsCounter` but that is now an `AtomicInteger`, so calling `increaseAndGet` would be sufficient, no while loop needed.

Or did I miss something.

erikvanoosten:
Nice work! I think it can be simplified further (see the other comment).

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
ytalashko: Yeah, you are right Erik, thats awesome 🔥
Many thanks for the suggestion!
I was scared that without having the latest updated version of `sentResults` during current `update`, we may run into issue with losing concurrent update result to some other index (similar to word tearing), but from some googling I can see that JVM handles this, and it is safe to do as you suggesting.

ytalashko:
> Nice work! I think it can be simplified further (see the other comment).

Thanks!
Yeah, did so

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -484,68 +485,56 @@ private[producer] final class ProducerLive(
...
-                        exec {
-                          runtime.unsafe.run(done.succeed(sentResultsChunk))
-                        }
-                      }
-
-                      newSentRecordsCount
-                    }
-                  )
-                }
+                val _ = runtime.unsafe.run(done.succeed(sentResultsChunk))
```
ytalashko, line 506: inlined `exec` here as it was used only here.

erikvanoosten (approved):
Good work!

Issue #1331, opened 2024-09-19
Serialization throwing defects


lemony312:
I have a case class using zio json.
My message is failing to serialize as a string to due to an accidental null.
The problem is the library causes a defect in the program as opposed to a typed exception.

```
 case class MyMessage(errors: List[String])
  object MyMessage {
    implicit val decoder: JsonDecoder[MyMessage] = DeriveJsonDecoder.gen[MyMessage]
    implicit val encoder: JsonEncoder[MyMessage] = DeriveJsonEncoder.gen[MyMessage]

   implicit val serde =  Serde.string.contramap[T](_.toJson)
}
```

accidentally, we created a message with a null in the list 
```
val myMessage = MyMessage(List(null, "hello"))
``` 

the code `_.toJson` ends up throwing an exception since it is supposed to be a string and it contains a null.

The problem is that the library does not surface this error - it is modeled as an RIO[] but this is actually a defect.

the problem is in the `Serde.string.contramap`
```
 def serialize(topic: String, headers: Headers, value: T): RIO[R, Array[Byte]]

  /**
   * Create a serializer for a type U based on the serializer for type T and a mapping function
   */
  def contramap[U](f: U => T): Serializer[R, U] =
    Serializer((topic, headers, u) => serialize(topic, headers, f(u)))
```

it invokes the `f(u)` without wrapping in the zio - this causes an exception to be thrown that is not caught.

one naive solution is wrap it in zio.attempt (which the library does later in the code) - but obviously fails to prevent the defect from happening here

the problem is in the `Serde.string.contramap`
```
 def serialize(topic: String, headers: Headers, value: T): RIO[R, Array[Byte]]

  /**
   * Create a serializer for a type U based on the serializer for type T and a mapping function
   */
  def contramap[U](f: U => T): Serializer[R, U] =
    Serializer((topic, headers, u) =>  ZIO.attempt(f(u)).flatMap(serialize(topic, headers, _))
```

erikvanoosten:
Hello @lemony312. I have never liked Kafka's serialization support, precisely because of these kinds of reasons. In general I recommend using `Array[byte]` consumers/producers and do the (de)serialization in your own code.

I will look at your PR though.

svroonland:
Just to clarify: did you get a 'died zio' (defect) somewhere or did zio-kafka silently ignore this error somehow?

PR #1332, opened 2024-09-19
surface serialization error

lemony312:
solution to 
https://github.com/zio/zio-kafka/issues/1331

### DIFF
--- a/zio-kafka-test/src/test/scala/zio/kafka/serde/SerializerSpec.scala
+++ b/zio-kafka-test/src/test/scala/zio/kafka/serde/SerializerSpec.scala
@@ -2,7 +2,7 @@ package zio.kafka.serde
 
 import org.apache.kafka.common.header.internals.RecordHeaders
 import zio.test.Assertion._
-import zio.test._
+import zio.test.{ test, _ }
 import zio.ZAny
 import zio.kafka.ZIOSpecDefaultSlf4j
 
@@ -21,7 +21,11 @@ object SerializerSpec extends ZIOSpecDefaultSlf4j {
           )
         }
       }
-    )
+    ),
+    test("prevent serialization defect") {
+      val failingSerializer = Serde.string.contramap[Int](_ => throw new RuntimeException("null field"))
+      assertZIO(failingSerializer.serialize("topic1", new RecordHeaders, 123).exit)(failsWithA[RuntimeException])
+    }
   )
   private lazy val stringSerializer: Serializer[Any, String] = Serde.string
 }
--- a/zio-kafka-test/src/test/scala/zio/kafka/serde/SerializerSpec.scala
+++ b/zio-kafka-test/src/test/scala/zio/kafka/serde/SerializerSpec.scala
@@ -2,7 +2,7 @@ package zio.kafka.serde
 
 import org.apache.kafka.common.header.internals.RecordHeaders
 import zio.test.Assertion._
-import zio.test.{ test, _ }
+import zio.test._
 import zio.ZAny
 import zio.kafka.ZIOSpecDefaultSlf4j
 
--- a/zio-kafka/src/main/scala/zio/kafka/serde/Serializer.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/serde/Serializer.scala
@@ -20,7 +20,7 @@ trait Serializer[-R, -T] {
    * Create a serializer for a type U based on the serializer for type T and a mapping function
    */
   def contramap[U](f: U => T): Serializer[R, U] =
-    Serializer((topic, headers, u) => serialize(topic, headers, f(u)))
+    Serializer((topic, headers, u) => ZIO.attempt(f(u)).flatMap(serialize(topic, headers, _)))
 
   /**
    * Create a serializer for a type U based on the serializer for type T and an effectful mapping function


zio-kafka-test/src/test/scala/zio/kafka/serde/SerializerSpec.scala
```diff
@@ -2,7 +2,7 @@ package zio.kafka.serde
 
 import org.apache.kafka.common.header.internals.RecordHeaders
 import zio.test.Assertion._
-import zio.test._
+import zio.test.{ test, _ }
```
erikvanoosten, line 5: This shouldn't be necessary. Does it still compile after reverting this line?

zio-kafka/src/main/scala/zio/kafka/serde/Serializer.scala
```diff
@@ -20,7 +20,7 @@ trait Serializer[-R, -T] {
    * Create a serializer for a type U based on the serializer for type T and a mapping function
    */
   def contramap[U](f: U => T): Serializer[R, U] =
-    Serializer((topic, headers, u) => serialize(topic, headers, f(u)))
+    Serializer((topic, headers, u) => ZIO.attempt(f(u)).flatMap(serialize(topic, headers, _)))
```
erikvanoosten, line 23: Could you make the same change for `contramapM` please?

erikvanoosten (commented):
This won't be good for performance, regardless, it looks like a good idea. Would you address the small comments?

svroonland:
> This won't be good for performance, regardless, it looks like a good idea. 

That's my concern as well, as this is executed for every message. Would it be better to add a warning to the docs that exceptions are not handled for this reason and result in a defect + refer to `contramapM` if you want error handling?

erikvanoosten:
> > This won't be good for performance, regardless, it looks like a good idea.
> 
> That's my concern as well, as this is executed for every message. Would it be better to add a warning to the docs that exceptions are not handled for this reason and result in a defect + refer to `contramapM` if you want error handling?

Another option is to accept the performance hit and recommend people to do (de)serialization in their own code. We can even give some example code how to do this efficiently (especially taking chunks into account). WDYT?

svroonland:
The Consumer already deserializes in chunks, not sure how that could be done more efficiently in user code, but probably I'm missing something. What did you have in mind?

erikvanoosten:
Yes, that is the case. The point is more that if you run (de)serialization in your own code you should take chunking into account.
Still, it could go faster because you can make more assumptions. In particular, it is often okay to convert the entire chunk in a single ZIO. This saves a lot of zio-flatmapping.

lemony312:
It's not ideal for performance but the code base already wraps serializers in ZIO.attempt for the java one
https://github.com/zio/zio-kafka/blob/master/zio-kafka/src/main/scala/zio/kafka/serde/Serializer.scala#L70

this is partially why the bug was hard to diagnose - it looks as though the error channel is strongly typed.

zio-kafka/src/main/scala/zio/kafka/serde/Serializer.scala
lemony312: `contramapM` takes in a ZIO that can fail with throwable - do you think its worth wrapping?

erikvanoosten: Yes, that is good question. If I read https://zio.dev/guides/migrate/zio-2.x-migration-guide#guidelines-for-library-authors point 2 right, then yes we should do this.

erikvanoosten:
> > This won't be good for performance, regardless, it looks like a good idea.
> 
> That's my concern as well, as this is executed for every message. Would it be better to add a warning to the docs that exceptions are not handled for this reason and result in a defect + refer to `contramapM` if you want error handling?

@svroonland If I read https://zio.dev/guides/migrate/zio-2.x-migration-guide#guidelines-for-library-authors point 2 right, then we should do the same wrapping in `contramapM`.

ytalashko:
Just leaving my 2c, in case it is gonna be helpful, otherwise ignore it:
I like the suggestion in https://github.com/zio/zio-kafka/pull/1332#issuecomment-2363001954.
Basically, the error is happening in case the mapping function produces it.
The comments seem to be discussing two options to tackle that:
1) Lib handles exceptions for user, but this hurts performance.
2) It is users responsibility to not `throw` exceptions.

With first option, library works faster for users not throwing exceptions (which seems like a general case, at least my assumption) and produces defect in case of bugs (I say "bugs", 'cause if your code throws exceptions, you should ensure they are handled, and also in functional programming throwing exceptions is kinda an anti pattern). I believe `ZIO.map` works exactly in the same way. I also believe, it is a right way (produce defect in case of bug), 'cause otherwise performance hit is taken.

With second option, library works slower for users not throwing exceptions (which seems like a general case, at least my assumption) and handles user's bugs. This way library hurts users who not having bugs, which seems to me like a bad thing. Having API method taking pure function, you (at least I) assume it is not doing some additional job by trying to handle an exception. In case my mapping function throws errors, I would look for something that can take that into account (e.g.  `contramapM`) gives an ability to express that. Even, from the [issue description](https://github.com/zio/zio-kafka/issues/1331) it seems to be a bug to me, and having a defect in case of bugs is seems expected.

svroonland:
Agreed with @ytalashko. As long as exceptions in `f` result in a died zio that is surfaced to user code through zio-kafka, this is acceptable to me and consistent with how ZIO's map and flatMap behave. If it's not surfaced we should see if we can fix that in the Producer. We can add a comment to `Serializer#contramap` warning about this and recommending to use `contramapM` if you expect errors in your `f`.

What do you think @erikvanoosten?

erikvanoosten:
> Agreed with @ytalashko. As long as exceptions in `f` result in a died zio that is surfaced to user code through zio-kafka, this is acceptable to me and consistent with how ZIO's map and flatMap behave. If it's not surfaced we should see if we can fix that in the Producer. We can add a comment to `Serializer#contramap` warning about this and recommending to use `contramapM` if you expect errors in your `f`.
> 
> What do you think @erikvanoosten?

Yes, I agree. Thanks @ytalashko.

erikvanoosten:
@lemony312 I am going to close this PR as per the discussion above.
For documenting the behavior we can create another PR.
Thanks for your effort. It didn't result in a code change, but at least we have some clarity.

PR #1337, opened 2024-09-29
Document Serdes in more depth

### LINKED ISSUE #1331: Serialization throwing defects

erikvanoosten:
Fixes #1331.

### DIFF
--- a/docs/serialization-and-deserialization.md
+++ b/docs/serialization-and-deserialization.md
@@ -17,7 +17,6 @@ This document contains:
 - How to create a custom serde
 - How to create and use a custom serde that wraps invalid data
 - How to do deserialization in the consumer stream
-
 - A warning about using `mapZIO`
 
 ## Handling failures in a serde
--- a/docs/serialization-and-deserialization.md
+++ b/docs/serialization-and-deserialization.md
@@ -28,7 +28,7 @@ provided `Serdes.byteArray` and `Serdes.string`. Custom serdes should not fail o
 remote lookups that could fail.
 
 For cases where failure is possible, we recommend using `Serdes.byteArray`, put the deserialization code in the consumer
-stream, and do serialization before handing the data to zio-kafka. This way you can handle failure any way you want.<br>
+stream, and do serialization before handing the data to zio-kafka. This way you can handle failure any way you want.<br/>
 Another option is to make a serde that wraps data in a Try.
 
 These approaches are further discussed below.
--- a/docs/serialization-and-deserialization.md
+++ b/docs/serialization-and-deserialization.md
@@ -1,24 +1,91 @@
 ---
 id: serialization-and-deserialization
-title: "Serialization And Deserialization"
+title: "Serialization and Deserialization"
 ---
+
+Zio-kafka supports serializers and deserializers (serde) to convert record keys and values to and from byte arrays.
+
+A `Serde` combines a `Deserializer` and a `Serializer`. Common serdes are provided in the `Serdes` object, e.g.
+`Serdes.byteArray`, `Serdes.string` and `Serdes.long`. A serde can be converted to other serdes, or you can create a
+custom serde by implementing the `Serde` trait directly.
+
+This document contains:
+
+- Handling failures in a serde
+- How to write a custom serde
+- How to do deserialization in the consumer stream
+- How to write and use a custom serde that wraps invalid data
+- A warning about using `mapZIO`
+
+## Handling failures in a serde
+
+By default, a consumer stream will fail if it encounters a deserialization error in the serde. Unfortunately, the
+resulting failure might not clearly indicate that the cause is in the serde. It is also not straightforward to handle
+failures from a serde.
+
+Therefore, we recommend serdes (custom or not) only when they always succeed. This is for example the case with the
+provided `Serdes.byteArray` and `Serdes.string`. Custom serdes should not fail on invalid input, and should not do
+remote lookups that could fail.
+
+For cases where failure is possible, we recommend using `Serdes.byteArray`, put the deserialization code in the consumer
+stream, and do serialization before handing the data to zio-kafka. This way you can handle failure any way you want.<br>
+Another option is to make a serde that wraps data in a Try.
+
+These approaches are further discussed below.
 
 ## Custom Data Type Serdes
 
-Serializers and deserializers (serdes) for custom data types can be constructed from scratch or by converting existing serdes. For example, to create a serde for an `Instant`:
+Serializers and deserializers for custom data types can be created from scratch, or by converting existing
+serdes. For example, to create a serde for an `Instant` from a serde for a `Long`:
 
 ```scala
 import java.time.Instant
 import zio.kafka.serde._
 
-val instantSerde: Serde[Any, Instant] = Serde.long.inmap(java.time.Instant.ofEpochMilli)(_.toEpochMilli)
+val instantSerde: Serde[Any, Instant] = Serdes.long.inmap(java.time.Instant.ofEpochMilli)(_.toEpochMilli)
 ```
 
-## Handling deserialization failures
+To handle missing data (an empty key or value), you can use the `Serde.asOption` transformer. For example:
+`Serdes.string.asOption`. This results in a `None` if the key or value is empty, and in a `Some(string)` otherwise.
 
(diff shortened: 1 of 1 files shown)


erikvanoosten:
@ytalashko, would you like to review this as well?

docs/serialization-and-deserialization.md
```diff
@@ -1,24 +1,91 @@
 ---
 id: serialization-and-deserialization
-title: "Serialization And Deserialization"
+title: "Serialization and Deserialization"
 ---
 
+Zio-kafka supports serializers and deserializers (serde) to convert record keys and values to and from byte arrays.
```
svroonland, line 6: from byte arrays to scala case classes and vice versa.

docs/serialization-and-deserialization.md
```diff
@@ -1,24 +1,91 @@
...
+
+Deserialization in the consumer stream can be done with regular ZIO operators. Here is an example:
+
+```scala
+import zio._, stream._
+import zio.kafka.consumer._
+
+val consumer = ZLayer.scoped(Consumer.make(consumerSettings))
+
+val stream = Consumer
```
svroonland, line 61: This feels quite low level, I don't think we should be promoting byteArray based serdes as the first example here. For most use cases providing the key and value Serde to the `Consumer` is fine, especially if you use `asOption` or `asTry` to handle any errors. If you want to squeeze extra performance, use byteArrays and do your own deserialization.

svroonland (changes requested):

docs/serialization-and-deserialization.md
erikvanoosten: For me it is not about performance but more about freedom in how to handle chunks and failure scenarios. Regardless, I don't mind switching the order.

ytalashko: I would also vote to have a custom key & value serdes in the first example, since it covers a bit more, e.g. how to send/get desired ADTs to/from kafka, instead of byte arrays, which should easier to start from for some real or showcase application (to me).

docs/serialization-and-deserialization.md
```diff
@@ -1,24 +1,91 @@
...
 
-## Handling deserialization failures
+## Custom serdes that wraps invalid data
 
-The default behavior for a consumer stream when encountering a deserialization failure is to fail the stream. In many cases you may want to handle this situation differently, e.g. by skipping the message that failed to deserialize or by executing an alternative effect. For this purpose, any `Deserializer[T]` for some type `T` can be easily converted into a `Deserializer[Try[T]]` where deserialization failures are converted to a `Failure` using the `asTry` method.
+Any `Deserializer[A]` for a given type `A` can be converted into  a `Deserializer[Try[A]]` where deserialization
+failures are converted to a `Failure` using the `asTry` method.
 
-Below is an example of skipping messages that fail to deserialize. The offset is passed downstream to be committed.
+Below is an example that skips messages that fail to deserialize. The offset is passed downstream to be committed.
```
ytalashko, line 88: Minor one. I kinda like previous wording more, but maybe it only me.

docs/serialization-and-deserialization.md
```diff
@@ -50,3 +117,39 @@ stream
   .runDrain
   .provideSomeLayer(consumer)
 ```
+
+## A warning about `mapZIO`
```
ytalashko, line 121: Great hints/insights inside this section, 👍

ytalashko (approved):
> @ytalashko, would you like to review this as well?

Sure

Looks great overall!
Left few minor/optional suggestions

docs/serialization-and-deserialization.md
erikvanoosten: Indeed, this text is not very clear. I improved it in https://github.com/zio/zio-kafka/pull/1337/commits/d5bd32f956a309e7d021a569d45328395eba2252

docs/serialization-and-deserialization.md
erikvanoosten: Alright. In https://github.com/zio/zio-kafka/pull/1337/commits/d5bd32f956a309e7d021a569d45328395eba2252 we no longer recommend handling data in the consumer stream as the primary solution, and using `.asTry` is the first offered solution.

docs/serialization-and-deserialization.md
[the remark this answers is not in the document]
erikvanoosten: In https://github.com/zio/zio-kafka/pull/1337/commits/d5bd32f956a309e7d021a569d45328395eba2252 I completely rewrote this paragraph as it was far from clear. Thanks for the comments.

docs/serialization-and-deserialization.md
```diff
@@ -1,24 +1,91 @@
...
+Zio-kafka supports serializers and deserializers (serde) to convert record keys and values to and from byte arrays.
+
+A `Serde` combines a `Deserializer` and a `Serializer`. Common serdes are provided in the `Serdes` object, e.g.
+`Serdes.byteArray`, `Serdes.string` and `Serdes.long`. A serde can be converted to other serdes, or you can create a
+custom serde by implementing the `Serde` trait directly.
+
+This document contains:
+
+- Handling failures in a serde
+- How to write a custom serde
```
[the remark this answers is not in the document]
erikvanoosten: Accepted in https://github.com/zio/zio-kafka/pull/1337/commits/d5bd32f956a309e7d021a569d45328395eba2252

docs/serialization-and-deserialization.md
```diff
@@ -1,24 +1,91 @@
...
+- How to do deserialization in the consumer stream
+- How to write and use a custom serde that wraps invalid data
```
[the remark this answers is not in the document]
erikvanoosten: Accepted in https://github.com/zio/zio-kafka/pull/1337/commits/d5bd32f956a309e7d021a569d45328395eba2252

docs/serialization-and-deserialization.md
erikvanoosten: I am fine with either. Reverted to previous words in https://github.com/zio/zio-kafka/pull/1337/commits/d5bd32f956a309e7d021a569d45328395eba2252.

erikvanoosten:
@svroonland @ytalashko Thanks for your comments. They were very useful. I have addressed them in https://github.com/zio/zio-kafka/pull/1337/commits/d5bd32f956a309e7d021a569d45328395eba2252.

svroonland (commented):
It would be good to have an example here of deserializing json, with zio-json

svroonland (approved):

PR #1351, opened 2024-10-29
Enable mima check during release

erikvanoosten:
This automatically enables a binary compatibility check using Mima when we build a release (when we build from a tag).


erikvanoosten:
Unfortunately this currently prints something to stderr (in the form of `fatal: no tag exactly matches 'aa3bc7c5bb2e8f95751ea7dec0e4998f5880babd'`). I am still figuring out how to disable this...

svroonland (approved):

PR #1356, opened 2024-11-01
Open telemetry support (alt)

erikvanoosten:
Retain open telemetry context as provided by the zio-telemetry libraries by copying the relevant fiber ref values between fibers within the zio-kafka consumer and producer. This allows zio-kafka / zio-telemetry programs to integrate with Java Kafka clients that have been instrumented by the default OpenTelemetry java agent.

This is an alternative implementation to #1345.

At the moment this PR only contains the code that will get and set fiber ref values. When this PR is ready, this code will be integrated into the consumer and producer.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/telemetry/TelemetryPropagation.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/telemetry/TelemetryPropagation.scala
@@ -1,0 +1,74 @@
+package zio.kafka.telemetry
+
+import zio._
+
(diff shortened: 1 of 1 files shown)


erikvanoosten:
Stopping the work. From the Otel documentation I get the idea that Otel is a hotchpotch of open and vendor libraries. It is still unclear to me how the we can do the required low level integration. 

Anyone with more knowledge that is interested, please feel free to take over this task.

PR #1358, opened 2024-11-02
Fix slow rebalanceSafeCommits behavior

svroonland:
After consumer 1 is shutdown (using stopConsumption), rebalances happen and partitions from consumer 2 are assigned. These streams are never started, so the finalizer completing completedPromise is never called. Waiting for these to complete takes 3 minutes (default maxRebalanceDuration).

In case that streams were assigned and no record was ever put in their queues, there's no use in waiting for the stream to complete by committing some offset.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
@@ -65,6 +65,8 @@ final class PartitionStreamControl private (
 
   def queueSize: UIO[Int] = queueInfoRef.get.map(_.size)
 
+  def lastPulledOffset: UIO[Option[Offset]] = queueInfoRef.get.map(_.lastPulledOffset)
+
   /**
    * @return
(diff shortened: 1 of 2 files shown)


svroonland:
@erikvanoosten If this mechanism is okay by you, we can add to it:
* Logging (debug) the situation that the stream has not pulled any data so we're not awaiting for it to complete
* Logging (warning) when we're done awaiting a stream to complete but the deadline has passed. Right now it's not clear why the wait takes so long.

erikvanoosten (approved):
Smart!

erikvanoosten:
> @erikvanoosten If this mechanism is okay by you, we can add to it:
> 
>     * Logging (debug) the situation that the stream has not pulled any data so we're not awaiting for it to complete
>     * Logging (warning) when we're done awaiting a stream to complete but the deadline has passed. Right now it's not clear why the wait takes so long.

IMHO this PR is already very valuable in itself and I wouldn't await merging. These 2 improvements can be put in follow up PRs.

svroonland:
Alright let's do that.

Issue #1132, opened 2023-12-07
Slow (timeout) shutdown and rebalancing with `withRebalanceSafeCommits `


jgordijn:
I tried out the new `withRebalanceSafeCommits` feature and it has unexpected behavior:
1. Shutdown takes long time now. Result is that all partitions handled by the stopping application are stuck
    It seems that on shutdown the stream is stopped, but the rebalanceListener is waiting until the last message is committed. As the stream is stopped, this will never happen and it takes 3 min (3/5 of the maxPollInterval) to stop. 
2. Joining the group takes a long time.
   When I start the second application it seems that rebalancing is started. However it still takes 3 minutes to join. 
   
   
Maybe it has something to do with the slow (100ms) processing per message, but having a 10 message poll should mitigate this. I would thus expect rebalancing to happen in (worst case) 1 sec (10x100ms).

```scala
package listener

import org.apache.kafka.clients.consumer.ConsumerConfig
import org.slf4j.LoggerFactory
import zio.kafka.consumer.{Consumer, ConsumerSettings, RebalanceListener, Subscription}
import zio.kafka.serde.Serde
import zio.stream.ZStream
import zio.{ZIO, ZIOAppDefault, ZLayer, durationInt}

object Main extends ZIOAppDefault {
  private val LOG = LoggerFactory.getLogger(Main.getClass)

  private val TOPIC = "topic1"

  def consumer: ZStream[Consumer, Throwable, Nothing] = Consumer
    .plainStream(Subscription.topics(TOPIC), Serde.string, Serde.string)
    .tap(r =>
      ZIO.succeed(LOG.info(
        s"${r.value} p:${r.offset.partition}, o:${r.offset.offset}"
      ))
    )
    .mapZIO(ZIO.succeed(_).delay(100.millis))
    .map(_.offset)
    .aggregateAsync(Consumer.offsetBatches)
    .tap(offsets =>
      ZIO.succeed(LOG.info(s"Going to commit: ${offsets.offsets.map { case (k, v) => s"${k.partition()}: ${v.offset()}" }}"))
    )
    .mapZIO(_.commit)
    .drain

  def consumerLayer = {
    val value: Map[String, AnyRef] = Map[String, AnyRef](
      ConsumerConfig.SESSION_TIMEOUT_MS_CONFIG -> "60000",
      ConsumerConfig.HEARTBEAT_INTERVAL_MS_CONFIG -> "250"
    )
    ZLayer.scoped(
      Consumer.make(
[...]

PR #1360, opened 2024-11-03
More logging around rebalancing when rebalanceSafeCommits is true

svroonland:
Follow up of #1358 

See also #1132

### DIFF
--- a/zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
+++ b/zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
@@ -864,7 +864,7 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
           testForPartitionAssignmentStrategy[RangeAssignor],
           testForPartitionAssignmentStrategy[CooperativeStickyAssignor]
         )
-      }: _*) @@ TestAspect.sequential,
+      }: _*),
       test("partitions for topic doesn't fail if doesn't exist") {
         for {
           topic  <- randomTopic
--- a/zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
+++ b/zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
@@ -769,7 +769,8 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
                               clientId = clientId,
                               groupId = Some(groupId),
                               `max.poll.records` = 1,
-                              rebalanceSafeCommits = rebalanceSafeCommits
+                              rebalanceSafeCommits = rebalanceSafeCommits,
+                              maxRebalanceDuration = 30.seconds
                             )
                 consumer <- Consumer.make(settings)
               } yield consumer
@@ -863,7 +864,7 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
           testForPartitionAssignmentStrategy[RangeAssignor],
           testForPartitionAssignmentStrategy[CooperativeStickyAssignor]
         )
-      }: _*),
+      }: _*) @@ TestAspect.sequential,
       test("partitions for topic doesn't fail if doesn't exist") {
         for {
           topic  <- randomTopic
--- a/zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
+++ b/zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
@@ -770,7 +770,7 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
                               groupId = Some(groupId),
                               `max.poll.records` = 1,
                               rebalanceSafeCommits = rebalanceSafeCommits,
-                              maxRebalanceDuration = 30.seconds
+                              maxRebalanceDuration = 60.seconds
                             )
                 consumer <- Consumer.make(settings)
               } yield consumer
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -217,7 +217,7 @@ private[consumer] final class Runloop private (
           .takeWhile(_ => java.lang.System.nanoTime() <= deadline)
           .scan(Chunk.empty[Runloop.Commit])(_ ++ _)
           .mapZIO(commits => endingStreamsCompletedAndCommitsExist(commits).map((_, commits)))
-          .takeUntil { case (completed, commits @ _) => completed }
+          .takeUntil { case (completed, _) => completed }
           .runLast
           .map(_.getOrElse((false, Chunk.empty)))
           .flatMap { case (completed, commits) => logFinalStreamCompletionStatuses(completed, commits) } *>
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -135,7 +135,10 @@ private[consumer] final class Runloop private (
         isDone: Boolean,
         lastPulledOffset: Option[Long],
         endOffsetCommitStatus: EndOffsetCommitStatus
-      )
+      ) {
+        override def toString: String =
+          s"${tp}: isDone=${isDone}, lastPulledOffset=${lastPulledOffset.getOrElse("none")}, endOffsetCommitStatus: ${endOffsetCommitStatus}"
+      }
 
       def getStreamCompletionStatuses(newCommits: Chunk[Commit]): ZIO[Any, Nothing, Chunk[StreamCompletionStatus]] =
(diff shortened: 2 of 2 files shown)


svroonland:
In a follow up PR I'd like to move the rebalance safe rebalance handling to a separate class / file somehow.

svroonland:
What's the desired behavior in case of a timeout? Is it ok to silently continue rebalancing there or should we fail the entire subscription/stream with some error?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -125,36 +125,71 @@ private[consumer] final class Runloop private (
...
         }
 
+      def logFinalStreamCompletionStatuses(newCommits: Chunk[Commit]): ZIO[Any, Nothing, Unit] =
+        getStreamCompletionStatuses(newCommits).flatMap { completionStatuses =>
+          val statusStrings = completionStatuses.map(_.toString)
+          ZIO
+            .logWarning(
+              s"Exceeded deadline waiting for streams to end, will continue with rebalance: ${statusStrings.mkString("; ")}"
+            )
+            .when(java.lang.System.nanoTime() >= deadline)
```
erikvanoosten, line 181: Since `nanoTime` progresses, this `when` is wrong. The decision to log should be based on the result of whatever line 219 produces.
To prevent having to execute the logic again, we could move `logFinalStreamCompletionStatuses` into `endingStreamsCompletedAndCommitsExist` (i.e. log when it would return `true`). (A side-benefit is that the diff becomes smaller 😄.)

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
```diff
@@ -863,7 +864,7 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
           testForPartitionAssignmentStrategy[RangeAssignor],
           testForPartitionAssignmentStrategy[CooperativeStickyAssignor]
         )
-      }: _*),
+      }: _*) @@ TestAspect.sequential,
```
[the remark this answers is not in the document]
svroonland: Was helpful when debugging, but I can revert it now.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
svroonland: Sharp on the nanotime progressing.

We want to log the warning also when `endingStreamsCompletedAndCommitsExist` returns false, so moving `logFinalStreamCompletionStatuses` inside it is not an option. Instead we need to look at the last value in the stream and log based on that.

svroonland: I'm afraid that didn't make the diff much better 😅

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -174,15 +209,18 @@ private[consumer] final class Runloop private (
...
         ZStream
           .fromZIO(blockingSleep(commitQueuePollInterval) *> commitQueue.takeAll)
           .tap(commitAsync)
           .forever
           .takeWhile(_ => java.lang.System.nanoTime() <= deadline)
           .scan(Chunk.empty[Runloop.Commit])(_ ++ _)
-          .takeUntilZIO(endingStreamsCompletedAndCommitsExist)
-          .runDrain *>
+          .mapZIO(commits => endingStreamsCompletedAndCommitsExist(commits).map((_, commits)))
+          .takeUntil { case (completed, commits @ _) => completed }
```
erikvanoosten, line 220:
```suggestion
          .takeUntil { case (completed, _) => completed }
```

svroonland: Both are fine with me. The current form makes it explicit what you're ignoring.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -125,36 +125,71 @@ private[consumer] final class Runloop private (
...
+          .unit
+
+      def endingStreamsCompletedAndCommitsExist(newCommits: Chunk[Commit]): UIO[Boolean] =
+        for {
+          completionStatuses <- getStreamCompletionStatuses(newCommits)
+          statusStrings = completionStatuses.map(_.toString)
+          _ <- ZIO.logDebug(s"Waiting for ${streamsToEnd.size} streams to end: ${statusStrings.mkString("; ")}")
+        } yield completionStatuses.forall(status =>
+          status.lastPulledOffset.isEmpty || (status.isDone && status.endOffsetCommitStatus != EndOffsetNotCommitted)
+        )
```
erikvanoosten, line 191:
```suggestion
        } yield completionStatuses.forall { status =>
          // A stream is complete when it never got any records, or when it committed the offset of the last consumed record
          status.lastPulledOffset.isEmpty || (status.isDone && status.endOffsetCommitStatus != EndOffsetNotCommitted)
        }
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -125,36 +125,71 @@ private[consumer] final class Runloop private (
...
 
-      def endingStreamsCompletedAndCommitsExist(newCommits: Chunk[Commit]): Task[Boolean] =
+      sealed trait EndOffsetCommitStatus
+      case object EndOffsetNotCommitted  extends EndOffsetCommitStatus
+      case object EndOffsetCommitPending extends EndOffsetCommitStatus
+      case object EndOffsetCommitted     extends EndOffsetCommitStatus
+
+      final case class StreamCompletionStatus(
+        tp: TopicPartition,
+        isDone: Boolean,
```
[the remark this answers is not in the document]
erikvanoosten: We could also get rid of `lastPulledOffset` (unless you want to log it) by introducing a
```
case object NoOffsetToCommit extends EndOffsetCommitStatus
```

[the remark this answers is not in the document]
erikvanoosten: No, `isDone` really means the stream ended. This does not have a straight relation with whether commits were done.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -125,34 +125,70 @@ private[consumer] final class Runloop private (
...
-                }
-              }
-              endOffsetWasCommitted || endOffsetCommitIsPending
+
+                endOffsetCommitStatus = endOffset match {
+                                          case Some(endOffset)
+                                              if committedOffsets.contains(stream.tp, endOffset.offset) =>
+                                            EndOffsetCommitted
+                                          case Some(endOffset) if allPendingCommitOffsets.exists { case (tp, offset) =>
+                                                tp == stream.tp && offset.offset() >= endOffset.offset
```
erikvanoosten, line 159: `offset.offset() >= endOffset.offset` this is extremely tricky to get right. In zio-kafka we always work with the last consumed offset, in the java kafka client, they always work work with next offset to consume (which is the last consumed offset + 1).

Please double and triple check!

svroonland: I believe this is what was already the logic before this change, but good point

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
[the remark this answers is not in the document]
svroonland: 👍 Can you resolve this conversation?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
svroonland: We do the '+1' in `asyncCommitParameters` and return its result as the `offsets` parameter. This is passed only to `KafkaConsumer#commitAsync` and to some diagnostic events (`Commit.Success` and `Commit.Failure`). The latter could be inconsistent with the offsets that the user has available, we may want to create an issue for that and change it. 

So in the rest of the code, including the rebalance safe part, we work with the last consumed offsets.

This change (adding the +1) fails tests with a timeout:
```
                                          case Some(endOffset)
                                              if committedOffsets.contains(stream.tp, endOffset.offset + 1) =>
                                            EndOffsetCommitted
```

while doing `- 1` fails on the assertion. So at least that bit seems right.

The comparison for the `EndOffsetCommitPending` is harder to verify because it's such a transient situation. We don't do any logic on that status though, so it's not so critical. I would argue that we should change the `>=` to `==` here as well, because we assume (rightly) that the committed offset can never exceed the last pulled offset.

svroonland:
With the stronger test I found some new issues with unexpected interruptions, which by a lucky shot I was able to fix with a `forkDaemon.flatMap(_.join)`. I can't explain exactly why this happens, but with the new mechanism the test works well with a `nonFlaky(100)`, where it previously failed with just 10 repetitions.

Issue #241, opened 2020-11-16
Error types should be modeled more strongly


jypma:
Right now, all zio-kafka effect types have `Throwable` as their error type, requiring users to look into the kafka API itself to find out what exceptions to expect.

An argument could be made to introduce concrete error types for the expected error scenarios, and use those instead of `Throwable`. This can have some clear advantages:

- We can show the difference between a stream that silently reconnects, vs. a (not implemented yet) stream that fails when kafka's connection goes down (as requested in  https://github.com/zio/zio-kafka/issues/175 ).

- We can model deserialization errors into this type, if desired, as an alternative to the current `.asTry` construct.

In fact, in its current state, which exceptions can be expected in that `Throwable`?

svroonland:
I like this idea a lot, at the very least as an experiment to see how feasible this is and what would be a good approach. I've been considering doing the same for zio-kinesis. 

Taking `def partitionedStream` as a starting point, which can fail with any `Throwable`, we'd have to look into the call hierarchy to see which exceptions we can expect. Methods that are called on `KafkaConsumer` as part of that hierarchy are:
* `poll`
* `resume`
* `pause`
* `assignment`
* `seek`

For `pause`, `resume`, `assignment` we can refineOrDie on the `IllegalStateException`, this is already done in some places using ZIO.effectTotal.

Most of the exceptions can come from `poll`:
```scala
    /**     
     * @throws org.apache.kafka.clients.consumer.InvalidOffsetException if the offset for a partition or set of
     *             partitions is undefined or out of range and no offset reset policy has been configured
     * @throws org.apache.kafka.common.errors.WakeupException if {@link #wakeup()} is called before or while this
     *             function is called
     * @throws org.apache.kafka.common.errors.InterruptException if the calling thread is interrupted before or while
     *             this function is called
     * @throws org.apache.kafka.common.errors.AuthenticationException if authentication fails. See the exception for more details
     * @throws org.apache.kafka.common.errors.AuthorizationException if caller lacks Read access to any of the subscribed
     *             topics or to the configured groupId. See the exception for more details
     * @throws org.apache.kafka.common.KafkaException for any other unrecoverable errors (e.g. invalid groupId or
     *             session timeout, errors deserializing key/value pairs, your rebalance callback thrown exceptions,
     *             or any new error cases in future versions)
     * @throws java.lang.IllegalArgumentException if the timeout value is negative
[...]

PR #1361, opened 2024-11-03
More specific errors than Throwable for Consumer

svroonland:
Mostly replaces Throwable with `ConsumerError`, a hierarchy of custom types we can later extend, and a fallback `UnknownConsumerException`. Some exceptions that can be thrown from the underlying `KafkaConsumer` which would indicate programming errors in zio-kafka are translated to fiber deaths (as per ZIO recommendations). 

At some places we have more specific errors.
* The partition stream can fail with just a `PartitionStreamPullTimeout` or `DeserializationError`.
* Committing can fail with a `CommitError`, which is a `CommitTimeout` or `UnknownCommitException`.

I hope in the future these sub-hierarchies don't cause any conflicts. 

Build still fails because test and bench projects have not been migrated yet, first want to gather some feedback on this change.

Implements #241

### DIFF
--- a/zio-kafka-example/src/main/scala/zio/kafka/example/Main.scala
+++ b/zio-kafka-example/src/main/scala/zio/kafka/example/Main.scala
@@ -2,6 +2,7 @@ package zio.kafka.example
 
 import io.github.embeddedkafka.{ EmbeddedK, EmbeddedKafka, EmbeddedKafkaConfig }
 import zio._
+import zio.kafka.consumer.Consumer.ConsumerError
 import zio.kafka.consumer.diagnostics.Diagnostics
 import zio.kafka.consumer.{ Consumer, ConsumerSettings, Subscription }
 import zio.kafka.serde.Serde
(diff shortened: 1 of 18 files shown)


svroonland:
Yes, my thought was that this should be released in v3. Doesn't have to be right now, we can shelve it for a while together with otther major changes we may want to do.

I would prefer not to make ConsumerError a subtype of Throwable, it feels like it goes against the idea of this PR.

erikvanoosten:
Errors that are not a subtype of `Throwable` IMHO only makes sense when you _always_ know how to handle them. That is definitely not the case here.

DX will go down the drain as you will need to sprinkle your code with `.mapError(e => new RuntimeException(e))`. And that is only _after_ you realized that the compile errors are because of ZIO error channel type mismatches. Really, it sounds nice, but it only adds friction.

svroonland:
Fair point. There should at least be a `toException` or something on `ConsumerError`. Let me think about it a bit more

PR #1367, opened 2024-11-10
More rebalance safe commits logging alt

erikvanoosten:


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -154,45 +162,57 @@ private[consumer] final class Runloop private (
...
               } yield StreamCompletionStatus(stream.tp, isDone, lastPulledOffset.map(_.offset), endOffsetCommitStatus)
             }
         } yield streamResults
 
-      def logInitialStreamCompletionStatuses: ZIO[Any, Nothing, Unit] =
-        getStreamCompletionStatuses(newCommits = Chunk.empty).flatMap { completionStatuses =>
-          val statusStrings = completionStatuses.map(_.toString)
-          ZIO.logInfo(s"Waiting for ${streamsToEnd.size} streams to end: ${statusStrings.mkString("; ")}")
-        }
+      @inline
```
erikvanoosten, line 177: Not sure why I put that `inline` here. Feel free to remove it.

svroonland (approved):
Nice. As mentioned before, after merging we should try to refactor a lot of this stuff to a separate class & file, to reduce the size of the Runloop somewhat.

PR #1374, opened 2024-11-10
Brag about being faster than raw java kafka

erikvanoosten:


erikvanoosten:
Cheered to early! The speed up is probably caused by different configurations between the benchmarks. We'll make those equal and then write some new texts.

PR #1375, opened 2024-11-10
Extract Committer and RebalanceCoordinator classes from Runloop + unit tests

svroonland:
The code is mostly just moved to a different place, the logic was left mostly intact. The 'interface' between the components has been decoupled more, i.e. the rebalance listener no longer access the full Runloop's State and the pending commits are stored internally in the Committer. 

Care has been taken to make the Committer usable during rebalancing as well, with the proper access to the Consumer for example. The part that waits the end of the streams and their commits has been changed to use the Committer.

Includes unit tests for the two new components.


zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopRebalanceListener.scala
```diff
@@ -0,0 +1,289 @@
...
+      ZIO.attempt(consumer.commitSync(java.util.Collections.emptyMap(), commitTimeout))
+
+    // Outline:
+    // - Every `commitQueuePollInterval` until the deadline has been reached:
+    //   - Get all commits from the commit queue.
+    //   - Start an async commit for these commits.
+    // - Collect all these new (pending) commits.
+    // - repeat the above until:
+    //   - All streams that were ended have completed their work, and
+    // (TODO is the pending part still true? Code did not reflect that)
```
svroonland, line 140: For the reviewer

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Committer.scala
```diff
@@ -0,0 +1,207 @@
...
+) {
+  private val commitTimeout = settings.commitTimeout
+
+  /** This is the implementation behind the user facing api `Offset.commit`. */
+  val commit: Map[TopicPartition, OffsetAndMetadata] => Task[Unit] =
+    offsets =>
+      for {
+        p <- Promise.make[Throwable, Unit]
+        startTime = java.lang.System.nanoTime()
+        _ <- commitQueue.offer(Commit(java.lang.System.nanoTime(), offsets, p))
```
erikvanoosten, line 36:
```suggestion
        startTime = java.lang.System.nanoTime()
        _ <- commitQueue.offer(Commit(startTime, offsets, p))
```

erikvanoosten (commented):
The idea of extracting the `Committer` is really nice. Without this, extracting the rebalance listener would be very hard.
I don't see any thing obviously wrong but since so much code moved around this is a huge PR. So I will have to take a deeper look once this PR exists draft.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopRebalanceListener.scala
erikvanoosten: Yes, this is still true. We wait until the last offset has been committed, and by that we mean 'commitAsync' has been called.  Then at the end we do a single sync commit which blocks until all pending commits are complete.

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
```diff
@@ -1399,9 +1400,13 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
       }: _*) @@ TestAspect.nonFlaky(2),
       test("running streams don't stall after a poll timeout") {
         for {
-          topic      <- randomTopic
-          clientId   <- randomClient
-          _          <- ZIO.fromTry(EmbeddedKafka.createCustomTopic(topic))
+          topic    <- randomTopic
+          clientId <- randomClient
+          _ <- ZIO.attempt(EmbeddedKafka.createCustomTopic(topic)).flatMap(ZIO.fromTry(_)).retryN(3).catchSome {
```
erikvanoosten, line 1405: Why is the retrying needed here, and not in other tests?

svroonland: I couldn't figure that out either..

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Committer.scala
```diff
@@ -0,0 +1,86 @@
...
+
+  def queueSize: UIO[Int]
+
+  def pendingCommitCount: UIO[Int]
+
+  def getPendingCommits: UIO[CommitOffsets]
+
+  def updatePendingCommitsAfterPoll: UIO[Unit]
+
+  def pruneCommittedOffsets(assignedPartitions: Set[TopicPartition]): UIO[Unit]
```
erikvanoosten, line 40:
```suggestion
  def keepCommitsForPartitions(partitions: Set[TopicPartition]): UIO[Unit]
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -612,9 +286,7 @@ private[consumer] final class Runloop private (
 
                                 // Remove committed offsets for partitions that are no longer assigned:
                                 // NOTE: the type annotation is needed to keep the IntelliJ compiler happy.
-                                _ <-
-                                  committedOffsetsRef
-                                    .update(_.keepPartitions(updatedAssignedStreams.map(_.tp).toSet)): Task[Unit]
+                                _ <- committer.pruneCommittedOffsets(updatedAssignedStreams.map(_.tp).toSet): Task[Unit]
```
erikvanoosten, line 289: Not relevant for this PR, but why are we doing this? If there is a fiber that is awaiting completion of a commit, it will wait forever, right?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -840,13 +513,19 @@ private[consumer] final class Runloop private (
...
+      state.subscriptionState.isSubscribed && (state.pendingRequests.nonEmpty || pendingCommitCount > 0 || state.assignedStreams.isEmpty)
+    }
+
   private def observeRunloopMetrics(runloopMetricsSchedule: Schedule[Any, Unit, Long]): ZIO[Any, Nothing, Unit] = {
     val observe = for {
       currentState     <- currentStateRef.get
       commandQueueSize <- commandQueue.size
-      commitQueueSize  <- commitQueue.size
+      commitQueueSize  <- committer.queueSize
+      pendingCommits   <- committer.pendingCommitCount
```
erikvanoosten, line 526:
```suggestion
      pendingCommitCount   <- committer.pendingCommitCount
```

erikvanoosten (approved):
I was a bit disappointed that cleaner code gives _more_ code (this PR effectively adds 593 lines of code). Otherwise I am happy with the split.

It is no longer directly obvious which code executes on the same-thread-runtime (with its restrictions in ZIO operators). Therefore, I propose that methods that are used from the rebalancer will be clearly marked. I proposed a few changes for that, but perhaps there are more.

Approving, but please look at the suggestions.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RebalanceCoordinator.scala
```diff
@@ -0,0 +1,292 @@
...
+              committedOffsets.get(stream.tp),
+              endOffsetCommitStatus
+            )
+          }
+      } yield streamResults
+
+    @inline
+    def logStreamCompletionStatuses(completionStatuses: Chunk[StreamCompletionStatus]): UIO[Unit] = {
+      val statusStrings = completionStatusesAsString(completionStatuses)
+      ZIO.logDebug(
```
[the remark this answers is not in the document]
svroonland: True, it's quite verbose at the moment. Could you create an issue for it?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RebalanceCoordinator.scala
```diff
@@ -0,0 +1,292 @@
...
+          .forever
+          // Even if there is nothing to commit, continue to drive communication with the broker
+          // so that commits can complete and the streams can make progress, by setting
+          // executeOnEmpty = true
+          .tap(_ =>
+            committer.processQueuedCommits(
+              (offsets, callback) => ZIO.attempt(consumer.commitAsync(offsets, callback)),
+              executeOnEmpty = true
+            )
+          )
```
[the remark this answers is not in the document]
svroonland: That doesn't work when it's a ZStream like here :)

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
svroonland: I think you wrote that originally, I'm not sure 😅

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RebalanceCoordinator.scala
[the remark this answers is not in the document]
erikvanoosten: Okay, will do.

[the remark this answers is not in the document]
erikvanoosten: https://github.com/zio/zio-kafka/issues/1405

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RebalanceCoordinator.scala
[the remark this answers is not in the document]
erikvanoosten: Ah! I missed that.

Then at least please use `{}` for multi line blocks.
```suggestion
          .tap { _ =>
            committer.processQueuedCommits(
              (offsets, callback) => ZIO.attempt(consumer.commitAsync(offsets, callback)),
              executeOnEmpty = true
            )
          }
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: Git blame probably points to me indeed, but as far as I remember I simply tried to maintain what it was already doing. I am only questioning it now because I know zio-kafka a lot better now.

svroonland: I'd say it is not strictly necessary, it just prevents offsets being in there for partitions that are no longer assigned to this consumer. That wouldn't affect the logic. I can remove it in a followup PR so that we're being explicit about this functional change apart from this refactoring.

erikvanoosten: I'll create an issue to track follow up.

erikvanoosten: No issue needed, I figured it out. The promise that a committing fiber is waiting for will be invoked directly from the kafka callback, it is not communicated via the stored pending-commits so there is no problem. We keep the pending commits only for 1) knowing there are outstanding commits (determines if we should continue polling), and 2) for metrics (perhaps I missed a use case).

Issue #853, opened 2023-05-19
Discussion: Trying to understand the motivations for the Serializer API


lukestephenson:
Below I'm listing a couple of things that don't feel right to me about the Serializer API.  I'm not expecting anyone to go and change the API for me, I'm happy to do that. This was more to see if other contributors / users are in agreement with the points before I take the discussion further or submit a PR which isn't welcome and find out there are good reasons for the current API design. Thanks

# Suggestion 1: `Serializer` trait rework

Currently we have:
```
trait Serializer[-R, -T] {
  def serialize(topic: String, headers: Headers, value: T): RIO[R, Array[Byte]]
```
And note that `RIO` is an alias for `ZIO[R, Throwable, A]`

Serialization should not fail. Look at libraries like Circe and zio-json (https://zio.dev/zio-json/) - encoding to json always works, we have compiler guarantees that all data can be encoded. Obviously the same can't be said for deserialization where the inbound data is outside of our control.

# Suggestion 2 - Serializers / Deserializers should be safe by default

I'm doing Scala / FP I like avoiding bugs by taking advantage of a powerful type system. A building block for me in this regards is making use of Option to avoid any nasty NPEs. Give me this first before an effect system.  Anyway, the Serializer / Deserializers invoked by zio-kafka can currently pass in null because the base representation of ConsumerRecord / ProducerRecord is the standard Java implementation. It's not safe by default. For example, if I want to handle that a key / value then I have to remember to add `.asOption` to my Deserializer. I feel like the safe behaviour, and highlighting that Kafka doesn't guarantee the key / value are present should be the default, and if I want to acknowledge that I'm happy to discard that type safety (of having `Option[V]` as the default), then I tell the Deserializer it is ok by calling `.valueWillAlwaysExist` (or some better method name).

Even on my `Serializer` I have to call `asOption`. And this is in my nice Scala code where I represent all my models with Option if appropriate, but throw that away at the last minute on the boundary to working with Java (ie when creating the kafka ProducerRecord). Then I get a callback which can potentially fail because the type safety has been given up too early.

erikvanoosten:
Discussions around zio-kafka usually take place on Discord: https://discord.com/channels/629491597070827530/629497941719121960 Would you mind placing your question there?

Re. 'Serialization should not fail', serialization might depend on external systems. For example, you may have to fetch an id from a database.

[...]

PR #1380, opened 2024-11-12
Non-effectful serialization and deserialization

svroonland:
This is an exploration of #853.

Serializers cannot fail, like in zio-json. Deserialization is no longer allowed to be side-effectful, so serializers like the Confluent Avro serde (which communicates with a schema registry) will have to be run in the user's own code. We should add an example to the documentation.

I thought a free error type for the deserialization error would be nice, but at some point it has to be integrated with other errors in the Runloop, which means wrapping it in an exception for now. Needs something better..


svroonland:
> I thought a free error type for the deserialization error would be nice, but at some point it has to be integrated with other errors in the Runloop, which means wrapping it in an exception for now. Needs something better..

May revisit this at some point in the future, closing for now

PR #1381, opened 2024-11-13
Use same `max.poll.records` in all benchmarks

erikvanoosten:


svroonland (approved):
Should not have a functional effect due to order of assigning the properties, but is clearer

erikvanoosten:
> Should not have a functional effect due to order of assigning the properties, but is clearer

Updated title and description to reflect this.

PR #1382, opened 2024-11-13
Use same amount of messages for all consumer benchmarks

erikvanoosten:
In this change common benchmark code is moved into `ZioBenchmark`, `ConsumerZioBenchmark` and `ProducerZioBenchmark` so that it becomes easier to make the different benchmark comparable.

After running the consumer benchmarks with different number of records, and different records sizes per run, this PR settled on 50k records of ~512 bytes per run for all consumer benchmarks. With these amounts the zio-kafka based benchmarks and the 'comparison' benchmarks have roughly the same scaling elasticity (where 'scaling elasticity' is defined as the throughput growth factor divided by the number of records growth factor).

After this PR is merged, the benchmark history will be rewritten with linear scaling so that we can compare historic runs against new runs.

### DIFF
--- a/zio-kafka-bench/src/main/scala/zio/kafka/bench/ZioBenchmark.scala
+++ b/zio-kafka-bench/src/main/scala/zio/kafka/bench/ZioBenchmark.scala
@@ -5,6 +5,21 @@ import zio.{ ZLayer, _ }
 
 import java.util.UUID
 
+trait ConsumerZioBenchmark[Environment] extends ZioBenchmark[Environment] {
+  protected val messageCount: Int               = 100000
+  protected def messageData(i: Int): String     = i.toString + scala.util.Random.alphanumeric.take(1024).mkString
+  protected val kvs: Iterable[(String, String)] = Iterable.tabulate(messageCount)(i => (s"key$i", messageData(i)))
+  protected val topic1                          = "topic1"
+  protected val partitionCount                  = 6
+}
+
+trait ProducerZioBenchmark[Environment] extends ZioBenchmark[Environment] {
+  protected val messageCount                = 500
+  protected val kvs: List[(String, String)] = List.tabulate(messageCount)(i => (s"key$i", s"msg$i"))
+  protected val topic1                      = "topic1"
+  protected val partitionCount              = 6
+}
+
 trait ZioBenchmark[Environment] {
   var runtime: Runtime.Scoped[Environment] = _
 
--- a/zio-kafka-bench/src/main/scala/zio/kafka/bench/ZioBenchmark.scala
+++ b/zio-kafka-bench/src/main/scala/zio/kafka/bench/ZioBenchmark.scala
@@ -6,8 +6,8 @@ import zio.{ ZLayer, _ }
 import java.util.UUID
 
 trait ConsumerZioBenchmark[Environment] extends ZioBenchmark[Environment] {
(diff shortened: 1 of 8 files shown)


zio-kafka-bench/src/main/scala/zio/kafka/bench/comparison/ComparisonBenchmark.scala
```diff
@@ -21,7 +21,7 @@ trait ComparisonBenchmark extends ZioBenchmark[Env] {
   protected final val nrPartitions: Int = 6
   protected final val topicPartitions: List[TopicPartition] =
     (0 until nrPartitions).map(TopicPartition(topic1, _)).toList
-  protected final val numberOfMessages: Int           = 1000000
+  protected final val numberOfMessages: Int           = 50000
```
svroonland, line 24: Is it still enough to get a stable benchmark result..?

erikvanoosten: Most probably. We deemed it sufficient for the zio-kafka based benchmarks. Historically there is variation in the results, but I suspect its fine.

erikvanoosten: On the other hand, it might be better to raise both to 100_000. I suspect the benchmark duration will not be hugely affected.

svroonland (approved):

erikvanoosten:
This is quite amazing. Doubling the number of messages also almost doubles the throughput! (Both for zio-kafka and the direct consumers.)

That means that the benchmarks are not really representative. We probably need to increase the number of messages further. Also maybe make them larger.

svroonland:
We could also look at latency metrics for various RPS loads

erikvanoosten:
Going from 100k 1024 byte records, to 50k 512 byte records make the benches 3 times faster (not 4 times as could be expected). This true for all consumer benchmarks.

With 50k 512 byte records the benchmark workflow takes about as long as before (which has 50k 10 byte records).

I will merge this change and rewrite bench history a bit so that it all stays (roughly) comparable.

zio-kafka-bench/src/main/scala/zio/kafka/bench/comparison/ComparisonBenchmark.scala
erikvanoosten: 100k messages does affect total runtime while ratio between the different benches does not change. So back to 50k.

zio-kafka-bench/src/main/scala/zio/kafka/bench/ZioBenchmark.scala
```diff
@@ -5,6 +5,21 @@ import zio.{ ZLayer, _ }
 
 import java.util.UUID
 
+trait ConsumerZioBenchmark[Environment] extends ZioBenchmark[Environment] {
+  protected val messageCount: Int               = 50000
+  protected def messageData(i: Int): String     = i.toString + scala.util.Random.alphanumeric.take(512).mkString
```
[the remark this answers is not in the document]
erikvanoosten: Fixed in https://github.com/zio/zio-kafka/pull/1382/commits/b4d4e629564b4c3328732b0ebfacd979033f5a06

PR #1392, opened 2024-11-16
Run profiler only on release and on demand

erikvanoosten:
Since we are not using the cpu/heap profiles that much (if at all), IMHO it is an environmental offense to run the profiler on every commit.

With this change we only run the profiler on releases. In addition, it is now easy to run the profiler on demand from any branch, tag or commit.


.github/workflows/profile.yml
```diff
@@ -24,6 +22,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
+          ref: ${{ github.event.inputs.ref }}
```
svroonland, line 25: Does it supply the right value for `ref` when running a release?

erikvanoosten: The `ref` parameter has default value "" which means take the default branch. This is what we had before as well.

When this action is run from a release the `ref` input it not set. So I am actually not sure it this will work (does this evaluate to empty string when the `ref` is not set?). Perhaps we should give the input parameter an explicit default value of "".

erikvanoosten: I added an explicit default value. So now it should be fine.

svroonland (approved):

PR #1400, opened 2024-11-21
Revert "Update zio, zio-streams, zio-test, ... to 2.1.13"

erikvanoosten:
Reverts zio/zio-kafka#1399 since the tests are failing...

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -42,7 +42,7 @@ lazy val _scala3   = "3.3.4"
 inThisBuild(
   List(
     name         := "ZIO Kafka",
-    zioVersion   := "2.1.13",
+    zioVersion   := "2.1.12",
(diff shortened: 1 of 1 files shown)


svroonland:
Interesting, with 2.1.13 I see this in the logs when I run the "rebalanceSafeCommits prevents processing messages twice when rebalancing" test. Never seen it before, no more context unfortunately:

```
21:47:38.895 [ZScheduler-Worker-4] [consumer=1] DEBUG zio.kafka.consumer.ConsumerLive - Fiber zio-fiber-107 did not handle an error
zio.FiberFailure: Right(())
```

Also when I increase the buffer parameter of `plainStream` from 4 to 32 the test succeeds without timeout..

svroonland (approved):

PR #1403, opened 2024-11-24
Document when zio-kafka is faster than raw java kafka

erikvanoosten:


docs/index.md
```diff
@@ -135,3 +137,15 @@ Want to see your company here? [Submit a PR](https://github.com/zio/zio-kafka/ed
...
 * [Rocker](https://rocker.com)
 
+## Performance
+
+By default, zio-kafka programs process partitions in parallel. The default java Kafka client does not provide parallel
+processing. Of course, there is some overhead in buffering records and distributing them to the fibers that need them.
+On 2024-11-23, we estimated this overhead to be 1.2 ms per 1k records
+(comparing [benchmarks](https://zio.github.io/zio-kafka/dev/bench/) `throughput` and `kafkaClients`, using the standard
+GitHub Action runners (4 cores), and with the Kafka broker in the same JVM). This means that for this particular
+combination, when processing needs more than 1.2 ms per 1k records, a zio-kafka based program will have **higher
```
svroonland, line 147: > when processing needs more than 1.2 ms per 1k records, a zio-kafka based program will have **higher throughput**

I'm not sure we can conclude that. The zio-kafka part will still be 1.2 ms slower per 1k records than fetching it via the plain kafka client, and on top of that you get the extra processing time which will slow the system down. I don't see how the processing time cancels the overhead.

svroonland:
In `ZioKafkaConsumerBenchmark.throughput`:

```
Consumer
             .tap { _ =>
               counter.updateAndGet(_ + 1).flatMap(count => Consumer.stopConsumption.when(count == recordCount))
             }
```

Do you think it would be more fair to replace the `tap` with a `tapChunk`? In the manual kafka equivalent benchmark, we also process records by the batch instead of per record.

docs/index.md
erikvanoosten: In a week or so I'll have time to write a longer article that will hopefully explain my intuition on this 😄

svroonland:
See https://github.com/zio/zio-kafka/pull/1409

docs/index.md
svroonland: Be sure to take the new `throughput` results into account in your new calculation :)

erikvanoosten: I did!

svroonland (approved):

PR #1404, opened 2024-11-24
Convert commitAsync callback handling to ZIO sooner

svroonland:
KafkaConsumer's `commitAsync` takes a callback, which we program against with complicated followup code. This PR attempts to convert everything to ZIO's earlier on, making chaining followup effects easier to reason about.

As this changes some functionality around locking and same / single threads, here's a summary of what do we need to ensure:
* We have an exclusive lock on the consumer when calling `commitAsync`. In `Runloop.run` this is done using `ConsumerAccess`. In the rebalance coordinator (while rebalancing) we already have the lock as we're calling `poll()` so no need for extra locking.
* The consumer is not used from more than one thread at the same time. For use in `Runloop.run` we get this for free by guaranteeing exclusive access. In the rebalance coordinator a `poll()` call is in the middle of being executed and we need to call `commitAsync` on the same thread as the rebalance listener is invoked.

Anything that is not calling commitAsync is free to run on any thread as executed by the default ZIO runtime.


zio-kafka/src/main/scala/zio/kafka/consumer/internal/LiveCommitter.scala
```diff
@@ -20,7 +17,6 @@ private[consumer] final class LiveCommitter(
   consumerMetrics: ConsumerMetrics,
   onCommitAvailable: UIO[Unit],
   committedOffsetsRef: Ref[CommitOffsets],
-  sameThreadRuntime: Runtime[Any],
```
svroonland, line 23: We don't need to call commitAsync on the same thread for regular commits (outside of rebalance), there we just need exclusive access to the consumer. The RebalanceListener has a version of `commitAsync` that is executed on the same thread.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/LiveCommitter.scala
```diff
@@ -44,31 +43,32 @@ private[consumer] final class LiveCommitter(
...
+
+           for {
+             _                <- pendingCommits.update(_ ++ commits)
+             startTime        <- ZIO.clockWith(_.nanoTime)
+             getCommitResults <- commitAsyncZIO(consumer, offsetsWithMetaData)
+             _ <- getCommitResults.exit
+                    .flatMap(handleCommitCompletion(commits, offsetsWithMetaData, startTime, _))
+                    // We don't wait for the completion of the commit here, because it
+                    // will only complete once we poll again.
+                    .forkDaemon
```
erikvanoosten, line 65: I don't like `forkDeamon`. It is hurts performance of your entire zio application. You can see this in the benchmarks, this PR is fairly consistenly 10ms slower than before.

In addition, the fiber should be interrupted when the consumer is stopped. What would forkScoped (attached to the scope of the consumer) do?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/LiveCommitter.scala
```diff
@@ -79,44 +79,72 @@ private[consumer] final class LiveCommitter(
...
       }
-    (offsetsWithMetaData.asJava, callback, onFailure)
-  }
+      .unit
+
+  /**
+   * Wrapper that converts KafkaConsumer#commitAsync to ZIO
+   *
+   * @return
+   *   Task whose completion indicates completion of the commitAsync call. The inner task completes when the callback is
```
erikvanoosten, line 120: Instead of a `Task[Task]`, this method could also return `Task[Promise]` which is a bit easier to reason about and I suspect it will also simplify the caller.

erikvanoosten (commented):
I am not done but I have to go. Will look further another time.

Please be aware that this PR is consistently 10ms slower than `main` (see the benchmarks).

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -492,11 +495,14 @@ private[consumer] final class Runloop private (
...
-                   ZIO.attempt(consumer.commitAsync(offsets, callback))
-                 )
-               }
+          _ <-
+            consumer.runloopAccess { consumer =>
+              committer
+                .processQueuedCommits(consumer)
+                .onExecutor(
+                  commitExecutor
+                ) // processQueuedCommits does a fork to await the commit callback, which is not supported by the single-thread executor
```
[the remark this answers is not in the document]
svroonland: It does, or am I misreading this?

```
      // Run the entire loop on a dedicated thread to avoid executor shifts
      executor <- RunloopExecutor.newInstance
      fiber    <- ZIO.onExecutor(executor)(runloop.run(initialState)).forkScoped
```

```
  val newInstance: URIO[Scope, Executor] = newSingleThreadedExecutor
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -492,11 +496,12 @@ private[consumer] final class Runloop private (
...
-                 committer.processQueuedCommits((offsets, callback) =>
-                   ZIO.attempt(consumer.commitAsync(offsets, callback))
-                 )
-               }
+          _ <-
+            committer
+              .processQueuedCommits(commitAsyncZIO)
+              .onExecutor(
+                commitExecutor
+              ) // processQueuedCommits does a fork to await the commit callback, which is not supported by the single-thread executor
```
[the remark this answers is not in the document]
erikvanoosten: There are 2 specialized executors/runtimes:
* singleThreadExecutor, used by RunLoop.run
* sameThreadRuntime, used by the rebalance listener, but effectively it runs on the same thread as provided by singleThreadExecutor

So there is a good reason to do the commit handling on another executor (the default executor is just fine), but it is not because the singleThreadExecutor does not support it.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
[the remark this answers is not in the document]
erikvanoosten: There are 2 specialized executors/runtimes:
* singleThreadExecutor, used by RunLoop.run
* sameThreadRuntime, used by the rebalance listener, but effectively it runs on the same thread as provided by singleThreadExecutor

So there is a good reason to do the commit handling on another executor (the default executor is just fine), but it is not because the singleThreadExecutor does not support it.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/LiveCommitter.scala
svroonland: Agreed, I don't like it either. What do we actually prefer, running the stuff on the kafka thread or on a ZIO thread? We could also shove completion results in a queue from the kafka thread and then in a regular forked fiber process that queue.

erikvanoosten: As far as I remember the commit callback is not doing a lot of work, it only completes a promise (right?). In that case I'd prefer to run it directly on the runloop thread (so no forking needed). It there was more work involved it might be beneficial to run on the default ZIO executor (but then using forkScoped instead of forkDaemon).

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -497,11 +497,10 @@ private[consumer] final class Runloop private (
           _ <- ZIO.logDebug(s"Processing ${commands.size} commands: ${commands.mkString(",")}")
           _ <-
             consumer.runloopAccess { consumer =>
+              // processQueuedCommits does forks to await commit callbacks, move those to another executor to keep the runloop executor clear for the runloop
               committer
                 .processQueuedCommits(consumer)
-                .onExecutor(
-                  commitExecutor
-                ) // processQueuedCommits does a fork to await the commit callback, which is not supported by the single-thread executor
+                .onExecutor(commitExecutor)
```
erikvanoosten, line 503: Perhaps we could try to run the commit stuff on the runloop thread again (by removing this line)? Perhaps the extra coordination causes the benchmarks to a bit slower than before.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/LiveCommitter.scala
svroonland: Done

zio-kafka/src/main/scala/zio/kafka/consumer/internal/LiveCommitter.scala
svroonland: Different approach now

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: No longer relevant.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/LiveCommitter.scala
```diff
@@ -79,44 +79,70 @@ private[consumer] final class LiveCommitter(
...
+  /**
+   * Wrapper that converts KafkaConsumer#commitAsync to ZIO
+   *
+   * @return
+   *   Task whose completion indicates completion of the commitAsync call.
+   */
+  private def commitAsyncZIO(
+    consumer: ByteArrayKafkaConsumer,
+    offsets: Map[TopicPartition, OffsetAndMetadata],
+    doOnComplete: Either[Exception, Map[TopicPartition, OffsetAndMetadata]] => UIO[Unit]
```
erikvanoosten, line 124: In this PR we now always provide a callback to `commitAsync`, even when the given list of offsets is empty (so a callback is not neccesary). During a rebalance, this method is called _a lot_ (every 100ms) and it is likely that most of these calls do not need a callback. 

WDYT of making `doOnComplete` an `Option`, and then when it's `None` we pass `null` to `commitAsync` like we did before?

erikvanoosten (approved):
Looks good to me!

zio-kafka/src/main/scala/zio/kafka/consumer/internal/LiveCommitter.scala
erikvanoosten: On the other hand, doing this removes the ability to react to errors... 🤔

svroonland: Ah yeah I totally missed this. But I don't think it would make a large performance difference if it's only once every 100 ms.

PR #1407, opened 2024-11-26
[Internal] Micro-optimisation: Reuse previously calculated `java.langSystem.nanoTime`

guizmaii:
Hey :)

I'm not contributing much anymore. Sorry about that. I lost the motivation, TBH, but I really enjoy following what you continue to do 👏

I was passing by and was reading some of the new code when I saw this. LMKWYT :)

With this change, we lose some nanoseconds of precision, so it might not be the expected behaviour. 🤔


svroonland (approved):
Hey Jules, good to hear from you again. I understand, motivation fluctuates and interests shift over time, it's familiar to me (in a bit of a contributing spree right now). Feel free to chip in any time in the future of course!

The change looks good to me

svroonland:
It looks like the whole `createdAt` field in `Commit` is never used, actually

erikvanoosten:
Isn't this PR colliding with #1375?

PR #1409, opened 2024-11-30
ZioKafkaConsumerbenchmark.throughput: increase count per batch

svroonland:
This is more consistent with the manual kafka client benchmark equivalent

### DIFF
--- a/zio-kafka-bench/src/main/scala/zio/kafka/bench/ZioKafkaConsumerBenchmark.scala
+++ b/zio-kafka-bench/src/main/scala/zio/kafka/bench/ZioKafkaConsumerBenchmark.scala
@@ -46,7 +46,7 @@ class ZioKafkaConsumerBenchmark extends ConsumerZioBenchmark[Kafka with Producer
              .tap { batch =>
                counter
                  .updateAndGet(_ + batch.size)
-                 .flatMap(count => Consumer.stopConsumption.when(count == recordCount))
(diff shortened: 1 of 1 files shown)


zio-kafka-bench/src/main/scala/zio/kafka/bench/ZioKafkaConsumerBenchmark.scala
```diff
@@ -42,8 +42,11 @@ class ZioKafkaConsumerBenchmark extends ConsumerZioBenchmark[Kafka with Producer
       counter <- Ref.make(0)
       _ <- Consumer
              .plainStream(Subscription.topics(topic1), Serde.byteArray, Serde.byteArray)
-             .tap { _ =>
-               counter.updateAndGet(_ + 1).flatMap(count => Consumer.stopConsumption.when(count == recordCount))
+             .chunks
+             .tap { batch =>
+               counter
+                 .updateAndGet(_ + batch.size)
+                 .flatMap(count => Consumer.stopConsumption.when(count == recordCount))
```
erikvanoosten, line 49: Since its per batch, I think we should make this change as well:
```suggestion
                 .flatMap(count => Consumer.stopConsumption.when(count >= recordCount))
```

erikvanoosten (approved):

svroonland:
Seems to speed up the benchmark a bit as well

PR #1411, opened 2024-12-03
Fork offering results to stream

svroonland:
Performance experiment


erikvanoosten:
That simple 🙂 
BTW, the benchmarks timed out. Probably because we need to shift this work to another executor (e.g. the default one). If we do a fork from the single thread executor, the task will need to wait until the runloop yields.

erikvanoosten:
I hope I fixed the benchmarks.

Note: this must never be merged. We need to guarantee that records are offered in order. This is a nice experiment though!

erikvanoosten:
afbeelding
Doesn't speed it up at all ... 😞

erikvanoosten:
As it doesn't seem to work, shall we close this issue?

PR #1418, opened 2024-12-15
Fix scala-steward flow

erikvanoosten:
Scala-steward needs sbt installed, also the new example template gives it additional permissions.


svroonland (approved):
LGTM

PR #1425, opened 2024-12-29
Support rebalance-safe-commits with external commits

svroonland:
Allow rebalance-safe-commits to be used in combination with external commits. External commits are commits done to some other system than the kafka broker, e.g. a relational database.

This new capability also supports the improved `TransactionalProducer` that will be introduced in zio-kafka 3.0.0.

### DIFF
--- a/zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
+++ b/zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
@@ -28,6 +28,8 @@ import zio.test.Assertion._
 import zio.test.TestAspect._
 import zio.test._
 
+import java.time.Instant
+import java.time.temporal.ChronoUnit
 import java.util.concurrent.ExecutionException
 import scala.reflect.ClassTag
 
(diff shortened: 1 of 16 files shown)


erikvanoosten:
This is nice. Some questions:
* Can we somehow guarantee that `rebalanceSafeCommits` is enabled when we use this new transactional producer?
* If we simply roll this out, will applications with the existing transactional producers break? Maybe not, but if that is so, I propose we either make this backward compatible, or we move this to zio-kafka 3.

erikvanoosten:
The updated `TransactionProducer` is not binary backward compatible; it requires a `Consumer` in the environment, so the ZIO type is different. It will be source compatible for many applications because many will have a `Consumer` available anyway.

However, this is a strange set-up. The `Consumer` and the `TransactionProducer` are tied together; they must connect to the same Kafka broker cluster. Therefore it makes more sense to produce them together, or at least let one refer to the other.

I propose that we change the construction of `TransactionProducer`s such that they require a `Consumer`.
This has some advantages:
 - Method `TransactionProducer.createTransaction` stays backward compatible because its type doesn't change.
 - We could even make it fully backward compatible: if a consumer is provided we do the new thing pioneered in this PR, if its not provided, nothing changes.
 - We may even be able to check that the provided `Consumer` connects to the same broker as the `TransactionProducer` that is being constructed.

Re. method `Consumer.registerOffsetsCommittedInTransaction`, this is an internal method and should not be exposed too easily. We could do this by creating a separate trait (e.g. something like `ConsumerInternal`) and let `ConsumerLive` extend both traits. This is not the best way, we can probably think of something better.

erikvanoosten:
What is `Consumer.markCommittedInTransaction` actually used for? I don't see where it is used.

svroonland:
Agreed, providing the `Consumer` upon `TransactionalProducer` creation is better. I think it's fine to disregard backwards compatibility here, since the old method was very complicated and not well supported, this is such an improvement in usability.

If we want to check the settings, we could have a `Consumer.settings` method to retrieve the settings with which the Consumer was created. Is that what you have in mind as well?

There is no `Consumer.markCommittedInTransaction`, only `Committer.markCommittedInTransaction`. I'm not sure how we can hide this method, but agree that this is not ideal. In an earlier commit this was done with an internal method on the `OffsetBatch`, I guess we could revert to that.

erikvanoosten:
> I think it's fine to disregard backwards compatibility here...
Yeah, that is a good point. We'd better make very good release notes then :)

> we could have a Consumer.settings method ... Is that what you have in mind as well?
Not necessarily, but I guess that is what is needed indeed.

> There is no `Consumer.markCommittedInTransaction`, only `Committer.markCommittedInTransaction`.

Check, I get it now. Well done.

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -604,4 +609,9 @@ private[consumer] final class ConsumerLive private[consumer] (
   override def metrics: Task[Map[MetricName, Metric]] =
     consumer.withConsumer(_.metrics().asScala.toMap)
 
+  override def registerOffsetsCommittedInTransaction(
+    offsetBatch: OffsetBatch
+  ): Task[Unit] =
+    runloopAccess.withRunloopZIO(true)(runloop => runloop.registerOffsetsCommittedInTransaction(offsetBatch))
```
erikvanoosten, line 615: We can consider adding method `registerOffsetsCommittedInTransaction` to `runloopAccess`. This would be more in line with how ConsumerLive uses RunloopAccess in other places.

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -158,6 +158,11 @@ trait Consumer {
    * Expose internal consumer metrics
    */
   def metrics: Task[Map[MetricName, Metric]]
+
+  /**
+   * Used internally by the [[zio.kafka.producer.TransactionalProducer]]
+   */
```
erikvanoosten, line 164:
```suggestion
  /**
   * This is not a public API.
   * It is used internally by the [[zio.kafka.producer.TransactionalProducer]].
   */
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/LiveCommitter.scala
```diff
@@ -23,6 +23,12 @@ private[consumer] final class LiveCommitter(
   pendingCommits: Ref.Synchronized[Chunk[Commit]]
 ) extends Committer {
 
+  override val markCommittedInTransaction: Map[
+    TopicPartition,
+    OffsetAndMetadata
+  ] => Task[Unit] = offsets =>
+    committedOffsetsRef.modify(_.addCommits(Chunk(Commit(java.lang.System.nanoTime(), offsets, null)))).unit
```
erikvanoosten, line 30: Let's put this below `val commit` to keep the order of the trait.

```suggestion
    committedOffsetsRef
      .modify {
        // The continuation promise can be `null` because this commit is not handled by the consumer,
        // and transactional commits have no completion handler anyway.
        _.addCommits(Chunk(Commit(java.lang.System.nanoTime(), offsets, null)))
      }
      .unit
```

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
```diff
@@ -1234,96 +1234,67 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
...
-                          clientId,
-                          consumerGroupId,
-                          restartStreamOnRebalancing = true,
-                          properties = Map(
-                            ConsumerConfig.PARTITION_ASSIGNMENT_STRATEGY_CONFIG ->
-                              implicitly[ClassTag[T]].runtimeClass.getName,
-                            ConsumerConfig.MAX_POLL_RECORDS_CONFIG -> "200"
-                          ),
-                          rebalanceListener = transactionalRebalanceListener(streamCompleteOnRebalanceRef)
+                ZIO.scoped {
```
erikvanoosten, line 1246: Would this extra `scoped` fix the spurious test failures?

erikvanoosten (approved):
Great work! Some nitpicks in the comments.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/LiveCommitter.scala
erikvanoosten: I updated the proposed comment.

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
erikvanoosten: There is another approach. We can also expose this method intentionally because it opens up a new use case: using rebalance-safe-commits in combination with non-kafka commits! In fact, IMHO this is such a nice use that it warrants exposing this method in a separate PR (even pre-3.0)!

erikvanoosten: With the previous idea, we could rename the method. E.g. to something like `registerExternalCommits`.

erikvanoosten:
As discussed on Discord, we pivot this PR to support rebalance-safe-commits with external commits. The transactional producer changes are moved to a future PR.

erikvanoosten:
The changes in transactional producing have been pushed in the https://github.com/zio/zio-kafka/tree/transaction-rebalance-safe-commits-followup branch.

svroonland (commented):
👍

svroonland:
So this can be merged and released in 2.10.0?

erikvanoosten:
> So this can be merged and released in 2.10.0?

Yes! Though that reminds me, we neer to change the version in the docs.

erikvanoosten (approved):

PR #1432, opened 2025-01-09
Document transactions

erikvanoosten:
Document the transactional producer introduced in #1434.

### DIFF
--- a/docs/transactions.md
+++ b/docs/transactions.md
@@ -33,39 +33,55 @@ val settings = ConsumerSettings(bootstrapServers)
 In order to produce records transactionally, we need a `Consumer` and a `TransactionalProducer` that will work together
 to safely commit transactions.
 
-First we create layers for the `ConsumerSettings` and the `Consumer`.
+On this page we assume you're using [ZIO service pattern](https://zio.dev/reference/service-pattern/) and that the
+`Consumer` and `TransactionalProducer` are created as a layer.
+
+First we create the `Consumer` layer.
 Note that rebalance-safe-commits must be enabled (see the background section below for more information).
 Since the default `maxRebalanceDuration` is quite high, you may also consider setting it to a lower duration, one in
 which you are sure all records of a poll can be processed.
 
 ```scala
 val bootstrapBrokers = List("localhost:9092")
-val consumerSettings = ZLayer.succeed {
-  ConsumerSettings(bootstrapBrokers)
+
+val consumerLayer: ZLayer[Any, Throwable, Consumer] = ZLayer {
+  val consumerSettings = ConsumerSettings(bootstrapBrokers)
     .withGroupId("somegroupid")
     .withRebalanceSafeCommits(true) // required!
     .withMaxRebalanceDuration(30.seconds)
+  Consumer.make(consumerSettings)
 }
-val consumer = Consumer.live
 ```
 
-Now we can create layers for the `ProducerSettings` and the `TransactionalProducer`.
+Now we can create the `TransactionalProducer` layer.
 Note that the producer connects (and must connect) to the same brokers as the consumer.
 
-DEVELOPER NOTE: here we assume that we no longer have `TransactionalProducerSettings`. TODO: check this assertion
-
 ```scala
-val producerSettings = ZLayer.succeed {
-  ProducerSettings(bootstrapServers = bootstrapBrokers)
-}
-val producer: ZLayer[TransactionalProducerSettings with Consumer, Throwable, TransactionalProducer] =
-  TransactionalProducer.live
+val producerLayer: ZLayer[Consumer, Throwable, TransactionalProducer] =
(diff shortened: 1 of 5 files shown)


docs/transactions.md
```diff
@@ -0,0 +1,257 @@
...
+[Kafka transactions](https://www.confluent.io/blog/transactions-apache-kafka/) are different from what you may know from
+relational databases. In Kafka a transaction means that your program consumes some records, then produces some other
+records (to the same broker) and then by committing the _consumed_ records the transaction is committed, and the
+produced records become available to other consumers.
+
+## A warning about consuming transactional records
+
+By default, produced records become immediately visible for consumers, even if they are produced in a not yet committed
+Kafka transaction! To prevent reading transactional records before they are committed, you need to configure every
+consumer with the `isolation.level` property set to `read_committed`. For example as follows:
```
svroonland, line 19: Nice, did not know it supported/required that.

docs/transactions.md
```diff
@@ -0,0 +1,257 @@
...
+        new ProducerRecord("my-output-topic", key, newValue)
+      (producerRecord, record.offset)
+    }
+  }
+```
+
+Typically, to optimize throughput, we want to produce records in batches. The underlying chunking structure of the
+consumer stream is ideal for that because zio-kafka guarantees that each chunk in the stream corresponds to the records
+that were fetched together. However, we need to be careful to retain the chunking structure. For example, we should
+not use `.map` because it results in a stream where each chunk contains only a single item. Therefore, we use
```
svroonland, line 94: `map` is fine, but `mapZIO` breaks chunking, IIRC

docs/transactions.md
```diff
@@ -0,0 +1,257 @@
...
+
+```scala
+val producerSettings = ZLayer.succeed {
+  ProducerSettings(bootstrapServers = bootstrapBrokers)
+}
+val producer: ZLayer[TransactionalProducerSettings with Consumer, Throwable, TransactionalProducer] =
+  TransactionalProducer.live
+```
+
+Notice that the `producer` layer also requires a `Consumer`. This is the consumer that will be used to commit the
```
svroonland, line 62: This may be confusing, the consumer is actually only used to track which offsets were committed in the transaction as part of the rebalance process, it does not do the commits.

docs/transactions.md
```diff
@@ -0,0 +1,257 @@
...
+epoch of when the records were consumed. If the broker detects a commit with a group epoch that is not equal to the
+current epoch, it will reject the commit.
+
+In addition, to prevent an old consumer from continuing to poll into the next epoch without participating in the
+rebalance, consumers polling within the wrong group epoch are 'fenced'. A fenced consumer dies with a
+`FencedInstanceIdException`.
+
+### How zio-kafka helps
+
+With zio-kafka you can easily couple the transactional producer to a consumer to commit transactions. ZIO scopes are
```
svroonland, line 248: Same here, does not actually commit

zio-kafka-example/src/main/scala/zio/kafka/example/Transactional.scala
```diff
@@ -0,0 +1,92 @@
...
+import zio._
+import zio.kafka.consumer.diagnostics.Diagnostics
+import zio.kafka.consumer._
+import zio.kafka.producer._
+import zio.kafka.serde.Serde
+import zio.logging.backend.SLF4J
+
+import java.util.UUID
+
+object Transactional extends ZIOAppDefault {
```
svroonland, line 13: Maybe `TransactionalProducing`

svroonland (commented):
Really nice and clear documentation! A few remarks

docs/transactions.md
erikvanoosten: Thanks, somehow I knew that and I still missed that. I'll refine the text.

docs/transactions.md
erikvanoosten: Fixed!

docs/transactions.md
erikvanoosten: Fixed!

docs/transactions.md
```diff
@@ -0,0 +1,259 @@
...
+    .withRebalanceSafeCommits(true) // required!
+    .withMaxRebalanceDuration(30.seconds)
+}
+val consumer = Consumer.live
+```
+
+Now we can create layers for the `ProducerSettings` and the `TransactionalProducer`.
+Note that the producer connects (and must connect) to the same brokers as the consumer.
+
+DEVELOPER NOTE: here we assume that we no longer have `TransactionalProducerSettings`. TODO: check this assertion
```
erikvanoosten, line 52: We need to evaluate if we still need the user to provide `transactionalId`. If not, we use `ProducerSettings`, if she does, we continue to need `TransactionalProducerSettings`.

svroonland (approved):

PR #1433, opened 2025-01-11
Add config for transactional consuming

erikvanoosten:
Make it easier to configure read-committed.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
@@ -1,6 +1,7 @@
 package zio.kafka.consumer
 
 import org.apache.kafka.clients.consumer.ConsumerConfig
+import org.apache.kafka.common.IsolationLevel
 import zio._
 import zio.kafka.consumer.Consumer.OffsetRetrieval
 import zio.kafka.consumer.fetch.{ FetchStrategy, QueueSizeBasedFetchStrategy }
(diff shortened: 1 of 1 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -330,6 +331,19 @@ final case class ConsumerSettings(
...
 
+  /**
+   * Controls how to consume records produced transactionally.
+   *
+   * @param read_committed
+   *   when `true`, only consume records which have been committed, when `false`, consume all records, even records
+   *   which are part of an aborted transaction. Non-transactional records will be consumed unconditionally in either
+   *   mode.
+   */
+  def withConsumeTransactionalOnly(read_committed: Boolean = true): ConsumerSettings = {
```
svroonland, line 342: I can see you started doing python ;) How about `readCommitted`

erikvanoosten: OMG!

erikvanoosten: Fixed.
Actually, it wasn't Python, but I copied in the Kafka config property name...

zio-kafka/src/main/scala/zio/kafka/consumer/ConsumerSettings.scala
```diff
@@ -330,6 +331,19 @@ final case class ConsumerSettings(
...
 
+  /**
+   * Controls how to consume records produced transactionally.
+   *
+   * @param readCommitted
+   *   when `true`, only consume records which have been committed, when `false`, consume all records, even records
+   *   which are part of an aborted transaction. Non-transactional records will be consumed unconditionally in either
+   *   mode.
+   */
+  def withConsumeTransactionalOnly(readCommitted: Boolean = true): ConsumerSettings = {
```
svroonland, line 342: Regarding the name, what do you think of `withIsolationLevel(readCommitted: Boolean = true)`?

erikvanoosten: I considered that name as well and if you prefer that, I can live with it.
The reason I chose `withConsumeTransactionalOnly` because it is more descriptive of what actually happens. I also remember that when I was a rookie I did not couple the words 'isolation level' and 'transaction' together, it would have been harder to find this method.

svroonland: Naming is always hard and somewhat subjective, but maybe it's best to stick to the concept naming as can be found in other kafka documentation, so `withIsolationLevel`

erikvanoosten: Okay, updated.

erikvanoosten: With this change, it no longer makes sense to have true as the default.

`consumerSettings.withIsolationLevel()` is not okay

it should be:
`consumerSettings.withIsolationLevel(true)`

or better yet:
`consumerSettings.withIsolationLevel(readCommitted = true)`

(This was another advantage of the previous name.)

svroonland: > consumerSettings.withIsolationLevel()

Ah right, that doesn't feel like the best use of the English language. I can live with `withReadCommitted`

erikvanoosten: Fixed.

svroonland (approved):

PR #1434, opened 2025-01-12
Simplify transactional producing

svroonland:
Introduce a much simpler way to use the transactional producer. It is no longer needed to install a complex rebalance listener, coupling your `TransactionalProducer` to a `Consumer`is enough.

See #1432 for documentation.


svroonland:
@erikvanoosten Help remind me, we want this one for 3.0 right, because of the incompatibility? Also, was there anything keeping this in draft state? 😅

erikvanoosten:
> @erikvanoosten Help remind me, we want this one for 3.0 right, because of the incompatibility? Also, was there anything keeping this in draft state? 😅

Correct, it is not binary/source compatible.
I kept things for 3.0.0 in draft to prevent it from being merged. But this PR is good to go (for 3.0.0). Do you prefer to remove the draft label?

svroonland:
Yeah, if we create a `series/3.0` branch or something we can merge it.

PR #1439, opened 2025-01-15
Deprecate `restartStreamOnRebalancing`

erikvanoosten:
The feature `restartStreamOnRebalancing` is no longer needed in zio-kafka 3.x as transactional producing will work without it.

### DIFF
--- a/zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
+++ b/zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
@@ -128,6 +128,10 @@ object KafkaTestUtils {
     clientInstanceId: Option[String] = None,
     allowAutoCreateTopics: Boolean = true,
     offsetRetrieval: OffsetRetrieval = OffsetRetrieval.Auto(AutoOffsetStrategy.Earliest),
+    // @deprecated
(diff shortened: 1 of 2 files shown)


zio-kafka-testkit/src/main/scala/zio/kafka/testkit/KafkaTestUtils.scala
```diff
@@ -136,6 +137,7 @@ object KafkaTestUtils {
     properties: Map[String, String] = Map.empty
   ): URIO[Kafka, ConsumerSettings] =
     ZIO.serviceWith[Kafka] { (kafka: Kafka) =>
+      @nowarn("msg=deprecated")
```
svroonland, line 140: Deprecate the `restartStreamOnRebalancing` parameter in this method as well, instead

erikvanoosten: You can't deprecate a parameter, only the entire method.
To do this, we need make a copy of this method into a trait that this class extends. In the original method we'll remove the parameter completely, the copied method in the trait remains the same but gets the deprecated annotation.

However, that is a lot of work for the test kit. IMHO its not worth it.

svroonland: Of course, my bad.

svroonland (approved):

PR #1452, opened 2025-02-05
Preparing for producer diagnostics (3.x)

### LINKED ISSUE #1406: Producer diagnostics
With `ConsumerDiagnostics` users can listen to important events in the consumer. We would like to have something very similar for the producer.

A non-exhaustive list of events that should be supported:
* record is offered for sending (topic, partition, size of record)
* batch is offered for sending (number of records, total size of records)
* record was sent (topic, partition, size of record)

See also #1039.

erikvanoosten:
It has been a long wish to have producer diagnostics. Although these are not really hard to add, the consumer diagnostics are a bit in the way; these already claim some names and make code sharing awkward. Now that we are moving towards zio-kafka 3.0 we can fix these at the cost of some backward incompatibilities.

This PR has the same changes as #1507 (which will be applied to the series/2.x branch) but then with all deprecated code removed.

Partially fixes #1406.


erikvanoosten:
Hint for the reviewer: enable 'hide whitespace' to minimize the diff.

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
```diff
@@ -252,10 +259,10 @@ object Consumer {
     javaConsumer: JConsumer[Array[Byte], Array[Byte]],
     settings: ConsumerSettings,
     access: Semaphore,
-    diagnostics: Diagnostics = Diagnostics.NoOp
+    diagnostics: ConsumerDiagnostics = NoDiagnostics
```
vermas7988, line 262: If we are already doing a breaking change, I would prefer to use a boolean here in parameter e.g. `diagnosticsEnabled` to hide the details of type from user. It would be beginner friendly and also, It would prevent breaking changes even if the types are changed internally like now or some other reshuffling happens.

zio-kafka/src/main/scala/zio/kafka/diagnostics/internal/ConcurrentDiagnostics.scala
```diff
@@ -0,0 +1,25 @@
...
+import zio.stream._
+
+object ConcurrentDiagnostics {
+
+  /**
+   * @return
+   *   a `Diagnostics` that runs the wrapped `Diagnostics` concurrently in a separate fiber. Events are emitting to the
+   *   fiber via an unbounded queue. Upon finalization, `finalEvent` is emitted as the last event.
+   */
+  def make[DiagnosticEvent](
```
vermas7988, line 14: Personally I would have had a parent sealed trait for the `DiagnosticEvent` if we know that we only have `ConsumerDiagnosticEvent` and `ProducerDiagnosticEvent` implementaions. Or do we plan flexibility to have some other implementation from user? If we do that we can also drop explicit need of `finalEvent` as it can be decided by the type of event.

vermas7988 (commented):
I did a review. Just two suggestions if they are applicable.

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
vermas7988: or `enableDiagnostics`

erikvanoosten: Unfortunately, it doesn't work that way. There is no 'enable', as a user you have to implement the `Diagnostics` trait and in that way provide a callback.

zio-kafka/src/main/scala/zio/kafka/diagnostics/internal/ConcurrentDiagnostics.scala
erikvanoosten: Theoretically we might have more `DiagnosticEvent`s. For example for the admin client.

However, here the `DiagnosticEvent` is fully unspecified because `ConcurrentDiagnostics` does not need to make any assumptions about it. The only downside is that the type can not be inferred (that downside does not go away with your proposal).

zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala
vermas7988: okay understood..

zio-kafka/src/main/scala/zio/kafka/diagnostics/internal/ConcurrentDiagnostics.scala
vermas7988: just curious why we cant infer the type once we have sealed trait and then we start to have type bound on the param? But it makes sense that `ConcurrentDiagnostics` doesnt need to make assumptions about the Event type.

erikvanoosten: Sorry, I mixed it up with `SlidingDiagnostics` which is also not making any assumptions and therefore needs an annotation:
```scala
SlidingDiagnostics. make[ConsumerDiagnosticEvent](100)
```

svroonland:
I can see the use in generalizing SlidingDiagnostics and ConcurrentDiagnostics by parameterizing them with the type of diagnostic event. But I'd like to see if we can keep the rest of the stuff in the same package and with the same name, just to minimize changes.

What would be the downside of having `zio.kafka.consumer.diagnostics.DiagnosticEvent` and `zio.kafka.producer.diagnostics.DiagnosticEvent` for example?

erikvanoosten:
> What would be the downside of having `zio.kafka.consumer.diagnostics.DiagnosticEvent` and `zio.kafka.producer.diagnostics.DiagnosticEvent` for example?

One downside is that its very annoying when you have a scala file that needs both classes.
But your proposal is not unreasonable. I'll try it out.

erikvanoosten:
@svroonland the last commit reduces the number of changes one needs for migrating from zio-kafka 2.x to 3.x:

The following changes are needed:

| Old | New | Required? |
| --- | --- |
| `zio.kafka.consumer.diagnostics.Diagnostics` | `zio.kafka.consumer.Consumer.ConsumerDiagnostics` | optional |
| `zio.kafka.consumer.diagnostics.Diagnostics.NoOp` | `zio.kafka.consumer.Consumer.NoDiagnostics` | optional |
| `zio.kafka.consumer.diagnostics.DiagnosticEvent.Finalization.*` | `zio.kafka.consumer.ConsumerDiagnosticEvent.*` | required |

Different invocation:
old: `zio.kafka.consumer.diagnostics.Diagnostics.SligingQueue.make()`
new: `zio.kafka.diagnostics.SlidingDiagnostics.make[zio.kafka.consumer.diagnostics.DiagnosticEvent]()`

Move to internal or test package:

| Old | New | New location |
| --- | --- | --- |
| `zio.kafka.consumer.diagnostics.Diagnostics.ConcurrentDiagnostics` | `zio.kafka.diagnostics.internal.ConcurrentDiagnostics` | internal |
| `zio.kafka.consumer.diagnostics.DiagnosticEvent.Finalization` | removed | - |

Notes:

- The intermediate trait `DiagnosticEvent.Finalization` was removed from the type hierarchy because it doesn't serve any user purpose (it was introduced to make it easier to test the shutdown sequence of the consumer).

svroonland (approved):

zio-kafka/src/main/scala/zio/kafka/diagnostics/Diagnostics.scala
```diff
@@ -0,0 +1,21 @@
...
+trait Diagnostics[-DiagnosticEvent] {
+  def emit(event: => DiagnosticEvent): UIO[Unit]
+}
+
+object Diagnostics {
+
+  /**
+   * A diagnostics implementation that does nothing.
+   */
+  val noOp: Diagnostics[Any] = new Diagnostics[Any] {
```
erikvanoosten, line 17: Should be named `NoOp`.

PR #1449, opened 2025-02-02
Deprecate accessor methods

erikvanoosten:
Accessor methods are [deprecated by the ZIO community](https://zio.dev/reference/service-pattern/accessor-methods). Therefore, all accessor methods for `Consumer`, `Producer` and `TransactionalProducer` have been deprecated so that they can be removed in zio-kafka 3.0.0.

Accessor methods were heavily used in the zio-kafka unit tests. Therefore, all tests and benchmarks have been rewritten to use services directly. `KafkaTestUtils` has been extended with several methods to make this easier. To nudge users of the zio-kafka-test-kit, anything that promotes accessor methods is also deprecated.

The unit tests now have less layer trickery and are easier to understand.

The documentation has been extended with a migration guide: https://github.com/zio/zio-kafka/blob/accessors/docs/migrating-to-2.11.md

Hint for the reviewer: enable "hide whitespace" for smaller diffs.

PR #1455, opened 2025-02-06
Remove accessor methods from docs

erikvanoosten:
In preparation of removing the accessor method in zio-kafka 3.0.0, this change already removes them from the user documentation.

The diff looks a bit large because how to create a consumer was split to a different file.

File `writing-tests.md` is updated in a separate PR (#1449) as it involves code changes as well.


svroonland (approved):
LGTM

PR #1464, opened 2025-02-12
Fetch consumerGroupMetaData less often

### LINKED ISSUE #1431: Fetch consumerGroupMetaData less often
Currently, we fetch `consumerGroupMetaData` after every poll (see method `offerRecordsToStreams` in `Runloop`). This metadata only changes after a rebalance so fetching it that often is overhead.

Fix this by only fetching `consumerGroupMetaData` after a rebalance.

vermas7988:
The consumer group metadata only changes after a rebalance, yet currently we retrieve it after every poll. This change only fetches the consumer group metadata after a rebalance.

Fixes #1431.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -30,7 +30,7 @@ private[consumer] final class Runloop private (
   rebalanceCoordinator: RebalanceCoordinator,
   consumerMetrics: ConsumerMetrics,
   committer: Committer,
-  consumerMetaDataRef: Ref[Option[ConsumerGroupMetadata]]
+  consumerMetadataRef: Ref[Option[ConsumerGroupMetadata]]
 ) {
   private def newPartitionStream(tp: TopicPartition): UIO[PartitionStreamControl] =
     PartitionStreamControl.newPartitionStream(
@@ -129,7 +129,7 @@ private[consumer] final class Runloop private (
     if (streams.isEmpty) ZIO.succeed(fulfillResult)
     else {
       for {
-        consumerGroupMetadata <- consumerMetaDataRef.get.flatMap {
+        consumerGroupMetadata <- consumerMetadataRef.get.flatMap {
                                    case None     => getConsumerGroupMetadataIfAny
                                    case metadata => ZIO.succeed(metadata)
                                  }
@@ -262,7 +262,7 @@ private[consumer] final class Runloop private (
                                 ignoreRecordsForTps <- doSeekForNewPartitions(c, assignedTps)
 
                                 // invalidate current consumer group metadata
-                                _ <- consumerMetaDataRef.set(None)
+                                _ <- consumerMetadataRef.set(None)
 
                                 // The topic partitions that need a new stream are:
                                 //  1. Those that are freshly assigned
@@ -595,7 +595,7 @@ object Runloop {
       lastRebalanceEvent <- Ref.Synchronized.make[RebalanceEvent](RebalanceEvent.None)
       initialState = State.initial
       currentStateRef     <- Ref.make(initialState)
-      consumerMetaDataRef <- Ref.make[Option[ConsumerGroupMetadata]](None)
+      consumerMetadataRef <- Ref.make[Option[ConsumerGroupMetadata]](None)
       sameThreadRuntime   <- ZIO.runtime[Any].provideLayer(SameThreadRuntimeLayer)
       executor            <- ZIO.executor
       metrics = new ZioConsumerMetrics(settings.metricLabels)
@@ -627,7 +627,7 @@ object Runloop {
                   consumerMetrics = metrics,
                   rebalanceCoordinator = rebalanceCoordinator,
                   committer = committer,
-                  consumerMetaDataRef = consumerMetaDataRef
+                  consumerMetadataRef = consumerMetadataRef
                 )
       _ <- ZIO.logDebug("Starting Runloop")
 
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -205,7 +205,7 @@ private[consumer] final class Runloop private (
           settings.authErrorRetrySchedule
       )
 
-  private def handlePoll(state: State): Task[State] = {
+  private def handlePoll(state: State, initialConsumerGroupMetadata: Option[ConsumerGroupMetadata]): Task[State] = {
     for {
       partitionsToFetch  <- settings.fetchStrategy.selectPartitionsToFetch(state.assignedStreams)
       pendingCommitCount <- committer.pendingCommitCount
@@ -214,8 +214,7 @@ private[consumer] final class Runloop private (
                s" ${pendingCommitCount} pending commits," +
                s" resuming $partitionsToFetch partitions"
            )
-      _                            <- currentStateRef.set(state)
-      initialConsumerGroupMetadata <- consumer.withConsumerZIO(c => getConsumerGroupMetadataIfAny(c))
+      _ <- currentStateRef.set(state)
       pollResult <-
         consumer.runloopAccess { c =>
           for {
@@ -503,10 +502,11 @@ private[consumer] final class Runloop private (
           _ <- ZIO.logDebug(s"Processing ${commands.size} commands: ${commands.mkString(",")}")
           _ <- consumer.runloopAccess(committer.processQueuedCommits(_))
           streamCommands = commands.collect { case cmd: RunloopCommand.StreamCommand => cmd }
-          stateAfterCommands <- ZIO.foldLeft(streamCommands)(state)(handleCommand)
+          stateAfterCommands           <- ZIO.foldLeft(streamCommands)(state)(handleCommand)
+          initialConsumerGroupMetadata <- consumer.withConsumerZIO(c => getConsumerGroupMetadataIfAny(c))
 
           updatedStateAfterPoll <- shouldPoll(stateAfterCommands).flatMap {
-                                     case true  => handlePoll(stateAfterCommands)
+                                     case true  => handlePoll(stateAfterCommands, initialConsumerGroupMetadata)
                                      case false => ZIO.succeed(stateAfterCommands)
                                    }
           // Immediately poll again, after processing all new queued commands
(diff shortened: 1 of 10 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -154,7 +154,8 @@ private[consumer] final class Runloop private (
   }
 
   private val getConsumerGroupMetadataIfAny: UIO[Option[ConsumerGroupMetadata]] =
-    if (settings.hasGroupId) consumer.runloopAccess(c => ZIO.attempt(c.groupMetadata())).fold(_ => None, Some(_))
+    if (settings.hasGroupId)
+      consumer.rebalanceListenerAccess(c => ZIO.attempt(c.groupMetadata())).fold(_ => None, Some(_))
```
erikvanoosten: There be dragons.

`rebalanceListenerAccess` is for using the consumer _from within_ the rebalance listener. `getConsumerGroupMetadataIfAny` is not invoked from within the rebalance listener.

The distinction is quite important. Within the rebalance listener we can assume that we already have access to the consumer. Outside of it, we first need to request access. This is what `consumer.runloopAccess` is for.

erikvanoosten (commented):
Thanks for the PR!
Unfortunately, it does not solve #1431 at all.

Was this PR generated with AI?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
vermas7988: Not really with AI but I have limited to 0 knowledge of the project. I will try a but more if I can improve the PR.

erikvanoosten: The trick is not to change this piece of code, the trick is to make sure it is only invoked when necessary, that is, when a rebalance happened.

This is the place where you know a rebalance happened: https://github.com/zio/zio-kafka/blob/master/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala#L251

Feel free to ask questions! For direct chatting you can also find us on Discord https://discord.com/channels/629491597070827530/629497941719121960

vermas7988: In my understanding we would need some initial values of group metadata for the commits happening before rebalance. I reached you out on discord.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -128,7 +129,10 @@ private[consumer] final class Runloop private (
     if (streams.isEmpty) ZIO.succeed(fulfillResult)
     else {
       for {
-        consumerGroupMetadata <- getConsumerGroupMetadataIfAny
+        consumerGroupMetadata <- consumerMetadataRef.get.flatMap {
+                                   case None     => getConsumerGroupMetadataIfAny
+                                   case metadata => ZIO.succeed(metadata)
+                                 }
```
erikvanoosten, line 135: Looks good, but its not complete yet.

Could you:
1. move this code into `getConsumerGroupMetadataIfAny`
2. make sure that the `consumerMetadataRef` is set with the retrieved value

vermas7988: I pushed the suggested changes..

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -257,6 +271,9 @@ private[consumer] final class Runloop private (
                               for {
                                 ignoreRecordsForTps <- doSeekForNewPartitions(c, assignedTps)
 
+                                // invalidate current consumer group metadata
+                                _ <- consumerMetadataRef.set(None)
```
svroonland, line 275: Can we immediately set the consumer group data here, instead of 'on-demand'?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -29,7 +29,8 @@ private[consumer] final class Runloop private (
   currentStateRef: Ref[State],
   rebalanceCoordinator: RebalanceCoordinator,
   consumerMetrics: ConsumerMetrics,
-  committer: Committer
+  committer: Committer,
+  consumerMetadataRef: Ref[Option[ConsumerGroupMetadata]]
```
erikvanoosten, line 33: If we're going to shorten the name, I'd prefer `groupMetadataRef`. This code is already in a 'consumer' context while this metadata is really only about groups.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -153,9 +154,22 @@ private[consumer] final class Runloop private (
...
-    if (settings.hasGroupId) consumer.runloopAccess(c => ZIO.attempt(c.groupMetadata())).fold(_ => None, Some(_))
-    else ZIO.none
+  /**
+   * @return
+   *   optionally consumer group metadata, first we try get it from consumerMetadataRef, or fetch it from consumer if
+   *   not present. If the group id is not set, we return None.
+   */
+  private def getConsumerGroupMetadataIfAny: UIO[Option[ConsumerGroupMetadata]] =
+    consumerMetadataRef.get.flatMap {
+      case None if settings.hasGroupId =>
```
erikvanoosten, line 164: Since `settings.hasGroupId` is the simpler test (no memory barriers required), IMHO it is better to move the if statement up one level. WDYT?

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -153,9 +154,22 @@ private[consumer] final class Runloop private (
...
 
-  private val getConsumerGroupMetadataIfAny: UIO[Option[ConsumerGroupMetadata]] =
-    if (settings.hasGroupId) consumer.runloopAccess(c => ZIO.attempt(c.groupMetadata())).fold(_ => None, Some(_))
-    else ZIO.none
+  /**
+   * @return
+   *   optionally consumer group metadata, first we try get it from consumerMetadataRef, or fetch it from consumer if
+   *   not present. If the group id is not set, we return None.
+   */
+  private def getConsumerGroupMetadataIfAny: UIO[Option[ConsumerGroupMetadata]] =
```
erikvanoosten, line 162: There is no need to make this a `def`, it can stay a `val`. ZIO workflows are values!

```suggestion
  private val getConsumerGroupMetadataIfAny: UIO[Option[ConsumerGroupMetadata]] =
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -153,9 +154,22 @@ private[consumer] final class Runloop private (
...
+    consumerMetadataRef.get.flatMap {
+      case None if settings.hasGroupId =>
+        consumer
+          .runloopAccess(c => ZIO.attempt(c.groupMetadata()))
+          .fold(_ => None, Some(_))
+          .tap(metadata => consumerMetadataRef.set(metadata))
+
+      // If consumer group metadata is already present, or if the group id is not set, we don't need to re-fetch it.
```
erikvanoosten, line 170: For me this comment (line 170) is not necessary, the method's scaladoc already explains that we're using the ref for caching.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten, line 275: I would suggest we move this above `ignoreRecordsForTps <- doSeekForNewPartitions(c, assignedTps)`. `doSeekForNewPartitions` doesn't actually use the groupmetadata, but it feels safer. WDYT?

erikvanoosten (changes requested):
Looks great, we're almost there!

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: 🤔  We could, but there is no guarantee that the group metadata is needed (it is likely though). On the other hand it would remove a bit of complexity.

There is another aspect: method `getConsumerGroupMetadataIfAny` is written such that if the call fails, it returns `None`. That means that the call is retried every time. We lose the retries when the logic is moved to this place. I am still doubting between whether this is good or bad 🤷

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
vermas7988: done

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
vermas7988: done

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
vermas7988: okay, it sounds a bit confusing. Because, as a beginner, I would guess vals would not be computed again each time. But I made changes to be consistent everywhere

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
vermas7988: as its not used in above statement, it feels safe. But, I agree invalidating metadata as first step feels like more natural.

vermas7988: I can do change, but I have limited understanding for the second aspect Erik shared. If we fail to set `getConsumerGroupMetadataIfAny`, would we retry the current poll and if so, would it be still marked as rebalance event?
If above happens as I said, it seems like a good change.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: That is correct. The ZIO value, a workflow, is only constructed once, and then executed as many times as needed.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: We will have to look in the git history to see why any error is being ignored from the call to `consumer.groupMetadata()`. I will have time for that on Wednesday but feel free to do this earlier. To be honest I don't expect to find an answer but we can try. Perhaps @iravid remembers?

On the other hand, as the PR is right now, it already solves issue #1431. So I am in favor of merging this now and create a new issue to explore the idea of moving the logic. @svroonland WDYT?

erikvanoosten (approved):
Thanks @vermas7988! You made a significant change in a very complex codebase 🙂

PR #1467, opened 2025-02-15
Explore event sourced rebalance listener

erikvanoosten:
In this experiment we remove even more logic from the rebalance listener; it now reports which callbacks happened and no longer collapses that information. This gives the runloop a full picture of what happened during the rebalance.

The logic stays exactly the same, it just moved to the runloop. This does however allow more complex logic if it becomes necessary.

This is a draft because I am not sure yet if this PR is an improvement.


svroonland (commented):
I like it on the one hand but since RebalanceCoordinator is now a bit mixed between simply storing the callback events and actually ending streams, I'm not sure it's really an improvement

erikvanoosten:
Replaced by #1472.

PR #1471, opened 2025-02-18
Restructure consumer group polling runloop (fixes #1470 )

### LINKED ISSUE #1470: Fetch consumer group metadate after a rebalance
As discussed in https://github.com/zio/zio-kafka/pull/1464#discussion_r1957270397 we can simplify the code that fetches consumer group metadata by fetching it immediately after a rebalance. However, we first need to explore why errors from the call to `consumer.groupMetadata()` are suppressed, and whether it makes sense to keep retrying after a failure.

This issue is done when:
* we know if we can ignore errors from `consumer.groupMetadata()`
* we know retrying `consumer.groupMetadata()` is useless
* if the answer is yes to both, fetching the group metadata was moved to directly after the rebalance

vermas7988:
Restructure consumer group polling runloop (fixes #1470  )

### DIFF
--- a/zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
+++ b/zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
@@ -965,7 +965,7 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
          *     commits (it does not actually commit anywhere). In addition we set `maxRebalanceDuration` to 20 seconds.
          *   - After a few messages consumer 2 is started and a rebalance starts.
          *   - We measure how long the rebalance takes.
-         *
+         *
          * When the rebalance finishes immediately, we know that the external commits were used. If it finishes in 20
          * seconds, we know that the external commits were not used.
          */
--- a/zio-kafka/src/main/scala/zio/kafka/admin/AdminClient.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/admin/AdminClient.scala
@@ -301,7 +301,7 @@ trait AdminClient {
 
   /*
    * Lists access control lists (ACLs) according to the supplied filter.
-   *
+   *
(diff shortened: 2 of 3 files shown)


zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
vermas7988, line 1: Differs in linespace, due to fmt.

zio-kafka/src/main/scala/zio/kafka/admin/AdminClient.scala
vermas7988, line 1: Differs in linespace, due to fmt.

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
erikvanoosten: Please run `sbt fmt`.

erikvanoosten (changes requested):
The idea posted in #1470 is to fetch the group metadata after a rebalance. The right place to do this is line 274. Currently, we clear the ref on that line. Instead, we want to fill the ref immediately with the value obtained by calling `getConsumerGroupMetadataIfAny`. Other places in `Runloop` can then read the ref directly and no longer need to call `getConsumerGroupMetadataIfAny`.

vermas7988:
> The idea posted in #1470 is to fetch the group metadata after a rebalance. The right place to do this is line 274. Currently, we clear the ref on that line. Instead, we want to fill the ref immediately with the value obtained by calling `getConsumerGroupMetadataIfAny`. Other places in `Runloop` can then read the ref directly and no longer need to call `getConsumerGroupMetadataIfAny`.

I raised another PR, will close this one. But probably we would need to initialise the group metadata ref before first rebalance, there is a test failing. Investigating more: https://github.com/zio/zio-kafka/pull/1474

PR #1484, opened 2025-02-22
Make benchmarks more accurate

erikvanoosten:
... by adding a 5ms delay in each call to the underlying java consumer. This solves the problem that the embedded kafka broker can respond virtually immediately, while in reality there is almost always a network hop.

Since the calls of the underlying java consumer do not correspond 1:1 to network traffic, this is a crude approximation. Nevertheless, we expect that redundant consumer calls will become visible in the benchmarks.

Also: remove a duplicate benchmark.

### DIFF
--- a/zio-kafka-bench/README.md
+++ b/zio-kafka-bench/README.md
@@ -23,6 +23,17 @@ the worst possible case for zio-kafka. This is because these consumers only coun
 processing. This makes the comparison look bad for zio-kafka because zio-kafka programs normally process records in
 parallel, while other Kafka consumers process records serially.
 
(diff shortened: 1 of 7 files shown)


svroonland (approved):
If I'm understanding it right, this is applied to all benchmarks so doesn't give zio-kafka an unfair advantage, just a more realistic comparison, right?

erikvanoosten:
> If I'm understanding it right, this is applied to all benchmarks so doesn't give zio-kafka an unfair advantage, just a more realistic comparison, right?

Yes, that is correct.

svroonland (approved):

svroonland (approved):

PR #1485, opened 2025-02-23
Fix rebalance coordinator spec

erikvanoosten:
Previously, some of the rebalanceSafeCommits tests passed regardless of whether rebalanceSafeCommits was enabled or not. This is now fixed.

Also:
- added test outlines
- made assertions more precise
- removed unnecessary code
- add a test for when rebalanceSafeCommits is disabled

### DIFF
--- a/zio-kafka-test/src/test/scala/zio/kafka/consumer/internal/RebalanceCoordinatorSpec.scala
+++ b/zio-kafka-test/src/test/scala/zio/kafka/consumer/internal/RebalanceCoordinatorSpec.scala
@@ -12,6 +12,9 @@ import zio.kafka.consumer.{ CommittableRecord, ConsumerSettings }
 import zio.test._
 import zio._
 
+import java.util.{ Map => JavaMap }
+import scala.jdk.CollectionConverters._
+
 object RebalanceCoordinatorSpec extends ZIOSpecDefaultSlf4j {
   type BinaryMockConsumer = MockConsumer[Array[Byte], Array[Byte]]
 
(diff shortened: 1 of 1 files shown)


zio-kafka-test/src/test/scala/zio/kafka/consumer/internal/RebalanceCoordinatorSpec.scala
```diff
@@ -39,165 +36,197 @@ object RebalanceCoordinatorSpec extends ZIOSpecDefaultSlf4j {
...
+            coordinator <- makeCoordinator(
+                             lastEvent,
+                             consumer,
+                             rebalanceSafeCommits = true,
+                             assignedStreams = Chunk(streamControl),
+                             committer = committer,
+                             maxRebalanceDuration = 10.seconds
+                           )
+            revokeDuration <- coordinator.toRebalanceListener.onRevoked(Set(tp)).timed.map(_._1)
+          } yield assertTrue(revokeDuration.toMillis > 150, revokeDuration.toMillis < 6000)
```
svroonland, line 134: Not a huge fan of timing conditions like this as they are often flaky. Any way to keep using the promise for signalling stuff and expressing the condition in a different way..?

erikvanoosten: I see your point and yet and I don't see how it could be done differently.

These tests are written such that at worst they can cause false positives. False negatives are either not possible or very unlikely.

In test `should wait for the last pulled offset to commit` we want to validate the that the rebalance is extended till all three offsets are committed.
Failure scenarios:
- Rebalance doesn't wait for commit
  -  validated by asserting minimum duration. This can cause false positives, but not false negatives: if it it takes less time it _has to be_ because it didn't wait
  - alternative: count commits. When commit count after rebalance is not number of expected records the rebalance ended too quickly. This is worse: we now we have a race condition on the commit count which can also cause false positives.
- Rebalance waits too long (no commits handled at all)
  - validated by asserting maximum duration. The max rebalance duration is quite long: 10s, the assertion is using max 6s. You need a _very_ slow machine to cause false negatives.
  - alternative: no idea

In test `should continue if waiting for the stream to continue has timed out` we want to validate the the rebalance did its best in waiting for commits that never happened.
Failure scenarios:
- Rebalance ends before max rebalance duration
  - validated by checking actual rebalance duration. This can cause false positives, but not false negatives: if it it did not wait long enough, it _has to be_ because it didn't wait
  - alternative: no idea

In test `should not wait for stream to commit` we want to validate the the rebalance doesn't wait.
Failure scenarios:
- Rebalance does wait for commits:
  - validated by checking that the rebalance took shorter than what it would have taken when it awaited the commits. The rebalance could be slow for other reasons, so this can indeed cause false negatives. However, with 600ms there is so much time that a false negative is unlikely. (False positives are not possible.)
  - alternative: no idea. We can't count commits because we don't do commits. Doing commits increases the complexity of the test because then it needs to address the uncertainty of whether the rebalance ended because of commits, or because its was not waiting.

svroonland: See alternative in #1498

erikvanoosten: We can combine the approaches though. I'll try it out.

erikvanoosten: Added the additional assertion in https://github.com/zio/zio-kafka/pull/1485/commits/6dbf077b4f97da0cb643a28a3fb1dfbe186dd707

In addition, we now ensure that the records were pulled by the stream in https://github.com/zio/zio-kafka/pull/1485/commits/203b77d3ef75b179c60abef9f36a2e65dfb8bed0, before the revoke could theoretically happen before the stream pulled the records,this would completely invalidate the test.

The remaining time based assertions are still needed because they validate the time based behavior of the rebalance listener.

svroonland (approved):

PR #1492, opened 2025-02-27
Check consumer settings for transactional producer

### LINKED ISSUE #1483: Add consumer settings checks for transactional producer
When a consumer is coupled to a transactional producer, the consumer must have rebalanceSafeCommits enabled.

This issue is to:
- make the consumer settings field public in the consumer
- validate the settings while creating a transactional producer

vermas7988:
The transactional producer requires rebalance-safe-commits from the consumer it is connected to. When rebalance-safe-commits is disabled, this now results in an error.

Fixes #1483.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/producer/TransactionalProducer.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/producer/TransactionalProducer.scala
@@ -23,7 +23,7 @@ object TransactionalProducer {
   case object UserInitiatedAbort
   final case class TransactionLeaked(offsetBatch: OffsetBatch) extends RuntimeException
   case object RebalanceSafeConsumerRequired
-      extends RuntimeException(
+      extends IllegalArgumentException(
         "Consumer settings must have rebalanceSafeCommits set to true to use transactional producer."
       )
 
@@ -92,7 +92,7 @@ object TransactionalProducer {
 
   def make(settings: TransactionalProducerSettings, consumer: Consumer): ZIO[Scope, Throwable, TransactionalProducer] =
     for {
-      _ <- ZIO.cond(consumer.consumerSettings.rebalanceSafeCommits, (), RebalanceSafeConsumerRequired)
+      _ <- ZIO.fail(RebalanceSafeConsumerRequired).unless(consumer.consumerSettings.rebalanceSafeCommits)
       rawProducer <- ZIO.acquireRelease(
                        ZIO.attempt(
                          new KafkaProducer[Array[Byte], Array[Byte]](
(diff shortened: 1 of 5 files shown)


zio-kafka/src/main/scala/zio/kafka/producer/TransactionalProducer.scala
```diff
@@ -22,6 +22,10 @@ trait TransactionalProducer {
 object TransactionalProducer {
   case object UserInitiatedAbort
   final case class TransactionLeaked(offsetBatch: OffsetBatch) extends RuntimeException
+  case object RebalanceSafeConsumerRequired
+      extends RuntimeException(
```
vermas7988, line 26: Should this be an `IllegalArgumentException` instead?

zio-kafka/src/main/scala/zio/kafka/producer/TransactionalProducer.scala
```diff
@@ -88,6 +92,7 @@ object TransactionalProducer {
 
   def make(settings: TransactionalProducerSettings, consumer: Consumer): ZIO[Scope, Throwable, TransactionalProducer] =
     for {
+      _ <- ZIO.cond(consumer.consumerSettings.rebalanceSafeCommits, (), RebalanceSafeConsumerRequired)
```
svroonland, line 95: Personal preference, but I think `ZIO.fail(RebalanceSafeConsumerRequired).unless(consumer.consumerSettings.rebalanceSafeCommits)` conveys the intention to fail somewhat more.

zio-kafka/src/main/scala/zio/kafka/producer/TransactionalProducer.scala
svroonland: I think that's better yes

svroonland (commented):
Awesome! Two nitpicks

zio-kafka/src/main/scala/zio/kafka/producer/TransactionalProducer.scala
vermas7988: done

zio-kafka/src/main/scala/zio/kafka/producer/TransactionalProducer.scala
vermas7988: makes sense, updated!

zio-kafka/src/main/scala/zio/kafka/producer/TransactionalProducer.scala
```diff
@@ -22,6 +22,10 @@ trait TransactionalProducer {
 object TransactionalProducer {
   case object UserInitiatedAbort
   final case class TransactionLeaked(offsetBatch: OffsetBatch) extends RuntimeException
+  case object RebalanceSafeConsumerRequired
```
erikvanoosten, line 25: This name suggest the consumer is safe in rebalances. This is unfortunately not the case, not can it ever be.
```suggestion
  case object RebalanceSafeCommitsRequired
```

erikvanoosten (changes requested):
Nice!
Just one more small thing.

zio-kafka/src/main/scala/zio/kafka/producer/TransactionalProducer.scala
vermas7988: done

erikvanoosten (approved):

PR #1494, opened 2025-03-01
More tests for rebalanceSafeCommit mode and Small stuff

erikvanoosten:
More tests for rebalanceSafeCommit mode.

Also:
- Improve structure of zio-kafka 3 migration docs.
- Remove redundant integration test. The behavior that should be tested is already covered in a unit test in `CommitterSpec`. The removed integration test probably doesn't test the wanted behavior anyway.
- Use `forkScoped` for endless producing in a unit test instead of a `Ref`.
- Use ConsumerSettings helper method where possible.
- Small spelling fixes.

### DIFF
--- a/docs/migrating-to-zio-kafka-3.md
+++ b/docs/migrating-to-zio-kafka-3.md
@@ -161,7 +161,11 @@ expect only a few very experienced users to make use of it. Therefore, only the 
 
 # 3. `restartStreamOnRebalancing` mode
 
-This mode will no longer be available in zio-kafka 3. Contact us on [Discord](https://discord.com/channels/629491597070827530/629497941719121960) for alternatives.
+This mode is longer be available in zio-kafka 3. With `restartStreamOnRebalancing` all streams are ended during a
+rebalance, even when the partition for that stream was not revoked. One of its purposes was to enable transactional
+consuming. Since zio-kafka 3 however, transactional consuming no longer needs this mode.
+
+If your goal is to [prevent duplicates](preventing-duplicates.md), please use `rebalanceSafeCommits` instead.
 
 # Other changes?
 
--- a/docs/migrating-to-zio-kafka-3.md
+++ b/docs/migrating-to-zio-kafka-3.md
@@ -3,16 +3,26 @@ id: migrating-to-zio-kafka-3
 title: "Migrating to zio-kafka 3"
 ---
 
+Zio-kafka 3.0.0 brings a number of backwards incompatible changes:
+
+1. Removal of all deprecated methods, _including accessor methods,_
+2. The transactional producer is now much easier to use, 
+3. `restartStreamOnRebalancing` mode is no longer supported.
+
+This document helps you migrate from zio-kafka 2 to zio-kafka 3.
+
+# 1. Deprecations
+
 Zio-kafka 3.0.0 removes everything that was deprecated in the zio-kafka 2.x series. In particular, this includes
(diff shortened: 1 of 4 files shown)


docs/migrating-to-zio-kafka-3.md
```diff
@@ -143,7 +153,13 @@ for {
...
-# `restartStreamOnRebalancing` mode
+# 2. Changes in the transactional producer
+
+The transactional producer is now much easier to use. The zio-kafka 2.x transactional API is so complicated that we
+expect only a few very experienced users to make use of it. Therefore, only the new API is described in
+[transactions](transactions.md).
+
+# 3. `restartStreamOnRebalancing` mode
 
 This mode will no longer be available in zio-kafka 3. Contact us on [Discord](https://discord.com/channels/629491597070827530/629497941719121960) for alternatives.
```
svroonland, line 164: Sounds like we have a few laying in the back of the shop we're willing to let go at a reasonable price 😅

erikvanoosten: 😄 
Do you want to change it?

zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
```diff
@@ -624,6 +642,62 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
...
+                         ) *>
+                         partitionStream.tap(_.offset.commit).runDrain
+                     )
+                     .as(tp)
+                 }
+                 .take(partitionCount.toLong / 2)
+                 .runDrain
+                 .forkScoped
+          _ <- consumer1Ready.await
+          _ <- consumer2
```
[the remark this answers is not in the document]
erikvanoosten: Awaiting the ref makes sure there is a rebalance that both consumers participate in. I see now that it would be better to continuously produce message. On a fast machine the first consumer will have consumed all the messages before the second consumer can start!

BTW, this test is modeled after the test above it.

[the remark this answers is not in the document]
erikvanoosten: Fixed.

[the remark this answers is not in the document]
erikvanoosten: Its also awaiting fiber completion now.

docs/migrating-to-zio-kafka-3.md
```diff
@@ -3,16 +3,26 @@ id: migrating-to-zio-kafka-3
 title: "Migrating to zio-kafka 3"
 ---
 
+Zio-kafka 3.0.0 brings a number of backwards incompatible changes:
+
+1. Removal of all deprecated methods, _including accessor methods,_
+2. The transactional producer is now much easier to use, 
+3. `restartStreamOnRebalancing` mode is no longer supported.
```
svroonland, line 10: .. is now replaced by `rebalanceSafeCommits`

svroonland (approved):

docs/migrating-to-zio-kafka-3.md
erikvanoosten: I never thought of `restartStreamOnRebalancing` and `rebalanceSafeCommits` as being compatible in any way. But you're right in that they have overlapping goals. I'll add some text about that.

erikvanoosten: Fixed.

PR #1501, opened 2025-03-15
Subscription stream control - alternative API

svroonland:
See #1201


svroonland:
I have not implemented pause and resume yet, but I think it should be possible.

erikvanoosten:
This is good stuff. I agree, this is much better than the interruption based approach!

PR #1507, opened 2025-03-23
Preparing for producer diagnostics (2.x)

### LINKED ISSUE #1406: Producer diagnostics

erikvanoosten:
It has been a long wish to have producer diagnostics. Although these are not really hard to add, the consumer diagnostics are in the way; these already claim some names and make code sharing awkward. Now that we are moving towards zio-kafka 3.0 we can fix this at the cost of some backward incompatibilities.

This PR prepares for the introduction of producer diagnostics by moving the sharable diagnostics code into a new higher level `diagnostics` package. In addition, zio-kafka internal code is moved to the `diagnostics.internal` package.

This also removes the trait `DiagnosticEvent.Finalization` from the type hierarchy because it doesn't serve any user purpose (it was introduced to make it easier to test the shutdown sequence of the consumer).

All old APIs are still available but are deprecated.

This change is source compatible, but _not_ binary compatible and therefore (following zio-kafka conventions) requires a new minor release.

Partially fixes #1406.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/diagnostics/Diagnostics.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/diagnostics/Diagnostics.scala
@@ -2,21 +2,22 @@ package zio.kafka.consumer.diagnostics
 
 import zio.stream.ZStream
 import zio._
-import zio.kafka.consumer.diagnostics.DiagnosticEvent.Finalization.ConsumerFinalized
+import zio.kafka.consumer.Consumer
+import zio.kafka.consumer.diagnostics.DiagnosticEvent.ConsumerFinalized
 
-trait Diagnostics {
-  def emit(event: => DiagnosticEvent): UIO[Unit]
-}
 object Diagnostics {
-  case object NoOp extends Diagnostics {
-    override def emit(event: => DiagnosticEvent): UIO[Unit] = ZIO.unit
-  }
 
-  final case class SlidingQueue private[Diagnostics] (queue: Queue[DiagnosticEvent]) extends Diagnostics {
+  /** A diagnostics implementation for [[DiagnosticEvent consumer DiagnosticsEvent]]s that does nothing. */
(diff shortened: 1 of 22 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/diagnostics/Diagnostics.scala
```diff
@@ -2,21 +2,22 @@ package zio.kafka.consumer.diagnostics
 
 import zio.stream.ZStream
 import zio._
-import zio.kafka.consumer.diagnostics.DiagnosticEvent.Finalization.ConsumerFinalized
+import zio.kafka.consumer.Consumer
+import zio.kafka.consumer.diagnostics.DiagnosticEvent.ConsumerFinalized
 
-trait Diagnostics {
```
svroonland, line 7: Should we keep this for source compatibility?

svroonland: Ah, moved to package

svroonland (commented):
As discussed, I wonder if a construct like `class MyDiagnostics extends zio.kafka.consumer.Consumer.Diagnostics { def emit(..) }` is still source compatible now.

erikvanoosten:
> As discussed, I wonder if a construct like `class MyDiagnostics extends zio.kafka.consumer.Consumer.Diagnostics { def emit(..) }` is still source compatible now.

Now checked via https://github.com/zio/zio-kafka/pull/1507/commits/8d89005e9429936a7fcada0953702a1394dd10ef

svroonland (approved):
Nice!

PR #1515, opened 2025-03-26
Provide diagnostics via `ConsumerSettings`

erikvanoosten:
Since most people do not use consumer diagnostics, it is a bit annoying that everybody has to provide a (dummy) diagnostics layer when the zio-kafka `Consumer` is constructed via a layer.

This is solved by moving the configuration of diagnostics to `ConsumerSettings`.

This approach also makes it consistent with the to-be-introduced `ProducerDiagnostics`.

### DIFF
--- a/docs/migrating-to-zio-kafka-3.md
+++ b/docs/migrating-to-zio-kafka-3.md
@@ -6,8 +6,9 @@ title: "Migrating to zio-kafka 3"
 Zio-kafka 3.0.0 brings a number of backwards incompatible changes:
 
 1. Removal of all deprecated methods, _including accessor methods,_
-2. The transactional producer is now much easier to use, 
-3. `restartStreamOnRebalancing` mode is no longer supported.
+2. The transactional producer is now much easier to use,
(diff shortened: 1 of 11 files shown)


docs/migrating-to-zio-kafka-3.md
```diff
@@ -159,7 +160,75 @@ The transactional producer is now much easier to use. The zio-kafka 2.x transact
 expect only a few very experienced users to make use of it. Therefore, only the new API is described in
 [transactions](transactions.md).
 
-# 3. `restartStreamOnRebalancing` mode
+# 3. Diagnostics set via consumer settings
+
+Diagnostics, an optional callback for key events in the consumer life-cycle, is now set via
+`ConsumerSettings.withDiagnostics`. In zio-kafka 2.x it was the second (optional) argument to `Consumer.make`, or an
```
svroonland, line 166: At some point we should remove the 2.0 docs, it's too much of maintenance burden

svroonland (approved):

PR #1516, opened 2025-03-30
Kafka 4.x (Integration tests for embedded-kafka 4.x - DO NOT MERGE)

gaeljw:
This adds commits on top of the ongoing work to support Kafka 4.x in https://github.com/zio/zio-kafka/pull/1504, just as a way to share changes. This PR is not intended to be merged.

This also serves as integration tests for embedded-kafka 4.x (https://github.com/embeddedkafka/embedded-kafka/pull/562)

---

Test status:

- `AdminSpec`: consistent failures in 4 tests, the errors makes it look like it's a timing issue, it's like the `listTopicsFiltered` gives obsolete data or doesn't wait for the previous action (create/delete topic) to actually be done.
  - `client admin test / create, describe, delete multiple topic`
  - `client admin test / create, list, delete single topic`
  - `client admin test / create, list, delete multiple topic`
  - `client admin test - create, describe topic config, delete multiple topic`
- `ConsumerSpec`: two tests failing after a 2mn timeout
  - `Consumer Streaming - does not process messages twice for transactional producer, even when rebalancing - org.apache.kafka.clients.consumer.RangeAssignor`
  - `Consumer Streaming - does not process messages twice for transactional producer, even when rebalancing - org.apache.kafka.clients.consumer.CooperativeStickyAssignor`
- `ProducerSpec`: a few tests failing, likely caused by the 1st error `Compacted topic cannot accept message without key in topic partition`. It's unclear to me why it would start failing now and I'm also unclear if this is deliberate in the tests to have no key or not.
- `AdminSaslSpec`: errors due to `org.apache.kafka.common.errors.ClusterAuthorizationException: Request Request(processor=0, connectionId=127.0.0.1:7001-127.0.0.1:56736-0-0, session=org.apache.kafka.network.Session@3d97f66e, listenerName=ListenerName(CONTROLLER), securityProtocol=PLAINTEXT, buffer=null, envelope=None) is not authorized.`. This one may be related to the setup of the tests and/or embedded-kafka. Unclear for now.

### DIFF
--- a/zio-kafka-testkit/src/main/scala/zio/kafka/testkit/Kafka.scala
+++ b/zio-kafka-testkit/src/main/scala/zio/kafka/testkit/Kafka.scala
@@ -29,7 +29,7 @@ object Kafka {
         Map(
           "group.min.session.timeout.ms"     -> "500",
           "group.initial.rebalance.delay.ms" -> "0",
-          "authorizer.class.name"            -> "kafka.security.authorizer.AclAuthorizer",
+          "authorizer.class.name"            -> "org.apache.kafka.metadata.authorizer.StandardAuthorizer",
           "super.users"                      -> "User:ANONYMOUS"
         ),
       customBrokerProps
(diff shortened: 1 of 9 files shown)


zio-kafka-testkit/src/main/scala/zio/kafka/testkit/Kafka.scala
```diff
@@ -47,15 +47,15 @@ object Kafka {
    */
   def saslEmbeddedWith(customBrokerProps: Ports => Map[String, String]): ZLayer[Any, Throwable, Kafka.Sasl] =
     embeddedWithBrokerProps(
-      ports =>
+      ports => // TODO: why not use constants?
```
erikvanoosten, line 50: > why not use constants?

Because with zio-test, tests run in parallel, often with their own embedded kafka instance.

gaeljw: I meant for the settings keys. For instance instead of `"listeners"`, you could use `org.apache.kafka.network.SocketServerConfigs.LISTENERS_CONFIG`.

erikvanoosten: Ah! I see what you mean.
🤷 
On the one hand many our bullish on not using magic values (mostly for good reasons), on the other hand these constants in the kafka library are less stable than the string values (they move to other classes now and than) and looking up the config docs requires the actual value and won't work with the constant name. (BTW, I didn't write this code.)

gaeljw: I can relate to the string values moving from one class to another and also the additional indirection to go to the documentation. That's fair :)

erikvanoosten:
Thanks @gaeljw! This was merged into #1504.

PR #1518, opened 2025-04-09
Add producer diagnostics

erikvanoosten:
Producer diagnostics allows the user to keep track of what happens in the producer. It gives information on where records are produced to, and the size of each record.

Benchmarks show no overhead when these new producer diagnostics are disabled. Enabling the new producer diagnostics should increase heap usage somewhat, but we did not see any effects in the benchmark.

Also: improve scaladoc of `ProducerSettings`

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/producer/ProducerEvent.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/producer/ProducerEvent.scala
@@ -1,0 +1,68 @@
+package zio.kafka.producer
+
+import zio._
+
+sealed trait ProducerEvent
+
+/**
+ * Diagnostics events for the [[Producer]].
+ *
+ * The producer always handles records in batches. Even single records results in batches of 1. The diagnostic events
+ * are therefore always about batches.
+ *
+ * For each batch the [[RecordsOffered]] event is emitted first, followed by the [[RecordsSent]] event. Since batches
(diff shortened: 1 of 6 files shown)


zio-kafka/src/main/scala/zio/kafka/producer/ProducerEvent.scala
```diff
@@ -0,0 +1,68 @@
...
+ */
+object ProducerEvent {
+
+  /**
+   * Represents key data about a record that was handled by the producer.
+   *
+   * @param size
+   *   the size of the record's value
+   */
+  final case class ProducedRecord(topic: String, partition: Int, size: Int)
```
svroonland, line 31: This could lead to a lot of memory churn, when producing many records, don't you think?

erikvanoosten: Yes, that is why these events are only generated when diagnostics are used.
Do you think it is too much?

erikvanoosten: We can remove this field, or we can add a performance warning in ProducerSettings.

erikvanoosten: Benchmarks with very small records (worse case) show no effect. This is not conclusive, but at least it shows that the performance drop is not large.

svroonland (approved):

PR #1527, opened 2025-05-03
Add comment highlighting motivation behind usage of lazy val during records publication

ytalashko:
https://github.com/zio/zio-kafka/pull/1437 introduced a clever way to retry records publication after specific errors (more details in the referred PR).
There is a place of `lazy val` usage in that implementation, which may raise questions of why we need it, from the folks unfamiliar with the implementation.
To add a bit more clarity, this PR adds a comment with the reasoning for using  `lazy val`.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
@@ -406,6 +406,8 @@ private[producer] final class ProducerLive(
       val totalRecordCount = records.size
       val diagEm           = diagnosticsEmitterMaker.makeDiagnosticsEmitter(records)
 
+      // Collects the send results over one or more retries.
+      // Lazy to prevent allocation when all sends succeed on the first attempt.
       lazy val finalResults: Array[Either[Throwable, RecordMetadata]] =
         Array.fill(totalRecordCount)(leftPublishOmitted)
 


zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -406,7 +406,7 @@ private[producer] final class ProducerLive(
       val totalRecordCount = records.size
       val diagEm           = diagnosticsEmitterMaker.makeDiagnosticsEmitter(records)
 
-      lazy val finalResults: Array[Either[Throwable, RecordMetadata]] =
+      val finalResults: Array[Either[Throwable, RecordMetadata]] =
```
ytalashko, line 409: Hey there,
Was familiarizing myself with changes from https://github.com/zio/zio-kafka/pull/1437, they look great btw, and saw one minor place that can be improved.
Basically, `finalResults` array accessed for each result to be inserted (in `storeResults` function), and each such access is checking were the array already initialized or not.
This is redundant and can omitted, especially taking into account that this flow is running on the producer callback running thread.
Also, it seem to be initialized anyway for all cases.

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -424,10 +424,10 @@ private[producer] final class ProducerLive(
         ): ZIO[Any, Nothing, Any] = {
           def retryFailedRecords(results: Chunk[Either[Throwable, RecordMetadata]]): UIO[Unit] = {
             // Note that if we get here, all Left's can be retried. Also, we know there is at least 1 Left.
-            val toRetry: Seq[(RuntimeFlags, ByteRecord)] =
+            val toRetry: Seq[(Int, ByteRecord)] =
               (recordIndices lazyZip records lazyZip results).flatMap {
-                case (_, _, Right(_))     => Seq.empty
-                case (i, record, Left(_)) => Seq((i, record))
+                case (_, _, Right(_))     => Iterable.empty
+                case (i, record, Left(_)) => Iterable.single((i, record))
```
ytalashko, line 430: Just a minor update to use simpler short lived collections

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
erikvanoosten: Hello @ytalashko, thanks for reviewing zio-kafka! And thanks for the compliment :)

You're saying that `finalResults` is always initialized. The intention is that it is only initialized when there is a failure. When all sends of the batch succeed, the expression `recordIndices.size == totalRecordCount` should be `true` and `finalResults` should not be used. How did you come to the conclusion that it is always initialized?

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
erikvanoosten: That is interesting.
However, when I peek in the scala collections library I see that`Seq.empty` directly returns `List.Nil`. `Iterable.empty` _also_ returns `Nil`. No benefit here.

Stepping through the debugger on `Seq(value)` results in a lot of array wrapping, and then iterating through it to add it to an empty list. `Iterable(value)` very quickly goes through the same path.
However, `Iterable.single(value)` is very simple. It instantiates a single class and that's it. That should be much faster indeed!

For comparison, I also looked at zio's `Chunk`. It provides very short execution paths for both cases (`Chunk.empty` and `Chunk.single`). It also has a very efficient concatenation operator. We only need to change the zip order to produce `Chunk`s:

```scala
            val toRetry: Chunk[(RuntimeFlags, ByteRecord)] =
              (records lazyZip results lazyZip recordIndices).flatMap {
                case (_, Right(_), _)     => Chunk.empty
                case (record, Left(_), i) => Chunk.single((i, record))
              }
```

I would guess this performs better than what's currently there and the suggested improvements (but probably, still not enough to notice in the benchmarks).

erikvanoosten: I also only now notice the code says `RuntimeFlags` there instead of `Int` 😞. Its not a problem since its an alias but it doesn't look nice.

erikvanoosten: Even better! By using `Chunk`, the `Chunk.from` on line 436 can go away! Nice!

erikvanoosten: Thanks for the ideas! Implemented this in https://github.com/zio/zio-kafka/pull/1528.

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
ytalashko: Yeah, you are right @erikvanoosten, I missed that 🤦 
Wdyt about putting this as a in-code comment, e.g.
> // Used only in case of publication error which resulted in applied retries

Will update this MR to add the comment, not sure if this is necessary, so feel free to accept or close MR

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
ytalashko: Sounds great, 👍 
Thanks for sharing so much details

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -406,6 +406,8 @@ private[producer] final class ProducerLive(
       val totalRecordCount = records.size
       val diagEm           = diagnosticsEmitterMaker.makeDiagnosticsEmitter(records)
 
+      // Lazy val used to omit array allocation in case of successful chunk publication.
+      //  The allocation happens only in case of publication error which resulted in applied retries
```
erikvanoosten, line 410:
```suggestion
            // Collects the send results over one or more retries.
            // Lazy because allocation happens only in case retries are needed.
```

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
erikvanoosten: No worries, it was fun to do this deep dive.
Since this is now covered by the other PR, can you revert this change please?

erikvanoosten (commented):
Its nice to have the comment about why laziness is needed. Please consider my suggestion there. Can you change the title and description of the PR as well please?

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
ytalashko: > Since this is now covered by the other PR, can you revert this change please?

Sure

zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
ytalashko: Thanks, for suggestions, this version sounds better.
I also think about `taken place.` instead of `needed`, e.g.  `Lazy because allocation happens only in case retries are taken place.`, wdyt?

erikvanoosten: `are taken place`, I think you mean `are taking place` or simply `are happening`.

Even better, instead of describing when allocation happens, we can describe when we can get away without allocation:

> Collects the send results over one or more retries.
> Lazy to prevent allocation when all sends succeed on the first attempt.

ytalashko: Great, taking this final version.
Thanks @erikvanoosten!

erikvanoosten (approved):
🚀

ytalashko:
> Can you change the title and description of the PR as well please?

Updated, let me know, if anything there should be improved

PR #1528, opened 2025-05-03
Micro optimization by using `Chunk`

erikvanoosten:
Small optimization in Prodcuer by using `Chunk` instead of `Seq`. `Chunk` has much more efficient constructors.

Inspired by @ytalashko's work in #1527.


ytalashko:
Nice, 👍

svroonland (approved):

PR #1529, opened 2025-05-04
Producer micro optimizations by using `Iterable` instead of `Chunk`

ytalashko:
Some nitpicking here:
Following https://github.com/zio/zio-kafka/pull/1527#discussion_r2072424219 I believe `Iterable` should be a better choice instead of `Chunk` for the temp collection.
The way this collection is used (inside `LazyZip.flatMap`) is by calling `.iterator` on it.
`Chunk.[empty|single].iterator` is comparatively way heavier than `Iterable.[empty|single].iterator` (sure we are talking about very-very small impact).

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
@@ -428,8 +428,8 @@ private[producer] final class ProducerLive(
             // Note that if we get here, all Left's can be retried. Also, we know there is at least 1 Left.
             val toRetry: Chunk[(Int, ByteRecord)] =
               (results lazyZip recordIndices lazyZip records).flatMap {
-                case (Right(_), _, _)     => Chunk.empty
-                case (Left(_), i, record) => Chunk.single((i, record))
+                case (Right(_), _, _)     => Iterable.empty
+                case (Left(_), i, record) => Iterable.single((i, record))
               }
             val (retryIndices, retryRecords) = toRetry.unzip
             ZIO.logInfo(


erikvanoosten:
Hello @ytalashko . Thanks for diving into this again 🙂 

You are right that lazyZip does not use chunk's concatenation mechanisms.

The advantage of the proposed change has to be offset against the more complex constructors needing `Chunk.from` later on in the method. The only way to find out is to create a micro-benchmark with JMH.

You inspired me to look into Chunk's zip methods. These use a `chunkIterator` which is very efficiently implemented for `Chunk.empty` and `Chunk.single` (even more efficient than `Iterable.empty/single`).

```scala
            val toRetry: Seq[(Int, ByteRecord)] =
              (results zip recordIndices zip records).flatMap {
                case (Right(_), _, _)     => Chunk.empty
                case (Left(_), i, record) => Chunk.single((i, record))
              }
```

ytalashko:
> The advantage of the proposed change has to be offset against the more complex constructors needing Chunk.from later on in the method. The only way to find out is to create a micro-benchmark with JMH.

If I'm not mistaken, not really, `Chunk` construction as a result of `LazyZip.flatMap` is not dependent on the type of `Iterable` produced by the function argument passed into `LazyZip.flatMap`. The only way result of function argument passed to `LazyZip.flatMap` is used, is converted to `.iterator` and that's it.
What affects resulting `Chunk` construction is the first collection used in the `lazyZip` chain.

> You inspired me to look into Chunk's zip methods. These use a chunkIterator which is very efficiently implemented for Chunk.empty and Chunk.single (even more efficient than Iterable.empty/single).

I believe `lazyZip` is a better solution here, as it is not forcing actual zipped collections construction, which is forced with `Chunk.zip`. So, disregard the `Chunk.zip` performance it is more than not zipping at all.

erikvanoosten:
> I believe `lazyZip` is a better solution here, as it is not forcing actual zipped collections construction

There have been some articles that show that forcing a collection is often faster than iterating through a view. Only when the collections become larger (if I remember correctly, something like more than 500 items) iterating wins. The exact tipping point is of course different per situation. In addition Chunk is really efficient as it is array based, its iterators are also really efficient

So really, we can't reason this through, only JMH micro benchmarks can tell which solution will perform better.

ytalashko:
> There have been some articles that show that forcing a collection is often faster than iterating through a view. Only when the collections become larger (if I remember correctly, something like more than 500 items) iterating wins. The exact tipping point is of course different per situation. In addition Chunk is really efficient as it is array based, its iterators are also really efficient
> 
> So really, we can't reason this through, only JMH micro benchmarks can tell which solution will perform better.

Yeah, I agree that it is speculative to compare `lazyZip` vs `Chunk.zip` theoretically like we do here.
At the same time, the overall execution is seem way heavier in theory exactly for our example (3 input collections, not 2), e.g.
`(col1 lazyZip col2 lazyZip col3).flatMap { ... }` does a single pass through each collection simultaneously to create a resulting one.
`(chunk1 zip chunk2 zip chunk3).flatMap { ... }` does 1 pass through chunk1 & chunk2 to create intermediate chunk12, then 1 pass through chunk12 & chunk3 to create intermediate chunk123 and then pass through chunk123 to create resulting chunk.

`Chunk.zip` option does 3 times more passes and 3 times more collections created.
Still `Chunk.zip` may work faster.

At the same time, comparing the options currently implemented in the `master` branch and in the PR, PR's option is definitely an improvement over what's currently in `master`.

erikvanoosten:
I like your perseverance 🙂 

I find the academic discussion nice, but once we need to actually do work (set up JMH etc.) I think we should balance that against the benefits. Perhaps the code becomes more readable and maintainable? Or perhaps the code is in the hot-path and lots of people will benefit?

Does the code become more readable? For me, no. It may even make it a tiny bit worse because you may pause for a second, wondering why `Iterable` is used while most of zio-kafka uses `Chunk` for everything. Anyways, there is no big benefit in this area.

Will a lot of people benefit? In this case, the code we are talking about is invoked only after an auth-error, and then only for people that enabled retries. Auth-errors are only encountered for very specific Kafka broker configurations. So this code is called for a tiny amount of users, a tiny amount of the time. So that is a clear 'no'.

All in all, I don't think this is worth pursuing further.

ytalashko:
Makes sense, sure, no more nitpicking 😄

PR #1543, opened 2025-06-03
Add implicit trace parameter

erikvanoosten:
Add an implicit trace parameter to all public and private methods that use ZIO, as recommended by the [guidlines for library author](https://zio.dev/guides/migrate/zio-2.x-migration-guide#guidelines-for-library-authors).

Warning: this change is not binary, nor source compatible.

Source compatibility is only broken when a trait of zio-kafka is implemented by the user. The following traits are affected:

- `Diagnostics`
- `FetchStrategy`
- `Serde` (`Deserializer` and `Serializer`)

### DIFF
--- a/zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
+++ b/zio-kafka-test/src/test/scala/zio/kafka/consumer/ConsumerSpec.scala
@@ -1443,14 +1443,15 @@ object ConsumerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
           _                       <- ZIO.logDebug("Starting consumer 1")
           rebalanceEndTimePromise <- Promise.make[Nothing, Instant]
           c1Diagnostics = new ConsumerDiagnostics {
-                            override def emit(event: => DiagnosticEvent): UIO[Unit] = event match {
-                              case r: DiagnosticEvent.Rebalance if r.assigned.size == 1 =>
-                                ZIO.logDebug(s"Rebalance finished: $r") *>
-                                  Clock.instant.flatMap(rebalanceEndTimePromise.succeed).unit
-                              case r: DiagnosticEvent.Rebalance =>
-                                ZIO.logDebug(s"Rebalance finished: $r")
-                              case _ => ZIO.unit
-                            }
+                            override def emit(event: => DiagnosticEvent)(implicit trace: Trace): UIO[Unit] =
+                              event match {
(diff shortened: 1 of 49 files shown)


hearnadam:
Can you import `import zio.stacktracer.TracingImplicits.disableAutoTrace` in all `src/main` files?

erikvanoosten:
> Can you import `import zio.stacktracer.TracingImplicits.disableAutoTrace` in all `src/main` files?

@hearnadam That sounds easy. But, what does it do? I could not find any documentation.

Another question: should we force users to have the implicit trace on traits they can implement?

erikvanoosten:
> Can you import `import zio.stacktracer.TracingImplicits.disableAutoTrace` in all `src/main` files?

How can we get a trace for default parameters? For example:

```scala
  def consumeWith[R: EnvironmentTag, R1: EnvironmentTag, K, V](
    settings: ConsumerSettings,
    subscription: Subscription,
    valueDeserializer: Deserializer[R, V],
    commitRetryPolicy: Schedule[Any, Any, Any] = Schedule.exponential(1.second) && Schedule.recurs(3)
  )(f: ConsumerRecord[K, V] => URIO[R1, Unit])(implicit trace: Trace): RIO[R & R1, Unit]
  
/zio-kafka/zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala:378:70
could not find implicit value for parameter trace: zio.Trace
    commitRetryPolicy: Schedule[Any, Any, Any] = Schedule.exponential(1.second) && Schedule.recurs(3)

/zio-kafka/zio-kafka/src/main/scala/zio/kafka/consumer/Consumer.scala:378:99
could not find implicit value for parameter trace: zio.Trace
    commitRetryPolicy: Schedule[Any, Any, Any] = Schedule.exponential(1.second) && Schedule.recurs(3)
```

erikvanoosten:
@hearnadam Okay, from all the compiler errors I now understand what `import zio.stacktracer.TracingImplicits.disableAutoTrace` does 😄 

Is there some way to get summon a `Trace` out of thin air? That would really help with the default parameters (I think) and cases like this where we initialize a global `val`:

```scala
private[consumer] object RunloopExecutor {
  private def newSingleThreadedExecutor(implicit trace: Trace): URIO[Scope, Executor] = ...

  val newInstance: URIO[Scope, Executor] = newSingleThreadedExecutor(... what to put here? ...)
}
```

erikvanoosten:
We understand now that this PR will hide all stack frames from zio-kafka. In our opinion this is not desirable. Therefore, this PR will be closed.

PR #1546, opened 2025-06-07
Simplify code, use uniform type names

erikvanoosten:
This collects a few small improvements that were made in the disbanded tracing PR:
- replace function `val`s with `def`
- use `A` instead of `T`
- small scaladoc fixes
- small type improvements

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopExecutor.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopExecutor.scala
@@ -9,7 +9,7 @@ private[consumer] object RunloopExecutor {
 
   private val counter: AtomicLong = new AtomicLong(0)
 
-  private val newSingleThreadedExecutor: URIO[Scope, Executor] =
+  private def newSingleThreadedExecutor: URIO[Scope, Executor] =
     ZIO.acquireRelease {
       ZIO.succeed {
         val javaExecutor =
--- a/zio-kafka/src/main/scala/zio/kafka/serde/Deserializer.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/serde/Deserializer.scala
@@ -8,78 +8,78 @@ import scala.jdk.CollectionConverters._
 import scala.util.{ Failure, Success, Try }
 
 /**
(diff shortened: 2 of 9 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopExecutor.scala
```diff
@@ -9,7 +9,7 @@ private[consumer] object RunloopExecutor {
 
   private val counter: AtomicLong = new AtomicLong(0)
 
-  private val newSingleThreadedExecutor: URIO[Scope, Executor] =
+  private def newSingleThreadedExecutor: URIO[Scope, Executor] =
```
svroonland, line 12: What's the benefit here? In other places you replaced val's that are functions, here it doesn't take any parameters.

zio-kafka/src/main/scala/zio/kafka/serde/Deserializer.scala
```diff
@@ -8,78 +8,78 @@ import scala.jdk.CollectionConverters._
...
-  def blocking: Deserializer[R, T] =
+  def blocking: Deserializer[R, A] =
     Deserializer((topic, headers, data) => ZIO.blocking(deserialize(topic, headers, data)))
 
   /**
-   * Create a deserializer for a type U based on the deserializer for type T and a mapping function
+   * Create a deserializer for a type U based on the deserializer for type A and a mapping function.
    */
-  def map[U](f: T => U): Deserializer[R, U] = Deserializer(deserialize(_, _, _).map(f))
+  def map[U](f: A => U): Deserializer[R, U] = Deserializer(deserialize(_, _, _).map(f))
```
svroonland, line 30: To be consistent with `ZIO#map`, we should also replace `U` with `B`

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopExecutor.scala
erikvanoosten: Benefit: slightly less memory usage (1 pointer) because we don't need to store the (once used) result of this code.

svroonland (approved):

PR #1554, opened 2025-07-18
Test improvements

erikvanoosten:
While trying to run the unit tests will Mill some interesting patterns were found. Mill runs more tests in parallel, and presumable uses some class loader tricks, because singletons were initialized multiple times. These changes prevent issues with that pattern.

- Avoid hang when using Mill to run tests.
- Use much safer free port detection.
- Run SubscriptionsSpec in parallel.


svroonland (approved):
LGTM

PR #1567, opened 2025-09-06
Make admin-client java converters internal

erikvanoosten:
The admin client re-defines java classes in their Scala equivalent. The scala ↔ java converters are part of these classes even though they should not be used, or even not be needed, by users of the admin client.

In this change the converter code is removed from the public API and moved to an internal package.

The converter code is still available by importing the internal package. However, please note that code in internal packages may change in any way in any zio-kafka release.
```
import zio.kafka.admin.internal.JavaConverters._
```


erikvanoosten:
Closing as discussed earlier.

PR #1576, opened 2025-09-17
Limit size of consumer batches to max.poll.records

erikvanoosten:
During shutdown, when rebalanceSafeCommits is enabled, we wait until the consumer has processed all records that were pulled from the partition queue. Before this change, the consumer would pull all available (pre-fetched) records. We now put an upper limit on the amount of records that can be pulled. This decreases the time we need to wait, while it doesn't hurt throughput too much.

**Note:** this does mean we zio-kafka can no longer guarantee that all records (for a given partition) fetched together from the broker are grouped into one chunk.

This change also extracts collecting the runloop configuration to a separate class. This has the additional advantage that it can be tested better.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
@@ -104,7 +104,7 @@ final class PartitionStreamControl private (
         taken <- dataQueue.takeAll.map(_.size)
         _     <- dataQueue.offer(Take.end)
         _     <- queueInfoRef.update(_.withQueueClearedOnLost)
-        _     <- ZIO.logDebug(s"Ignored ${taken} records on lost partition").when(taken != 0)
+        _     <- ZIO.logDebug(s"Ignored $taken records on lost partition").when(taken != 0)
       } yield ()
     }
 
@@ -143,6 +143,7 @@ object PartitionStreamControl {
     tp: TopicPartition,
     requestData: UIO[Unit],
     diagnostics: ConsumerDiagnostics,
+    maxPollRecords: Int,
     maxStreamPullInterval: Duration
   ): UIO[PartitionStreamControl] = {
     val maxStreamPullIntervalNanos = maxStreamPullInterval.toNanos
@@ -167,7 +168,7 @@ object PartitionStreamControl {
           _ <- requestData
           _ <- diagnostics.emit(DiagnosticEvent.Request(tp))
           taken <- dataQueue
-                     .takeBetween(1, Int.MaxValue)
+                     .takeBetween(1, maxPollRecords)
                      .raceFirst(interruptionPromise.await)
         } yield taken
 
@@ -183,9 +184,13 @@ object PartitionStreamControl {
                    } yield ()
                  } *>
                  ZStream.repeatZIOChunk {
-                   // First try to take all records that are available right now.
+                   // First try to take some records that are available right now.
                    // When no data is available, request more data and await its arrival.
-                   dataQueue.takeAll.flatMap(data => if (data.isEmpty) requestAndAwaitData else ZIO.succeed(data))
+                   dataQueue
+                     .takeUpTo(maxPollRecords)
+                     .flatMap { data =>
+                       if (data.isEmpty) requestAndAwaitData else ZIO.succeed(data)
+                     }
                  }.flattenTake.chunksWith { s =>
                    s.tap(records => registerPull(queueInfo, records))
                      // Due to https://github.com/zio/zio/issues/8515 we cannot use Zstream.interruptWhen.
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopConfig.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopConfig.scala
@@ -1,0 +1,54 @@
+package zio.kafka.consumer.internal
+
+import org.apache.kafka.clients.consumer.ConsumerConfig
+import zio._
+import zio.kafka.consumer.ConsumerSettings
+
+import java.util.{ Map => JavaMap }
+
+/**
+ * Configuration for the [[Runloop]].
+ *
+ * See [[ConsumerSettings]] for a description of each config.
+ */
+final case class RunloopConfig(
+  maxPollRecords: Int,
+  maxPollInterval: Duration,
+  maxStreamPullInterval: Duration,
+  maxRebalanceDuration: Duration
+)
+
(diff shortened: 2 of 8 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -38,7 +40,7 @@ abstract class PartitionStream {
 final class PartitionStreamControl private (
   val tp: TopicPartition,
   val stream: ZStream[Any, Throwable, ByteArrayCommittableRecord],
-  dataQueue: Queue[Take[Throwable, ByteArrayCommittableRecord]],
+  dataQueue: Queue[Chunk[ByteArrayCommittableRecord]],
```
guizmaii, line 43: Why changing from `Take[Throwable, ByteArrayCommittableRecord]` to `Chunk[ByteArrayCommittableRecord]`? 🤔

erikvanoosten: In short: it made the code simpler.

Introduced `Chunk` because everything is working (should have been working) with chunks:

> Before this change the partition queues contained records, now they contain chunks of records, and each chunk contains records fetched together in a single broker poll. This has the advantage that the consumer can never get more than max.poll.records records in a zstream chunk.

Got rid of `Take` because:

> Simplify the stream partition queue, the items don't need to represent failure and the end of stream can be represented by an empty chunk. This makes the code easier to read and more performant.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -182,15 +172,20 @@ object PartitionStreamControl {
...
                    // When no data is available, request more data and await its arrival.
-                   dataQueue.takeAll.flatMap(data => if (data.isEmpty) requestAndAwaitData else ZIO.succeed(data))
-                 }.flattenTake.chunksWith { s =>
-                   s.tap(records => registerPull(queueInfo, records))
-                     // Due to https://github.com/zio/zio/issues/8515 we cannot use Zstream.interruptWhen.
-                     .mapZIO(chunk => interruptionPromise.await.whenZIO(interruptionPromise.isDone).as(chunk))
-                 }
+                   dataQueue.poll.flatMap {
+                     case None        => requestAndAwaitData
+                     case Some(taken) => ZIO.succeed(taken)
```
guizmaii, line 180:
```suggestion
                     case Some(taken) => Exit.succeed(taken)
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopConfig.scala
```diff
@@ -0,0 +1,54 @@
...
+import zio.kafka.consumer.ConsumerSettings
+
+import java.util.{ Map => JavaMap }
+
+/**
+ * Configuration for the Runloop.
+ *
+ * See ConsumerSettings for a description of each config.
+ */
+final case class RunloopConfig(
```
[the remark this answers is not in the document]
erikvanoosten: No, most definitely not!
The data it needs is extracted from ConsumerSettings, that class is exposed.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -182,15 +172,20 @@ object PartitionStreamControl {
...
-                 }
+                   dataQueue.poll.flatMap {
+                     case None        => requestAndAwaitData
+                     case Some(taken) => Exit.succeed(taken)
+                   }
+                 }.takeUntil(_ eq EndOfStream)
+                   // Due to https://github.com/zio/zio/issues/8515 we cannot use Zstream.interruptWhen.
+                   .tap(_ => interruptionPromise.await.whenZIO(interruptionPromise.isDone))
+                   // When stream.end was invoked, some chunks of records may be in transit, filter those out.
+                   .filterZIO(_ => hasEndedRef.get.negate)
```
guizmaii, line 186: 🟢 Could be nice to add a variable with a good name which makes this `.filterZIO(_ => hasEndedRef.get.negate)` more readable

```scala
for {
  ....
  hasEndedRef         <- Ref.make(false)
  stillRunning        = hasEndedRef.get.negate
  ...
  _ <- ZSteam.repeatZIO {

           }.takeUntil(_ eq EndOfStream)
             .tap(_ => interruptionPromise.await.whenZIO(interruptionPromise.isDone))
             .filterZIO(_ => stillRunning)

...
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopConfig.scala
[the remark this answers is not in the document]
guizmaii: Then this case class and the companion object should be `private`

zio-kafka/src/main/scala/zio/kafka/consumer/RebalanceListener.scala
```diff
@@ -19,11 +19,14 @@ final case class RebalanceListener(
...
-      lost => onLost(lost) *> that.onLost(lost)
-    )
+    if (that eq RebalanceListener.noop) this
+    else if (this eq RebalanceListener.noop) that
+    else
+      RebalanceListener(
+        assigned => onAssigned(assigned) *> that.onAssigned(assigned),
+        revoked => onRevoked(revoked) *> that.onRevoked(revoked),
+        lost => onLost(lost) *> that.onLost(lost)
+      )
```
guizmaii, line 29: Nice!

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
erikvanoosten: Done in https://github.com/zio/zio-kafka/pull/1576/commits/8c6400b9660ab20852b3e7f6da92edf8f01de39f.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/RunloopConfig.scala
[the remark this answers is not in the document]
erikvanoosten: Right, good point. Indeed, I see we do the same for e.g. Runloop!
Thanks!

guizmaii (approved):

erikvanoosten:
Regarding **solution part 3:** `RangeAssigner` has been made the default when `rebalanceSafeCommits` is enabled in #1597.

PR #1578, opened 2025-09-22
micro optimization: use Exit instead of ZIO

erikvanoosten:
For preformance reasons,use `Exit.succeed` instead of `ZIO.succeed` when we're constructing a pure value.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -131,7 +131,7 @@ private[consumer] final class Runloop private (
     val streams =
       if (tps.isEmpty) Chunk.empty else partitionStreams.filter(streamControl => tps.contains(streamControl.tp))
 
-    if (streams.isEmpty) ZIO.succeed(fulfillResult)
+    if (streams.isEmpty) Exit.succeed(fulfillResult)
     else {
       for {
         consumerGroupMetadata <- groupMetadataRef.get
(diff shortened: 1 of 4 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -131,7 +131,7 @@ private[consumer] final class Runloop private (
     val streams =
       if (tps.isEmpty) Chunk.empty else partitionStreams.filter(streamControl => tps.contains(streamControl.tp))
 
-    if (streams.isEmpty) ZIO.succeed(fulfillResult)
+    if (streams.isEmpty) Exit.succeed(fulfillResult)
```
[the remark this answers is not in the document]
erikvanoosten: I am not sure I fully understand.
```
// zio-kafka:
val z = Exit.success(42)
// user code:
z.flatMap(....executed eagerly?...)
```
In this example the flatMap gets executed eagerly?

zio-kafka/src/main/scala/zio/kafka/serde/Serdes.scala
```diff
@@ -25,15 +25,15 @@ private[zio] trait Serdes {
...
    * are just identity functions.
    *
-   * That allows us to use [[ZIO.succeed]] instead of [[ZIO.attempt]].
+   * That allows us to use [[Exit.succeed]] instead of [[ZIO.attempt]].
    */
   val byteArray: Serde[Any, Array[Byte]] =
     new Serde[Any, Array[Byte]] {
       override final def serialize(topic: String, headers: Headers, value: Array[Byte]): RIO[Any, Array[Byte]] =
-        ZIO.succeed(value)
+        Exit.succeed(value)
```
jivanic-demystdata, line 33: Same here. You probably can't use `Exit` here

Only use `Exit` when it's internal code and when it's inside a `.flatMap`, etc.

zio-kafka/src/main/scala/zio/kafka/serde/Serializer.scala
```diff
@@ -40,7 +40,7 @@ trait Serializer[-R, -A] {
   def asOption[A1 <: A]: Serializer[R, Option[A1]] =
     Serializer { (topic, headers, valueOpt) =>
       valueOpt match {
-        case None        => ZIO.succeed(null)
+        case None        => Exit.succeed(null)
```
jivanic-demystdata, line 43: Same, you probably shouldn't be using `Exit` here

erikvanoosten:
Closing. It's not worth the trouble.

PR #1590, opened 2025-10-06
Make sure producer can shut down

### LINKED ISSUE #1588: Producer hangs on close when wrapped in `Reloadable`
#### Context

Starting with ZIO 2.1.0, `acquire` in `ScopedRef` was made uninterruptible (see https://github.com/zio/zio/pull/8638). Since `Reloadable` relies on `ScopedRef` internally, it's directly affected by this change. 
As a result, using `Reloadable[Producer]` no longer closes the producer properly and instead hangs indefinitely.

#### Problem

The likely culprit are the following two lines in [Producer.scala](https://github.com/zio/zio-kafka/blob/master/zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala#L270):
```scala
 _ <- sendQueue.size.flatMap(metrics.observeSendQueueSize).schedule(Schedule.fixed(10.seconds)).forkScoped
 _ <- ZIO.blocking(producer.sendFromQueue).forkScoped
```

Both effects are forked with `forkScoped` while in an uninterruptible region, which prevents the fibers from being interrupted when the scope is closed.

Making both effects interruptible appears to resolve the issue, but I'm not entirely sure whether this change could introduce other unexpected problems.

#### Minimal reproduction
```scala
//> using scala 3.3.6
//> using dep dev.zio::zio:2.1.21

import zio.*

object Example extends ZIOAppDefault:

  val layer = ZLayer.scoped:
    ZIO.never
      // .interruptible
      .onInterrupt(ZIO.log("Interrupted"))
      .forkScoped

  override val run =
    ZIO.unit.provideLayer(layer.reloadableManual)

end Example
```

erikvanoosten:
As documented in #1588, when a producer is constructed in an uninterruptable region (for example because it is wrapped in a `Reloadable` layer), it can not shut down. We fix this by making the producer's forked fibers interruptable.

Fixes #1588.

### DIFF
--- a/zio-kafka-test/src/test/scala/zio/kafka/ProducerSpec.scala
+++ b/zio-kafka-test/src/test/scala/zio/kafka/ProducerSpec.scala
@@ -7,13 +7,14 @@ import zio._
 import zio.kafka.admin.AdminClient.NewTopic
 import zio.kafka.consumer._
 import zio.kafka.producer.TransactionalProducer.{ TransactionLeaked, UserInitiatedAbort }
-import zio.kafka.producer.{ ByteRecord, Transaction }
+import zio.kafka.producer.{ ByteRecord, Producer, Transaction }
 import zio.kafka.serde.Serde
 import zio.kafka.testkit._
 import zio.stream.Take
 import zio.test.Assertion._
 import zio.test.TestAspect._
 import zio.test._
+
 import java.nio.charset.StandardCharsets
 import java.util.UUID
 
@@ -30,7 +31,7 @@ object ProducerSpec extends ZIOSpecDefaultSlf4j with KafkaRandom {
       c.plainStream(subscription, Serde.string, Serde.int).toQueue()
     }
 
-  private val produceSpec =
+  private val produceSpec: Spec[Scope & Kafka, Throwable] =
     suite("::produce(record: ProducerRecord[Array[Byte], Array[Byte]])")(
       test("produces messages") {
         for {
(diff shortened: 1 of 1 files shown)


zio-kafka/src/main/scala/zio/kafka/producer/Producer.scala
```diff
@@ -266,8 +266,11 @@ object Producer {
         )
       metrics  = new ZioProducerMetrics(settings.metricLabels)
       producer = new ProducerLive(settings, wrappedDiagnostics, javaProducer, runtime, sendQueue, metrics)
-      _ <- sendQueue.size.flatMap(metrics.observeSendQueueSize).schedule(Schedule.fixed(10.seconds)).forkScoped
-      _ <- ZIO.blocking(producer.sendFromQueue).forkScoped
+      // forkScoped needs to run in an interruptable region, otherwise the fibers can't be interrupted #1588
+      _ <- ZIO.interruptible {
+             sendQueue.size.flatMap(metrics.observeSendQueueSize).schedule(Schedule.fixed(10.seconds)).forkScoped *>
+               ZIO.blocking(producer.sendFromQueue).forkScoped
+           }
```
Mee-Tree, line 273: Shouldn’t we make only the child fiber interruptible?

Like this:
```scala
ZIO.blocking(producer.sendFromQueue)
  .interruptible
  .forkScoped
```

Mee-Tree: The proposed solution also works, since the forked fiber inherits the parent’s interruption status, but making only the child interruptible would be more precise, I think.

https://zio.dev/reference/fiber/fiber.md/#fork-and-join

erikvanoosten:
Added a unit test. I am assuming this is good enough. Going to merge as soon as its green 🙂

PR #1605, opened 2025-11-07
Remove KAFKA-18818 workaround

erikvanoosten:


erikvanoosten:
See discussion in https://issues.apache.org/jira/browse/KAFKA-18818, we're not there yet. Closing...

PR #1667, opened 2026-03-18
Limit number of concurrent admin write operations

erikvanoosten:
It seems Kafka does not handle many concurrent admin write operations well. As zio-kafka executes many tests in parallel, a massive amount of topics is created and deleted during the tests. It seems this is the cause for the flaky tests that have plagued the project for a long time.

In this change we limit the number of concurrent admin write operations to 5.
We also remove the kafka18818-workaround as it is apparently no longer needed (it seems Kafka-18818 was really fixed in Kafka 4.0 after all).

WARNING: the AdminClient is source compatible, but _not_ binary compatible with the previous version.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -38,8 +38,10 @@ Kafka has a mature Java client for producing and consuming events, but it has a 
   - allows for batches for highest throughput
   - optionally await broker acknowledgements
   - optional retries after authentication/authorization errors
+  - exposes metrics
 - Admin API:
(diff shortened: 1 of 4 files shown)


guizmaii (approved):
✅ LGTM. See suggestion

erikvanoosten:
> It seems this is the cause for the flaky tests that have plagued the project for a long time.

Too bad, there are less flaky errors, but not zero.

PR #1711, opened 2026-06-02
Fix: Consumer.consumeWith hangs indefinitely after TopicAuthorizationException

### LINKED ISSUE #1: Consider using `interruptible`
Instead of:

```scala
blocking(ZIO(...))
```

Consider:

```scala
interruptible(...)
```

Both will be placed on the blocking thread pool, but the latter can be interrupted.

acrow:
Problem

  When a TopicAuthorizationException is thrown at runtime (e.g., an ACL is revoked while the consumer is running), Consumer.consumeWith hangs indefinitely instead of propagating the error. We encountered this in production where a zombie pod would stop consuming but never exit, preventing Kubernetes from restarting it.
  
  We traced it to three related issues:

  1. Silent poll: poll() silently returns 0 records on DENY READ instead of throwing, so the runloop never sees the error.
  2. Blocked partition streams: When the runloop crashes, per-partition stream fibers blocked on dataQueue.take never wake up.
  3. offerAndAwaitCommand hangs: After the runloop dies, callers like removeSubscription block forever waiting for a promise that will never be fulfilled.

  Proposed fixes

  1. Silent-poll probe (Runloop.scala): after silentPollThreshold=3 consecutive empty polls on a topic, call partitionsFor() to force a metadata fetch that surfaces any TopicAuthorizationException.
  2. halt() unblocking (PartitionStreamControl.scala): offer Exit.fail to dataQueue so blocked fibers wake up immediately.
  3. runloopDone promise (Runloop.scala): complete a promise when the runloop exits, so offerAndAwaitCommand can unblock promptly.

  Tests

  Added two tests to RunloopSpec:
  - partition stream fails when Runloop crashes after auth retries exhaust with empty dataQueue
  - removeSubscription does not hang after Runloop crashes
  
  All existing tests continue to pass.

  Notes

  We're not entirely sure the silent-poll probe approach (fix #1) is the best design — it adds some complexity. We're open to suggestions on a better way to
  surface the error from poll(). Fixes #2 and #3 feel more straightforward.

  Any feedback is very welcome. Thank you for maintaining this library!

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
@@ -92,7 +92,9 @@ final class PartitionStreamControl private (
    * To be invoked when the stream is no longer processing. It fails the stream with the given cause.
    */
   private[internal] def halt(cause: Cause[Throwable]): UIO[Unit] =
-    interruptionPromise.failCause(cause).unit
+    interruptionPromise.failCause(cause).ignore *>
+      // Offer a failure element so that dataQueue.take unblocks immediately
+      dataQueue.offer(Exit.fail(cause.failureOption)).unit
 
   /** To be invoked when the partition was revoked, lost or otherwise needs to be ended. */
   private[internal] def end: ZIO[Any, Nothing, Unit] =
@@ -154,7 +156,7 @@ object PartitionStreamControl {
         for {
           _     <- requestData
           _     <- diagnostics.emit(DiagnosticEvent.Request(tp))
-          taken <- dataQueue.take.raceFirst(interruptionPromise.await).mapError(Option.apply).unexit
+          taken <- dataQueue.take.flatMap(exit => exit)
         } yield taken
       notEnded = hasEndedRef.get.negate
 
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
@@ -93,7 +93,8 @@ final class PartitionStreamControl private (
    */
   private[internal] def halt(cause: Cause[Throwable]): UIO[Unit] =
     interruptionPromise.failCause(cause).ignore *>
-      // Offer a failure element so that dataQueue.take unblocks immediately
+      // Also offer a failure to dataQueue so that a fiber currently waiting on dataQueue.take
+      // can unblock without waiting for the next poll cycle.
       dataQueue.offer(Exit.fail(cause.failureOption)).unit
 
   /** To be invoked when the partition was revoked, lost or otherwise needs to be ended. */
@@ -156,7 +157,7 @@ object PartitionStreamControl {
         for {
           _     <- requestData
           _     <- diagnostics.emit(DiagnosticEvent.Request(tp))
-          taken <- dataQueue.take.flatMap(exit => exit)
+          taken <- dataQueue.take.raceFirst(interruptionPromise.await).mapError(Option.apply).unexit
         } yield taken
       notEnded = hasEndedRef.get.negate
 
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -33,7 +33,7 @@ private[consumer] final class Runloop private (
   metricsObserver: ConsumerMetricsObserver,
   committer: Committer,
   groupMetadataRef: Ref[Option[ConsumerGroupMetadata]],
-  runloopDone: Promise[Nothing, Unit]
+  runloopDone: Promise[Throwable, Nothing]
 ) {
   private def newPartitionStream(tp: TopicPartition): UIO[PartitionStreamControl] =
     PartitionStreamControl.newPartitionStream(
@@ -71,15 +71,19 @@ private[consumer] final class Runloop private (
   private[internal] def endStreamsBySubscription(subscription: Subscription): UIO[Unit] =
     offerAndAwaitCommand(RunloopCommand.EndStreamsBySubscription(subscription, _))
 
-  private[internal] def offerAndAwaitCommand[E, A](f: Promise[E, A] => RunloopCommand): ZIO[Any, E, A] =
-    Promise.make[E, A].flatMap { promise =>
-      commandQueue.offer(f(promise)) *>
-        // Complete the promise via runloopDone if the runloop exits before processing it.
-        // This prevents callers (like removeSubscription) from blocking forever.
-        runloopDone.await.flatMap(_ => promise.succeed(().asInstanceOf[A]).unit).forkDaemon.flatMap { completionFiber =>
-          promise.await.ensuring(completionFiber.interrupt.ignore)
+  private[internal] def offerAndAwaitCommand[E, A](f: Promise[E, A] => RunloopCommand): ZIO[Any, E, A] = {
+    // If the runloop is already done, skip the commandQueue.offer entirely — the runloop
+    // is no longer processing commands, and offering to a shutting-down queue can block.
+    val runloopDoneAsEA: ZIO[Any, E, A] = runloopDone.await.asInstanceOf[ZIO[Any, E, A]]
+    runloopDone.isDone.flatMap {
+      case true => runloopDoneAsEA
+      case false =>
+        Promise.make[E, A].flatMap { promise =>
+          commandQueue.offer(f(promise)) *>
+            promise.await.raceFirst(runloopDoneAsEA)
         }
     }
+  }
 
   private def makeRebalanceListener: ConsumerRebalanceListener = {
     // Here we just want to avoid any executor shift if the user provided listener is the noop listener.
@@ -214,11 +218,12 @@ private[consumer] final class Runloop private (
           settings.authErrorRetrySchedule
       )
 
-  private def handlePoll(
-    state: State,
-    silentPollCountsRef: Ref[Map[String, Int]],
-    silentPollThreshold: Int
-  ): Task[State] = {
+  private def handlePoll(state: State): Task[State] = {
+    // Subscribed topics that have no assigned streams yet (e.g. ACL DENY before consumer group joins)
+    val unassignedTopics: Set[String] = state.subscriptionState match {
+      case SubscriptionState.Subscribed(_, Subscription.Topics(topics)) if state.assignedStreams.isEmpty => topics
+      case _                                                                                             => Set.empty
+    }
     for {
       runningStreamsBeforePoll <- ZIO.filterNot(state.assignedStreams)(_.hasEnded)
       partitionsToFetch        <- settings.fetchStrategy.selectPartitionsToFetch(runningStreamsBeforePoll)
@@ -229,7 +234,7 @@ private[consumer] final class Runloop private (
                s" resuming ${partitionsToFetch.size} out of ${state.assignedStreams.size} partitions"
            )
       _ <- currentStateRef.set(state)
-      pollResult <-
+      pollResultAndCounts <-
         consumer.runloopAccess { c =>
           for {
             assignment <- ZIO.attempt(c.assignment())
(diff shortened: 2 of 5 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -154,7 +156,7 @@ object PartitionStreamControl {
         for {
           _     <- requestData
           _     <- diagnostics.emit(DiagnosticEvent.Request(tp))
-          taken <- dataQueue.take.raceFirst(interruptionPromise.await).mapError(Option.apply).unexit
+          taken <- dataQueue.take.flatMap(exit => exit)
```
erikvanoosten: `interruptionPromise` was introduced to quickly exit partition stream control during a rebalance. For this use case, `dataQueue` might not be empty.

For this PR, the changes in this class are not needed. Interruption can continue to be done through `interruptionPromise`, to be used via `halt()`.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -210,7 +214,11 @@ private[consumer] final class Runloop private (
           settings.authErrorRetrySchedule
       )
 
-  private def handlePoll(state: State): Task[State] = {
+  private def handlePoll(
+    state: State,
+    silentPollCountsRef: Ref[Map[String, Int]],
+    silentPollThreshold: Int
```
erikvanoosten: It is not necessary to introduce a new state variable. It is also not necessary to wrap the Map in a Ref. Both things can be fixed by adding a `Map` field to `State`.

There is a (theoretical) problem with using a Map though: it might forever grow and never shrink. Therefore I propose to store the opposite: `pollCountToMetaRefresh` (or something like that). We only need to store (or decrease) the count for a topic when a poll was empty, as soon as a non-empty is seen, we remove the topic from the map. The partitions are fetched when the count reaches `0`. This keeps the map small and map traversal fast during normal operation.

BTW, I am not thrilled having to add such a heavy data structure to the hot loop of zio-kafka.

Re. `silentPollThreshold`, isn't there a `consumerSettings` field we can directly access? (Update: yes! `settings` is reacheble here.)

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -545,36 +581,51 @@ private[consumer] final class Runloop private (
...
-      .catchAllCause { cause =>
-        for {
-          state <- currentStateRef.get
-          _     <- ZIO.foreachDiscard(state.assignedStreams)(_.halt(cause))
-          _     <- partitionsHub.offer(Take.failCause(cause))
-        } yield ()
-      }
+    // Track how many consecutive polls each topic has been in partitionsToFetch but returned no
+    // records. Once a topic exceeds the threshold we call partitionsFor() to force an auth check.
+    val silentPollThreshold = 3
```
erikvanoosten: `silentPollThreshold` needs to be an optional consumer setting. Some people run a single kafka consumer against thousands of topics. In this use case it is normal that many topics are silent for a long time. At the same time other topics are alive and kicking, we don't want to hurt performance for topics that are active with the extra metadata refreshes.

erikvanoosten: The name does not express its purpose. The word 'silent' doesn't work for me (in my mental model polls don't make sound). 'Empty' would be better.

WDYT of something like `emptyPollCountToMetaRefresh`? Also not great yet, better name needed...

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -232,6 +240,34 @@ private[consumer] final class Runloop private (
...
+                               topicsToTrack.foldLeft(counts) { (acc, topic) =>
+                                 if (polledTopics.contains(topic)) acc - topic
+                                 else acc + (topic -> (acc.getOrElse(topic, 0) + 1))
+                               }
+                             }
+            silentTopicsToProbe = topicsToTrack.filter(t => updatedCounts.getOrElse(t, 0) >= silentPollThreshold)
+            _ <- ZIO.foreachDiscard(silentTopicsToProbe) { silentTopic =>
+                   // partitionsFor() forces a metadata fetch and throws TopicAuthorizationException
+                   // if the topic is unauthorized — unlike poll() which silently swallows it.
+                   ZIO.attempt(c.partitionsFor(silentTopic))
```
erikvanoosten: Did you validate that this triggers an Auth*Exception? The javadoc mentions that it can cache the partitions list on the consumer. I am therefore not certain that this forces a metadata refresh.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -71,11 +72,14 @@ private[consumer] final class Runloop private (
...
-    for {
-      promise <- Promise.make[E, A]
-      _       <- commandQueue.offer(f(promise))
-      r       <- promise.await
-    } yield r
+    Promise.make[E, A].flatMap { promise =>
+      commandQueue.offer(f(promise)) *>
+        // Complete the promise via runloopDone if the runloop exits before processing it.
+        // This prevents callers (like removeSubscription) from blocking forever.
+        runloopDone.await.flatMap(_ => promise.succeed(().asInstanceOf[A]).unit).forkDaemon.flatMap { completionFiber =>
```
erikvanoosten: Adding `runloopDone` brings yet another sync point. We have so many already. To be investigated if we need this...

Yeah, I think we can work around this by storing the outstanding promises somewhere, and completing those with an error when the runloop halts. Not great either, I will try to think of something else.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -92,7 +92,9 @@ final class PartitionStreamControl private (
    * To be invoked when the stream is no longer processing. It fails the stream with the given cause.
    */
   private[internal] def halt(cause: Cause[Throwable]): UIO[Unit] =
-    interruptionPromise.failCause(cause).unit
+    interruptionPromise.failCause(cause).ignore *>
+      // Offer a failure element so that dataQueue.take unblocks immediately
+      dataQueue.offer(Exit.fail(cause.failureOption)).unit
```
erikvanoosten: This comment is not correct. It would only unblock immediately when it is currently trying to pull data, AND, there is no other data in `dataQueue`. Zio-kafka tries very hard to always have _some_ data in `dataQueue`.

If you observe that partition stream control is blocked, it is either because the fiber is actually processing data (the fiber might be blocked elsewhere), or because `halt()` was not invoked.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: The expression is quite complex, I am not even sure what its doing.
Would this work as well?

```scala
promise.await.raceFirst(runloopDone.await)
```

(Where `runloopDone` should be of type `Promise[Throwable, Nothing]`.)

erikvanoosten (commented):
Hi @acrow, I am glad the previous bug was fixed. Always interesting to see how people use Kafka. This use case was never brought to our attention before!

Thanks for this PR as well! There are a few wrinkles in your proposal that can be fixed or improved. However, my main worry is that I am not certain that `partitionsFor` really refreshes metadata. WDYT?

I also think that the extra check should be opt-in and the poll count configurable. That means it needs to be added to ConsumerSettings.

acrow:
Hi @erikvanoosten, thank you for the detailed review. We have investigated all your points and want to share our findings before updating the code.

  On partitionsFor() refreshing metadata (Comment 6)

  We validated this locally with debug logging. In our test environment, partitionsFor() does throw
  TopicAuthorizationException when the topic has a DENY DESCRIBE ACL — the stack trace confirms the exception originates from Runloop.scala line 255 (our partitionsFor() call), not from poll(). So Fix #1 works as intended for the silent-poll scenario.

  On the consumeWith hang — a deeper finding

  While testing we instrumented the full consumeWith chain with step-by-step logging. We found the following sequence after ACL DENY is applied:

  1. Runloop crashes → catchAllCause fires ✓
  2. All per-partition streams receive halt(cause) and fail ✓
  3. flatMapPar reports failure ✓
  4. Commit fails ✓
  5. All inner streams end ✓
  6. subscribe scope finalizer starts calling removeSubscription — but never completes ✗

  The hang is in removeSubscription → offerAndAwaitCommand. Even though runloopDone was already signalled,
  commandQueue.offer(f(promise)) appears to block when called concurrently with the Runloop.make scope finalizer shutting down the commandQueue.
  
  This means the consumeWith hang is caused by a race between removeSubscription and commandQueue.shutdown in the scope finalizers. We will investigate the fix for this tomorrow, but we wanted to share the finding now.

  On all other feedback — we have the changes ready locally: emptyPollCountToMetaRefresh as an opt-in ConsumerSettings field, count moved into State, offerAndAwaitCommand simplified with raceFirst, interruptionPromise restored, comment corrected. We will push these once we resolve the consumeWith hang.
  
  Thank you again for the review.

erikvanoosten:
Thanks, that is all good news. I'll wait for further pushes.

acrow:
Hi @erikvanoosten, the updated code is now in commit 47c1bd0. Here is how each point was addressed:
  
  Comment 1 & PartitionStreamControl — interruptionPromise is restored in requestAndAwaitData. The changes to PartitionStreamControl are now limited to the comment fix in halt().

  Comment 2 (State field) — The empty-poll count is now a Map[String, Int] field on State instead of a Ref. Entries are removed when a topic returns records, so the map does not grow unboundedly for topics that are no longer silent.

  Comment 3 & 4 (opt-in, rename) — The check is now opt-in via  ConsumerSettings.withEmptyPollCountToMetaRefresh(n) (disabled by default). The name follows your suggestion.

  Comment 5 (partitionsFor() caching concern) — You were right to be uncertain. We replaced partitionsFor() with committed(). In our local tests, committed() reliably throws TopicAuthorizationException on DENY READ without using cached metadata, whereas partitionsFor() only throws on DENY DESCRIBE.

  Comment 6 & 8 (offerAndAwaitCommand) — We found the root cause through instrumentation: removeSubscription (called from the subscribe scope finalizer) was calling commandQueue.offer while the Runloop.make scope finalizer was concurrently shutting down the commandQueue, causing the offer to block. The fix checks runloopDone.isDone first — if already done, the offer is skipped entirely and runloopDone.await is returned directly. runloopDone is now Promise[Throwable, Nothing] as you suggested, and offerAndAwaitCommand uses raceFirst for the normal (not-yet-done) path.             

  Comment 7 (halt() comment) — Fixed. The comment now correctly says the offer to dataQueue unblocks a fiber that is currently waiting on dataQueue.take, not that it unblocks immediately in all cases.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -92,7 +92,10 @@ final class PartitionStreamControl private (
    * To be invoked when the stream is no longer processing. It fails the stream with the given cause.
    */
   private[internal] def halt(cause: Cause[Throwable]): UIO[Unit] =
-    interruptionPromise.failCause(cause).unit
+    interruptionPromise.failCause(cause).ignore *>
+      // Also offer a failure to dataQueue so that a fiber currently waiting on dataQueue.take
+      // can unblock without waiting for the next poll cycle.
+      dataQueue.offer(Exit.fail(cause.failureOption)).unit
```
erikvanoosten, line 98: The exit element on the queue will be queued behind other elements. Therefore `interruptionPromise` will abort this stream control earlier than the added exit element can. Please revert this change.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -70,12 +71,19 @@ private[consumer] final class Runloop private (
...
+    val runloopDoneAsEA: ZIO[Any, E, A] = runloopDone.await.asInstanceOf[ZIO[Any, E, A]]
+    runloopDone.isDone.flatMap {
+      case true => runloopDoneAsEA
+      case false =>
+        Promise.make[E, A].flatMap { promise =>
+          commandQueue.offer(f(promise)) *>
+            promise.await.raceFirst(runloopDoneAsEA)
+        }
+    }
+  }
```
erikvanoosten, line 86: The cast to `ZIO[Any, E, A]` is unsound. `runloop` fails with a `Throwable`, you can't cast it to some unknown `E`. We can fix this by making this function fail on Throwable and putting an upper type bound on `E` to make sure it is compatible with `Throwable`.

There is race condition between checking `runloopDone` and offering to a potentially shutting down queue.

This solves the race, and attempts to fix the unsound type cast:

```suggestion
  private[internal] def offerAndAwaitCommand[E <: Throwable, A](f: Promise[E, A] => RunloopCommand): ZIO[Any, Throwable, A] = {
    Promise.make[E, A].flatMap { promise => commandQueue.offer(f(promise)) *> promise.await }.raceFirst(runloopDone.await)
  }
```

Unfortunately, this new function signature has more consequences. In particular, method `runFoldChunksDiscardZIO` needs to be updated. I tried finding a solution but I am too tired to think it through. I'll take another look later on. I'd appreciate your idea here as well.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -211,6 +219,11 @@ private[consumer] final class Runloop private (
       )
 
   private def handlePoll(state: State): Task[State] = {
+    // Subscribed topics that have no assigned streams yet (e.g. ACL DENY before consumer group joins)
+    val unassignedTopics: Set[String] = state.subscriptionState match {
+      case SubscriptionState.Subscribed(_, Subscription.Topics(topics)) if state.assignedStreams.isEmpty => topics
+      case _                                                                                             => Set.empty
+    }
```
erikvanoosten, line 226: Set operations are serious CPU eaters. And this part is not opt-in.

erikvanoosten (commented):
Hi @acrow, thanks for the changes. Coming 2 weeks I have less time to review and think along than what this problem deserves.

Great that you found a better function in the form of`committed`.
I made some comments on the new changes.

Now that I am taking a deeper look, I unfortunately have a new worry. I see that this PR adds book keeping which hurts throughput for the majority of zio-kafka users which don't use topic permissions

WDYT of splitting this PR in 2? The first parts addresses the dead lock where a partition stream is not halted, introduces the runloopDone promise and makes all command's promises have Throwable as error type.

The second part would focus on a way to fail when a topic is denied. The trick to look for empty polls is pretty smart, but perhaps we can find a simpler way that requires less computing. Or perhaps we can find a way that makes the extra book keeping only active when this feature is enabled.

acrow:
Hi @erikvanoosten, thank you for the continued review. Here are our findings after testing your suggestions:
  
  On dataQueue.offer in halt() (comment 10 & 15):
  We tested removing it — consumeWith hangs again in our live tests. We believe interruptionPromise handles the case where a fiber is currently waiting on
  dataQueue.take, but dataQueue.offer handles the case where the fiber is between polls (not currently waiting). Without it, the fiber misses the
  interruption signal until the next poll cycle, which in our scenario never comes because the runloop is already dead. Could you help us understand the
  scenario you have in mind where it is not needed?
  
  On offerAndAwaitCommand (comment 9 & 16):
  You are right about the race condition in the isDone check — there is a window between isDone = false and commandQueue.offer where the runloop can die. We
  tried your suggested form but could not get it to compile without a cast due to Scala 2.13 inference. Could you clarify what change runFoldChunksDiscardZIO
  needs? That would help us implement your suggestion correctly.

  On Set operations (comment 17):
  Accepted — all bookkeeping is now skipped entirely when emptyPollCountToMetaRefresh is None. Zero overhead for users who do not enable it.

  Splitting the PR:
  We agree with your suggestion. We will close this PR and open two focused ones — one for the runloop deadlock fix and one for the empty-poll detection
  feature. Thank you for your patience and thorough review.

erikvanoosten:
Hi @acrow ,

I tried to apply my own suggestions but found that they are aren't (all) good: the tests block immediately. At least I managed to get rid of the type issues by using Throwable as upper type for the error channel. You can see my attempt in branch https://github.com/zio/zio-kafka/tree/fix-hang-at-error. Feel free to take over in your branch any part you find useful.

> On dataQueue.offer in halt() (comment 10 & 15): We tested removing it — consumeWith hangs again in our live tests. We believe interruptionPromise handles the case where a fiber is currently waiting on dataQueue.take, but dataQueue.offer handles the case where the fiber is between polls (not currently waiting). Without it, the fiber misses the interruption signal until the next poll cycle, which in our scenario never comes because the runloop is already dead. 

This is really strange, in my mental model ZIO.raceFirst starts 2 fibers and completes as soon as one of its child fibers complete. Therefore, a blocked dataQueue.take should not prevent ZIO.raceFirst from completing.

There is a thing that you need to keep in mind when using raceFirst, which is that raceFirst waits until both fibers complete all their finalizers. This can be prevented by using `disconnect`, e.g. `zio1.disconnect.raceFirst(zio2.disconnect)`.

Looking forward to your separate PRs.

acrow:
Hi @erikvanoosten, thank you for trying the fix yourself and sharing your branch — it was very helpful.
  
  We adopted your approach: offerAndAwaitCommand now returns Task[A] (no cast needed), and the type widening in RunloopCommand and RunloopAccess is in place.

  However, we found that raceFirst alone still causes hangs in our live tests — which matches what you observed yourself. The reason, as you pointed out, is
  that raceFirst waits for both fibers' finalizers to complete.

  We chose the isDone check over .disconnect because it avoids starting the second fiber entirely when the runloop is already done, rather than abandoning it
  mid-flight. This felt cleaner since commandQueue.offer would otherwise still execute in the background after .disconnect.

  We also kept dataQueue.offer(Exit.fail(...)) in halt() as our live tests still hang without it.

  The two separate PRs are now open:
  - #1712 — the deadlock fix (runloopDone + offerAndAwaitCommand)
  - #1713 — the opt-in empty-poll probe (withEmptyPollCountToMetaRefresh)

erikvanoosten:
Discussion continued in:
- #1712
- #1713

PR #1712, opened 2026-06-08
Fix: removeSubscription hangs after Runloop crashes

acrow:
The consumer doesn't always cleanly shut down after errors.

## Problem

When the Runloop crashes (e.g. after `TopicAuthorizationException` exhausts retries), scope finalizers for active subscriptions call `removeSubscription` → `offerAndAwaitCommand`. At this point the `Runloop.make` scope finalizer may be concurrently shutting down the `commandQueue`.  This causes `commandQueue.offer` to block, leaving `removeSubscription` hung indefinitely, never aborting the partition stream, letting `Consumer.consumeWith` hang forever.

## Fixes

1. Add `runloopDone: Promise[Throwable, Nothing]` that is completed (failed with the runloop cause) when the runloop exits. `offerAndAwaitCommand` offers the command and then checks `runloopDone.isDone` to see if the commands promise will ever be fulfilled:
  - If already done → don't wait on the command promise and return `runloopDone.await` directly
  - If not done → wait for `promise.await` (Note: this is wrong (caused by patch requested by maintainer (sorry)), we should race against `runloopDone.await` as well, to be fixed in next PR.) Update: fixed in #1733.

2. Ensure partition stream control can always end by racing `interruptionPromise.await` also when polling for more data.

## Also

- `RunloopAccess` changed `removeSubscription.orDie` to `.ignore` since failure is expected when the runloop has already crashed.
- Small optimization in partition stream control (the `Some` wrapper is only needed in error path).
- Method `shouldPoll` made private.

## Tests

Added `RunloopSpec` test: removeSubscription does not hang after Runloop crashes — verifies that `removeSubscription` completes promptly (does not time out) after the runloop crashes.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
@@ -154,7 +154,7 @@ object PartitionStreamControl {
         for {
           _     <- requestData
           _     <- diagnostics.emit(DiagnosticEvent.Request(tp))
-          taken <- dataQueue.take.raceFirst(interruptionPromise.await).mapError(Option.apply).unexit
+          taken <- dataQueue.take.raceFirst(interruptionPromise.await.mapError(Option.apply)).unexit
         } yield taken
       notEnded = hasEndedRef.get.negate
 
@@ -172,7 +172,8 @@ object PartitionStreamControl {
                  ZStream.repeatZIOOption {
                    // First try to take a chunk of records that are available right now.
                    // When no data is available, request more data and await its arrival.
-                   dataQueue.poll.flatMap {
+                   // Race with interruptionPromise so that halt() unblocks this path immediately.
+                   dataQueue.poll.raceFirst(interruptionPromise.await.mapError(Option.apply)).flatMap {
                      case None       => requestAndAwaitData
                      case Some(exit) => exit
                    }
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -32,7 +32,8 @@ private[consumer] final class Runloop private (
   rebalanceCoordinator: RebalanceCoordinator,
   metricsObserver: ConsumerMetricsObserver,
   committer: Committer,
-  groupMetadataRef: Ref[Option[ConsumerGroupMetadata]]
+  groupMetadataRef: Ref[Option[ConsumerGroupMetadata]],
+  runloopDone: Promise[Throwable, Nothing]
 ) {
   private def newPartitionStream(tp: TopicPartition): UIO[PartitionStreamControl] =
     PartitionStreamControl.newPartitionStream(
@@ -67,14 +68,27 @@ private[consumer] final class Runloop private (
   private[internal] def removeSubscription(subscription: Subscription): Task[Unit] =
     offerAndAwaitCommand(RunloopCommand.RemoveSubscription(subscription, _))
 
-  private[internal] def endStreamsBySubscription(subscription: Subscription): UIO[Unit] =
+  private[internal] def endStreamsBySubscription(subscription: Subscription): Task[Unit] =
     offerAndAwaitCommand(RunloopCommand.EndStreamsBySubscription(subscription, _))
 
   private[internal] def offerAndAwaitCommand[E, A](f: Promise[E, A] => RunloopCommand): ZIO[Any, E, A] =
     for {
       promise <- Promise.make[E, A]
       _       <- commandQueue.offer(f(promise))
-      r       <- promise.await
+      // If the runloop already crashed (runloopDone completed before our offer was processed),
+      // `drainCommandQueue` may have already run and missed our command. Fail the promise now so
+      // `promise.await` below does not block forever. The `asInstanceOf` is safe: all command
+      // error types (InvalidSubscriptionUnion, Throwable) are subtypes of Throwable, and ZIO
+      // stores Cause[E] as-is without type-erased differences at runtime.
+      _ <- runloopDone.isDone.flatMap { done =>
+             if (done)
+               runloopDone.await.exit.flatMap {
+                 case Exit.Failure(cause) => promise.failCause(cause.asInstanceOf[Cause[E]]).unit
+                 case _                   => ZIO.unit
+               }
+             else ZIO.unit
+           }
+      r <- promise.await
     } yield r
 
   private def makeRebalanceListener: ConsumerRebalanceListener = {
@@ -573,13 +587,28 @@ private[consumer] final class Runloop private (
           state <- currentStateRef.get
           _     <- ZIO.foreachDiscard(state.assignedStreams)(_.halt(cause))
           _     <- partitionsHub.offer(Take.failCause(cause))
+          _     <- runloopDone.failCause(cause).ignore
+          _     <- drainCommandQueue(cause)
         } yield ()
       }
+      .zipLeft(runloopDone.fail(new RuntimeException("Consumer shutting down")).ignore)
   }
 
   def shouldPoll(state: State): UIO[Boolean] =
     committer.pendingCommitCount.map { pendingCommitCount =>
       state.subscriptionState.isSubscribed && (state.pendingRequests.nonEmpty || pendingCommitCount > 0 || state.assignedStreams.isEmpty)
+    }
+
+  // Fail all pending promise-bearing commands so that any callers waiting on them are unblocked immediately.
+  private def drainCommandQueue(cause: Cause[Throwable]): UIO[Unit] =
+    commandQueue.takeAll.flatMap { commands =>
+      ZIO.foreachDiscard(commands) {
+        case RunloopCommand.AddSubscription(_, cont) =>
+          cont.failCause(cause.asInstanceOf[Cause[InvalidSubscriptionUnion]]).unit
+        case RunloopCommand.RemoveSubscription(_, cont)       => cont.failCause(cause).unit
+        case RunloopCommand.EndStreamsBySubscription(_, cont) => cont.failCause(cause).unit
+        case _                                                => ZIO.unit
+      }
     }
 
   private def observeRunloopMetrics(runloopMetricsSchedule: Schedule[Any, Unit, Long]): ZIO[Any, Nothing, Unit] = {
@@ -658,6 +687,7 @@ object Runloop {
   ): URIO[Scope, Runloop] =
     for {
       _                  <- ZIO.addFinalizer(diagnostics.emit(DiagnosticEvent.RunloopFinalized))
+      runloopDone        <- Promise.make[Throwable, Nothing]
       commandQueue       <- ZIO.acquireRelease(Queue.unbounded[RunloopCommand])(_.shutdown)
       lastRebalanceEvent <- Ref.Synchronized.make[RebalanceEvent](RebalanceEvent.None)
       initialState = State.initial
@@ -694,7 +724,8 @@ object Runloop {
                   metricsObserver = metricsObserver,
                   rebalanceCoordinator = rebalanceCoordinator,
                   committer = committer,
-                  groupMetadataRef = groupMetadataRef
+                  groupMetadataRef = groupMetadataRef,
+                  runloopDone = runloopDone
                 )
       _ <- ZIO.logDebug("Starting Runloop")
 
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -58,7 +58,7 @@ private[consumer] final class Runloop private (
         )
         .unit
 
-  private[internal] def addSubscription(subscription: Subscription): IO[InvalidSubscriptionUnion, Unit] =
+  private[internal] def addSubscription(subscription: Subscription): Task[Unit] =
     for {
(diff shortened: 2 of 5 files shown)


zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -545,36 +551,43 @@ private[consumer] final class Runloop private (
...
+                      // Signal runloopDone FIRST so that removeSubscription calls in partition
+                      // stream finalizers can return immediately without offering to commandQueue.
+                      _ <- runloopDone.failCause(cause).ignore
+                      _ <- ZIO.foreachDiscard(state.assignedStreams)(_.halt(cause))
+                      _ <- partitionsHub.offer(Take.failCause(cause))
+                    } yield ()
+                  }
+      // Also signal on clean exit so callers are not left waiting.
+      _ <- runloopDone.fail(new RuntimeException("Runloop stopped")).ignore
+    } yield result
```
erikvanoosten: The outer for loop is only needed for a single flatmap with `runloopDone.fail(...)`. I propose we use `.zipLeft(runloopDone.fail...)` on the result of `.catchAllCause {...}`. This gives the exact same result with less nesting and a much smaller diff.

acrow: Done, applied in the latest commit. Thank you for the suggestion — it is indeed cleaner.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
```diff
@@ -92,7 +92,8 @@ final class PartitionStreamControl private (
    * To be invoked when the stream is no longer processing. It fails the stream with the given cause.
    */
   private[internal] def halt(cause: Cause[Throwable]): UIO[Unit] =
-    interruptionPromise.failCause(cause).unit
+    interruptionPromise.failCause(cause).ignore *>
+      dataQueue.offer(Exit.fail(cause.failureOption)).unit
```
erikvanoosten: Please revert this line. Adding an item to `dataQueue` is a workaround for the real problem which is that `dataQueue.poll` does not race with `interruptionPromise`.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/PartitionStreamControl.scala
erikvanoosten: Since `dataQueue.take` can't fail, we only need `mapError` when `interruptionPromise.await` completes. This suggestion removes `mapError` from the hot path and should make zio-kafka slightly faster.

```suggestion
          taken <- dataQueue.take.raceFirst(interruptionPromise.await.mapError(Option.apply)).unexit
```

I you want, you can ignore this suggestion, I can also create a separate PR.

erikvanoosten: Here is the real problem, `dataQueue.poll` does not race with `interruptionPromise.await`. Here is the fix:

```suggestion
                   dataQueue.poll.flatMap {
                   dataQueue.poll.raceFirst(interruptionPromise.await.mapError(Option.apply)).flatMap {
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -67,15 +68,20 @@ private[consumer] final class Runloop private (
...
+    runloopDone.isDone.flatMap {
+      case true => runloopDone.await
+      case false =>
+        (Promise
+          .make[E, A]
+          .flatMap { promise =>
+            commandQueue.offer(f(promise)) *> promise.await
+          }: Task[A])
+          .raceFirst(runloopDone.await)
+    }
```
erikvanoosten: During normal (non-error) shut down, runloop does not end the streams directly, only via the unsubscribe command. Therefore, skipping the command is problematic, the tests fail on it.

```suggestion
  private[internal] def offerAndAwaitCommand[E <: Throwable, A](f: Promise[E, A] => RunloopCommand): Task[A] =
    runloopDone.await.raceFirst {
         for {
            promise <- Promise.make[E, A]
            _       <- commandQueue.offer(f(promise))
            r       <- promise.await
          } yield r
        }
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -571,10 +577,15 @@ private[consumer] final class Runloop private (
...
-          _     <- partitionsHub.offer(Take.failCause(cause))
+          // Signal runloopDone FIRST so that removeSubscription calls in partition
+          // stream finalizers can return immediately without offering to commandQueue.
+          _ <- runloopDone.failCause(cause).ignore
+          _ <- ZIO.foreachDiscard(state.assignedStreams)(_.halt(cause))
+          _ <- partitionsHub.offer(Take.failCause(cause))
         } yield ()
       }
+      // Also signal on clean exit so callers are not left waiting.
+      .zipLeft(runloopDone.fail(new RuntimeException("Runloop stopped")).ignore)
```
erikvanoosten: This exception might be seen by the zio-kafka user. Most zio-kafka users don't know about the runloop (and ideally, they shouldn't need to know). Therefore:

```suggestion
      .zipLeft(runloopDone.fail(new RuntimeException("Consumer shutting down")).ignore)
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -571,10 +577,15 @@ private[consumer] final class Runloop private (
       .catchAllCause { cause =>
         for {
           state <- currentStateRef.get
-          _     <- ZIO.foreachDiscard(state.assignedStreams)(_.halt(cause))
-          _     <- partitionsHub.offer(Take.failCause(cause))
+          // Signal runloopDone FIRST so that removeSubscription calls in partition
+          // stream finalizers can return immediately without offering to commandQueue.
+          _ <- runloopDone.failCause(cause).ignore
```
erikvanoosten: With the changes in `offerAndAwaitCommand` I think this should go last. Either way, with the other changes, the comment is no longer applicable.

erikvanoosten (commented):
I found the problem that caused partition streams to dead lock. Please see the comments for a solution.

The new unit test fails now though. Do you want to take another look at that?

acrow:
Hi @erikvanoosten, thank you for finding the real root cause — the dataQueue.poll not racing with interruptionPromise was indeed the missing piece. All your suggestions have
  been applied in the latest commit:

  - dataQueue.poll now races with interruptionPromise.await.mapError(Option.apply) ✓
  - mapError moved inside the interruptionPromise branch in requestAndAwaitData ✓
  - dataQueue.offer(Exit.fail(...)) reverted from halt() ✓
  - Error message changed to "Consumer shutting down" ✓
  - runloopDone.failCause moved to last in catchAllCause ✓

  Regarding offerAndAwaitCommand — we kept the isDone check because our live tests hang without it, even with your raceFirst-only approach. We believe the reason is that removeSubscription is called from inside ZIO.addFinalizer, which runs uninterruptibly. When runloopDone.await wins the race, raceFirst interrupts the right fiber and waits for its finalizers. But the right fiber is inside an uninterruptible context, so the interrupt signal is masked and promise.await blocks forever — raceFirst never returns.
  
  The isDone check avoids this entirely: if the runloop is already done, we never start the right fiber, so there is nothing to interrupt.

  Regarding the unit test — the RunloopSpec test removeSubscription does not hang after Runloop crashes passes with all your changes applied. The test runs outside a finalizer context so it does not hit the uninterruptible issue.

erikvanoosten:
@acrow Thanks for that analysis.

I agree that the problem is that removeSubscription is called from inside ZIO.addFinalizer. I experimented with wrapping the call with `interruptible` like this:

```scala
// RunloopAccess subscribe
      _ <- ZIO.addFinalizer {
             ZIO.interruptible {
               withRunloopZIO(requireRunning = false)(_.removeSubscription(subscription).ignore) <*
                 diagnostics.emit(DiagnosticEvent.SubscriptionFinalized)
             }
           }
```

and that works a lot better, but there are still failing tests. In addition, it doesn't solve the race condition.

The race condition can occur because `Runloop.offerAndAwaitCommand` checks `runloopDone` from outside the fiber that runs the 'runloop'. What if we move the check to inside the fiber that runs the runloop? That is, from where we handle the command (method `handleCommand`).

Thinking again, this brings other problems... We are just moving the race condition; how do we make sure the runloop actually processes the queued commands?
... 🤔 ... 🤔 

This might work:
1. We keep `Runloop.offerAndAwaitCommand` as proposed in this PR.
2. In `Runloop`'s `catchAllCause`, we add some code that loops through all pending commands in `commandQueue` and fails the open promises.

Note: with this the race condition doesn't go away, we just make it less likely to result in a problem. Step 2 should be done _last_ inside `Runloop`'s `catchAllCause`, to make the chances of a missed command as small as possible.

The only way to really make the race condition go away is by having 2-way communication between `Runloop.offerAndAwaitCommand` and the runloop... Not sure how to do it yet.

We'll need to do step 2. also at end of the runloop (the second place where we call `runloopDone.fail`). (Perhaps move calling `runloopDone.fail` and emptying the command queue to a separate method...)

WDYT?

A side note: `Runloop.shouldPoll` should also check `runloopDone`:

```scala
  private def shouldPoll(state: State): UIO[Boolean] =
    for {
      runloopIsDone      <- runloopDone.isDone
      pendingCommitCount <- committer.pendingCommitCount
    } yield !runloopIsDone &&
      state.subscriptionState.isSubscribed &&
      (state.pendingRequests.nonEmpty || pendingCommitCount > 0 || state.assignedStreams.isEmpty)
```

acrow:
Hi @erikvanoosten, thank you for the detailed analysis and the suggestions.
  
  Both changes have been applied in the latest commit:

  - drainCommandQueue: after runloopDone is set, we now drain all remaining commands from commandQueue and fail their open promises with the runloop's cause. This is called at the end of both catchAllCause and zipLeft, so the window in which a command can be queued but never answered is as small as possible.
  - shouldPoll guards on runloopDone.isDone: the runloop now stops polling immediately once it has signalled done, rather than relying solely on the takeWhile stream guard.
  - AddSubscription.cont widened from Promise[InvalidSubscriptionUnion, Unit] to Promise[Throwable, Unit] so all three command types (AddSubscription, RemoveSubscription, EndStreamsBySubscription) can be drained uniformly.

  We tried to write a unit test that specifically exercises the race window (command queued while catchAllCause is running), but the window is too narrow to hit reliably in a unit test without injecting artificial delays into catchAllCause. The existing test removeSubscription does not hang after Runloop crashes still passes and validates the overall no-hang property. Please let us know if you'd like a different test approach.

erikvanoosten:
I finally got some time again (not much though) to look at this PR. Did you run the unit tests? When I run them on my laptop some of the test time out. I tried to debug the issue but I have not found the problem yet. Do the tests (`sbt test`) run on your machine?

erikvanoosten:
Oh smart! You swapped the place where `runloopDone` is checked in `offerAndAwaitCommand`!

I have some small tweaks standing by. I'll try to find time later today to send those over.

acrow:
Hi @erikvanoosten, yes we found the issue and have pushed a fix.

  The root cause was raceFirst in offerAndAwaitCommand. When runloopDone.await wins the race, raceFirst tries to interrupt the losing fiber (promise.await). But
  removeSubscription and endStreamsBySubscription are both called from ZIO.addFinalizer, which is uninterruptible — so the interrupt is masked and raceFirst waits forever for the loser's finalizers. This caused the widespread timeouts.

  We replaced raceFirst with a two-pronged approach:

  1. drainCommandQueue in catchAllCause: when the runloop crashes, drain all pending promise-bearing commands from the queue and fail their promises with the crash cause.
  Handles commands already in the queue at crash time.
  2. Post-offer isDone check in offerAndAwaitCommand: after offering the command, check runloopDone.isDone. If the runloop already crashed (and drainCommandQueue ran before our offer), fail the promise immediately. Then simply promise.await — no racing at all.

  There is still a narrow race window between commandQueue.offer and the isDone check, but it is effectively closed because runloopDone.failCause is called before drainCommandQueue in catchAllCause — so either drainCommandQueue sees the command, or the isDone check fires.

  We also added two simulation tests to RunloopSpec that wrap removeSubscription and endStreamsBySubscription in ZIO.uninterruptible, replicating the ZIO.addFinalizer context.
  These would have caught the raceFirst bug earlier.

  Results: all 8 RunloopSpec tests and all 44 ConsumerSpec integration tests pass.

erikvanoosten:
Thanks @acrow that is awesome work.

Could you apply this patch file? With that I think this PR is good to go!
[pr1712-zio-kafka.patch](https://github.com/user-attachments/files/29433262/pr1712-zio-kafka.patch)

erikvanoosten:
The failing test should be fixed on main branch via #1729.

acrow:
Hi @erikvanoosten, patch applied and rebased onto the latest master. All 8 RunloopSpec and 44 ConsumerSpec tests pass locally.

erikvanoosten (approved):

erikvanoosten:
I merged, but the tests are hanging on my computer. There is still something not right 😞

erikvanoosten:
@acrow finally fixed in ~~#1733~~ #1734.

The problem was that there was still a race condition in `offerAndAwaitCommand` causing us to the runloopDone signal. The fix, use `promise.await.raceFirst(runloopDone.await)` didn't work because the code is executed from a finalizer and hence uninterruptable (the loser can't be interrupted). This was fixed by creating an interruptable region.

PR #1713, opened 2026-06-08
Feature: opt-in detection of TopicAuthorizationException after silent poll

acrow:
# Problem

When a `READ ACL` is revoked at runtime, `poll()` silently returns `0` records for the affected topic instead of throwing `TopicAuthorizationException`. This leaves `Consumer.consumeWith` running indefinitely with no indication that authorization has been lost.

# Fix

Add `ConsumerSettings.withEmptyPollCountToMetaRefresh(n: Int)`: after n consecutive polls where a subscribed topic returned no records, call `committed()` for the assigned partitions of that topic. Method `committed()` throws `TopicAuthorizationException` on `DENY READ`, unlike `poll()` which swallows it silently.

The feature is disabled by default (`None`) — there is no overhead for users who do not use topic-level ACLs. All bookkeeping (Set operations, counter updates, committed() calls) is skipped entirely when the setting is not set.

The empty-poll count is tracked in `State.emptyPollCounts: Map[String, Int]` per topic, and reset when the topic returns records.

# Notes
  
- `committed()` was chosen over `partitionsFor()` because it throws on `DENY READ`, while `partitionsFor()` only throws on `DENY DESCRIBE`.
- Users subscribing to a very large number of low volume topics should set a higher threshold or leave this disabled.

### DIFF
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -259,11 +259,13 @@ private[consumer] final class Runloop private (
                                    .map(_.tp)
                                    .filter(tp => counts.getOrElse(tp.topic(), 0) >= threshold)
                                    .distinctBy(_.topic())
-                                 (if (partitionsToProbe.nonEmpty)
-                                    // committed() throws TopicAuthorizationException on DENY READ,
-                                    // unlike poll() which silently returns 0 records.
-                                    ZIO.attempt(c.committed(partitionsToProbe.toSet.asJava)).unit
-                                  else ZIO.unit).as(counts)
+                                 if (partitionsToProbe.nonEmpty)
+                                   // committed() throws TopicAuthorizationException on DENY READ,
+                                   // unlike poll() which silently returns 0 records.
+                                   ZIO
+                                     .attempt(c.committed(partitionsToProbe.toSet.asJava))
+                                     .as(counts -- partitionsToProbe.map(_.topic()))
+                                 else ZIO.succeed(counts)
                              }
 
             _ <- metricsObserver.observePoll(toResumeCount, toPauseCount, pollDuration, polledRecords.count())
--- a/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
+++ b/zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
@@ -226,7 +226,7 @@ private[consumer] final class Runloop private (
                s" resuming ${partitionsToFetch.size} out of ${state.assignedStreams.size} partitions"
            )
       _ <- currentStateRef.set(state)
-      pollResult <-
+      pollResultAndCounts <-
         consumer.runloopAccess { c =>
           for {
             assignment <- ZIO.attempt(c.assignment())
@@ -236,6 +236,40 @@ private[consumer] final class Runloop private (
 
             pullDurationAndRecords <- doPoll(c).timed
             (pollDuration, polledRecords) = pullDurationAndRecords
+
+            // When emptyPollCountToMetaRefresh is set, track consecutive empty polls per topic.
+            // Once the count reaches the threshold, call committed() to force a broker round-trip
+            // that surfaces any TopicAuthorizationException that poll() would otherwise swallow silently.
+            // All bookkeeping is skipped when the setting is None to avoid overhead for users that
+            // do not use topic-level ACLs.
+            updatedCounts <- settings.emptyPollCountToMetaRefresh match {
+                               case None => ZIO.succeed(state.emptyPollCounts)
+                               case Some(threshold) =>
+                                 val polledTopics  = polledRecords.partitions().asScala.map(_.topic()).toSet
+                                 val fetchedTopics = partitionsToFetch.map(_.topic())
+                                 // Also include subscribed topics with no assigned streams yet
+                                 // (e.g. ACL DENY before the consumer group has joined).
+                                 val unassignedTopics: Set[String] = state.subscriptionState match {
+                                   case SubscriptionState.Subscribed(_, Subscription.Topics(topics))
+                                       if state.assignedStreams.isEmpty =>
+                                     topics
+                                   case _ => Set.empty
+                                 }
+                                 val topicsToTrack = fetchedTopics ++ unassignedTopics
+                                 val counts = topicsToTrack.foldLeft(state.emptyPollCounts) { (acc, topic) =>
+                                   if (polledTopics.contains(topic)) acc - topic
+                                   else acc + (topic -> (acc.getOrElse(topic, 0) + 1))
+                                 }
+                                 val partitionsToProbe = state.assignedStreams
+                                   .map(_.tp)
+                                   .filter(tp => counts.getOrElse(tp.topic(), 0) >= threshold)
+                                   .distinctBy(_.topic())
+                                 (if (partitionsToProbe.nonEmpty)
+                                    // committed() throws TopicAuthorizationException on DENY READ,
+                                    // unlike poll() which silently returns 0 records.
+                                    ZIO.attempt(c.committed(partitionsToProbe.toSet.asJava)).unit
+                                  else ZIO.unit).as(counts)
+                             }
 
             _ <- metricsObserver.observePoll(toResumeCount, toPauseCount, pollDuration, polledRecords.count())
             _ <- diagnostics.emit {
@@ -349,8 +383,9 @@ private[consumer] final class Runloop private (
                                 assignedStreams = updatedAssignedStreams
                               )
                           }
-          } yield pollresult
+          } yield (pollresult, updatedCounts)
         }
+      (pollResult, updatedCounts) = pollResultAndCounts
       fulfillResult <- offerRecordsToStreams(
                          pollResult.assignedStreams,
                          pollResult.pendingRequests,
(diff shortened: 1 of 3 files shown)


erikvanoosten:
A few questions before I dive into the implementation:
- When using a subscription on multiple topics (or on specific partitions), is it okay to continue with the remaining topics/partitions? Or only when the number of remaining topics is non-zero? When using a pattern subscription, how to detect the number of remaining topics?
- To detect if the client has topic access, do we need to call `committed` for all partitions, or only one per topic (e.g. the first)?
- Can we call `committed` for all partitions in bulk, in parallel, or do we need to do this serially?

acrow:
Hi @erikvanoosten, thank you for the questions. We ran tests to confirm the actual behaviour:

  1. Multiple topics — does the consumer continue on remaining topics?
  No. When DENY READ is applied to one topic while another topic has active traffic, committed() throws TopicAuthorizationException which propagates through handlePoll and kills
  the entire consumer. This is confirmed by our ZioKafkaBugRepro test which subscribes to two topics and only denies one. The behaviour may not be ideal — we are open to a
  different approach if you prefer the consumer to continue on remaining topics.

  For pattern subscriptions: the unassignedTopics path only works for Subscription.Topics — pattern subscriptions are not currently handled.

  2. One partition per topic or all?
  Currently we call committed() for all assigned partitions of the silent topic. One per topic is sufficient for detection — we can optimize this if preferred.

  3. Bulk or serial?
  c.committed(Set<TopicPartition>) is a single bulk call to the broker — all partitions are probed in one request.

acrow:
Hi @erikvanoosten, just a quick update: we went ahead and applied the one-partition-per-topic optimization mentioned in our previous reply. partitionsToProbe now calls .distinctBy(_.topic()) before passing to committed(), so at most one partition per silent topic is probed regardless of how many partitions are assigned. The change is in the latest commit.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -232,6 +232,40 @@ private[consumer] final class Runloop private (
...
+                                 val polledTopics  = polledRecords.partitions().asScala.map(_.topic()).toSet
+                                 val fetchedTopics = partitionsToFetch.map(_.topic())
+                                 // Also include subscribed topics with no assigned streams yet
+                                 // (e.g. ACL DENY before the consumer group has joined).
+                                 val unassignedTopics: Set[String] = state.subscriptionState match {
+                                   case SubscriptionState.Subscribed(_, Subscription.Topics(topics))
+                                       if state.assignedStreams.isEmpty =>
+                                     topics
+                                   case _ => Set.empty
+                                 }
```
erikvanoosten: Tip: read the next comment first.

Its fairly straightforward to also support manual subscriptions:

```scala
val unassignedTopics: Set[String] = if (state.assignedStreams.isEmpty) {
  state.subscriptionState match {
    case SubscriptionState.Subscribed(_, Subscription.Topics(topics)) => topics
    case SubscriptionState.Subscribed(_, Subscription.Manual(topicPartitions)) => topicPartitions.map(_.topic())
    case _ => Set.empty
  }
} else Set.empty
```

But I actually prefer the following to make sure we address new subscription modes:

```suggestion
                                 val unassignedTopics: Set[String] = if (state.assignedStreams.isEmpty) {
                                   state.subscriptionState match {
                                     case SubscriptionState.Subscribed(_, Subscription.Topics(topics)) => topics
                                     case SubscriptionState.Subscribed(_, Subscription.Manual(topicPartitions)) =>
                                       topicPartitions.map(_.topic())
                                     case SubscriptionState.Subscribed(_, Subscription.Pattern(_)) => Set.empty
                                     case SubscriptionState.NotSubscribed => Set.empty
                                   }
                                 } else Set.empty
```

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -232,6 +232,40 @@ private[consumer] final class Runloop private (
...
+                                 val topicsToTrack = fetchedTopics ++ unassignedTopics
+                                 val counts = topicsToTrack.foldLeft(state.emptyPollCounts) { (acc, topic) =>
+                                   if (polledTopics.contains(topic)) acc - topic
+                                   else acc + (topic -> (acc.getOrElse(topic, 0) + 1))
+                                 }
+                                 val partitionsToProbe = state.assignedStreams
```
erikvanoosten, line 258: I understand we don't want to probe partitions that are not yet assigned. This is however not consistent with the `unassignedTopics` calculated above. Here is the scenario I have in mind:

1. consumer starts, initiates subscription for a denied topic, and starts the runloop
1. as `state.assignedStreams.isEmpty` is true, the topic we subscribed to is already counted
1. some loops later, the topic is assigned, and appears in `state.assignedStreams`
  **Only now we call `committed` for this topic** (We might even call it immediately as the threshold might have been reached already.)

If we would remove variable `unassignedTopics`, almost the exact same thing would happen, only a few run loops later.

My conclusion therefore is that we don't need to consider topics that are unassigned at all (we can remove variable `unassignedTopics` completely). Or did I miss something?

erikvanoosten (commented):
Looks good. I think we can get rid of `unassignedTopics`, please check my reasoning.

We'll need some more documentation around this, but I can write that if you want.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: I see variable `unassignedTopics` was removed. Resolving this comment.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
erikvanoosten: No longer relevant, this code has been removed.

erikvanoosten:
@acrow did you think about adding a new test for this feature?

acrow:
Hi @erikvanoosten, we added a unit test to RunloopSpec that verifies the feature end-to-end:

  - Assigns a partition with no records (poll() silently returns empty, as Kafka does on ACL DENY READ)
  - After 3 consecutive empty polls (the configured threshold), committed() is called
  - committed() throws TopicAuthorizationException, which crashes the runloop and fails the partition stream

  The test subclasses MockConsumer to make committed() throw on demand, since MockConsumer.setOffsetsException only affects beginningOffsets/endOffsets, not committed().

  Regarding the CI failures — we investigated and it looks like runWithGracefulShutdown must end streams while still processing commits when being interrupted and should close old stream during rebalancing under load also fail on the current upstream master (we believe they were introduced in 970cd9f / #1733). We don't think these are related to our changes — but please let us know if we missed something!

erikvanoosten:
Thanks for confirming my earlier comment with the code change and adding the test.

> we believe they were introduced in https://github.com/zio/zio-kafka/commit/970cd9f723999e228599d26f9028652482140fa0 / https://github.com/zio/zio-kafka/pull/1733

Yeah, that could very well be the case. I'll have to fix that before this PR can be merged though. Its tricky stuff, we're working on the edge of what ZIO can do. The shutdown sequence behaves differently on fast and slow machines as well. So testing locally is hard. Anyways, I'll figure it out...

erikvanoosten:
Haha, turns out that the problem was already fixed in #1734 which I thought was already merged...
Anyways, I rebased this PR and it seems fine now.

zio-kafka/src/main/scala/zio/kafka/consumer/internal/Runloop.scala
```diff
@@ -241,6 +241,31 @@ private[consumer] final class Runloop private (
...
+                                   .map(_.tp)
+                                   .filter(tp => counts.getOrElse(tp.topic(), 0) >= threshold)
+                                   .distinctBy(_.topic())
+                                 (if (partitionsToProbe.nonEmpty)
+                                    // committed() throws TopicAuthorizationException on DENY READ,
+                                    // unlike poll() which silently returns 0 records.
+                                    ZIO.attempt(c.committed(partitionsToProbe.toSet.asJava)).unit
+                                  else ZIO.unit).as(counts)
```
erikvanoosten: WDTY of removing topics that have just been probed from emptyPollCounts? That would prevent the map from ever growing in the edge case where the list of topics keeps varying over time.

```suggestion
                                 if (partitionsToProbe.nonEmpty) {
                                    // committed() throws TopicAuthorizationException on DENY READ,
                                    // unlike poll() which silently returns 0 records.
                                    ZIO.attempt(c.committed(partitionsToProbe.toSet.asJava))
                                      .as(counts -- partitionsToProbe.map(_.topic()))
                                 } else ZIO.succeed(counts)
```

erikvanoosten: Once this is resolved, I think we're good for merging.

erikvanoosten:
@acrow Documentation added via #1736.

acrow:
Done — after calling committed(), the probed topics are now removed from emptyPollCounts (counts -- 
  partitionsToProbe.map(_.topic())), so the map stays bounded. All 9 RunloopSpec tests pass locally.
     
  I also noticed you pushed 327d591 and 9804ee6 while I was working on this — I've rebased my commit on top of your changes.

  The CI failure on Java 21 is unrelated to our changes: Consumer subscriptions - can resume a stream for the same subscription in SubscriptionsSpec timed out after 2 minutes while the embedded Kafka broker was flooded with OutOfOrderSequenceException errors. Java 17 and Java 25 both passed cleanly.

erikvanoosten (approved):
Nice work!

PR #1733, opened 2026-07-01
Fix: removeSubscription hangs after Runloop crashes - followup

erikvanoosten:
After merging #1712 some test were still failing fast computers. We tracked this to a remaining race condition in `offerAndAwaitCommand`.

Here we simplify `offerAndAwaitCommand` by awaiting both promises in an interruptible region.


erikvanoosten:
The test are still failing 🤔

